PR #13, opened 2025-04-08
including langchain-openai package in toml

rafaelpadilla:
While trying to reproduce the minimal example in the README.md, I noticed it requires some imports:

```python
import asyncio
from dotenv import load_dotenv
from langchain_openai import ChatOpenAI
from mcp_use import MCPAgent, MCPClient
```

However, the package `langchain_openai` is missing in `pyproject.toml`.

This PR includes `langchain_openai` in `pyproject.toml`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -22,6 +22,7 @@ dependencies = [
     "mcp",
     "langchain>=0.1.0",
     "langchain-community>=0.0.10",
+    "langchain-openai>=0.3.12",
     "websockets>=12.0",
     "aiohttp>=3.9.0",
     "pydantic>=2.0.0",


pietrozullo:
Hey @rafaelpadilla thank you so much for the contribution, I see that this could help a lot of people getting it to run quickly. Though langchain_openai is not really a dependency so we cannot include it there, it is already in optional dependencies. What I think is a good solution is document this better in the README showing that one should install the langchain provider of the llm he wants to use. Would you like to do it ?

pietrozullo:
Or is this good enough you think ? https://github.com/pietrozullo/mcp-use/commit/6a504e4b049a7fad003b866d0fcfbb707f0e4eed

rafaelpadilla:
Hi @pietrozullo 

Thank you for your reply.

Yes. Totally agree that as `langchain_openai` and `langchain-anthropic` are not part of the `mcp-use` core code base, they should not be included in the .toml. 

IMO, having such instructions in the README, as you suggested works great!

Thank you for your clarification. 

Will close this issue now :)

PR #47, opened 2025-04-19
fix: remove redundant session initialization in create_all_sessions

zandko:
## Changes

The primary change introduced by this PR is the elimination of redundant session initialization calls in the create_all_sessions method. This improves performance and reduces unnecessary operations.

## Implementation Details

1. The call to await session.initialize() within the loop in create_all_sessions was removed because create_session already handles initialization when auto_initialize is True.
2. The code organization was slightly modified to ensure clarity and efficiency.
3. No new dependencies were added or existing ones modified as part of this change.

## Documentation Updates

No documentation files were updated since the changes are internal optimizations and do not affect the public API or usage guidelines.

## Backwards Compatibility

These changes are fully backwards compatible. Users will not need to modify any of their existing code as the method signature and behavior remain consistent, aside from improved performance.

### DIFF
--- a/mcp_use/client.py
+++ b/mcp_use/client.py
@@ -162,9 +162,7 @@ class MCPClient:
 
         # Create sessions for all servers
         for name in servers:
-            session = await self.create_session(name, auto_initialize)
-            if auto_initialize:
-                await session.initialize()
+            await self.create_session(name, auto_initialize)
 
         return self.sessions
 


pietrozullo:
Hey thanks for the contribution, you should change in test_client.py these lines: 
```python
# In the implementation, initialize is called twice for each session:
# Once in create_session and once in the explicit initialize call
assert mock_session1.initialize.call_count == 2
assert mock_session2.initialize.call_count == 2
```

To:
```python 
# Once in create_session and once in the explicit initialize call
assert mock_session1.initialize.call_count == 1
assert mock_session2.initialize.call_count == 1
```

This was a mistake in the tests, your implementation is correct.

madebyzan:
> Collaborator

Thank you for your reminder, the changes have been made.

Issue #70, opened 2025-04-29
Streamable HTTP Support


ferrants:
MCP Released their new spec in March, https://modelcontextprotocol.io/specification/2025-03-26/changelog
And it supports a Streamable HTTP transport and new authentication methods. Are these supported (they do not appear to be supported yet, few tools are supporting these yet)? Are there plans to support this?

Thanks!

pietrozullo:
Hey thanks! Yes I think this should work off the bat, since the connection is handled by mcp library. Do you find it to not work ? I will have a look.

pietrozullo:
Hey @ferrants did you have a chance to test in your use case ? I just tested with https://github.com/modelcontextprotocol/servers/tree/main/src/everything
and as I assumed the streamablehttp works off the bat. 
I will close the issue let me know if you encounter any issue with this. 

Best and thank you for the feedback 🤗

leonph92:
@pietrozullo I couldn't get this to properly connect with a server through a streamable http transport. If I'm not mistaken, the [http connector](https://github.com/mcp-use/mcp-use/blob/main/mcp_use/connectors/http.py) used the [sse client](https://github.com/mcp-use/mcp-use/blob/main/mcp_use/task_managers/sse.py#L56) under the hood. We'd need one that use the [streamable one](https://github.com/modelcontextprotocol/python-sdk/blob/main/src/mcp/client/streamable_http.py). 

If it's out of your bandwidth, I could help contribute to this.

pietrozullo:
Hey @leonph92 ;) Thanks for your message, yeah I would be super happy to see your contribution! Let me know if you have bw otherwise I can take care of this as well

leonph92:
Just an update, I started this and just need to test to verify the functionality. Apologies for the delays.

PR #106, opened 2025-05-28
Feature/streamable http support

leonpham-mbh:
# Pull Request Description

## Changes

Extends the `HttpConnector` to support both the older SSE and newer streamable HTTP transport. This defaults to the SSE transport for backwards compatibility.

These changes update the `HttpConnector` instead of making a new one since the SSE one is considered [deprecated](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports#streamable-http).

## Implementation Details

1. Introduce new streamable HTTP task manager
2. Extends the `HttpConnector` with a `use_streamable_http` flag that will switch what connection manager will be used.
3. Update configuration to pass down the flag

## Example Usage (Before)

```python
config = {"mcpServers": {"http": {"url": "http://localhost:8000/sse"}}}

# Create MCPClient from config file
client = MCPClient.from_dict(config)
```

## Example Usage (After)

```python
config = {"mcpServers": {"http": {"url": "http://localhost:8000/mcpunsecured", "use_streamable_http": True}}}

# Create MCPClient from config file
client = MCPClient.from_dict(config)
```

## Documentation Updates
N/A

### Unit Tests
Added new unit tests to verify support of both SSE and streamable HTTP

### Manual Testing
1. Checked out [this](https://github.com/omarbenhamid/django-mcp-server/tree/main/examples/mcpexample) example MCP server that uses streamable http
2. Ran the following script:
```
import asyncio

from dotenv import load_dotenv
from langchain_openai import ChatOpenAI

from mcp_use import MCPAgent, MCPClient


async def main():
    """Run the example using a configuration file."""
    # Load environment variables
    load_dotenv()

    config = {"mcpServers": {"http": {"url": "http://localhost:8000/mcpunsecured", "use_streamable_http": True}}}

    # Create MCPClient from config file
    client = MCPClient.from_dict(config)

    # Create LLM
    llm = ChatOpenAI(model="gpt-4o")

    # Create agent with the client
    agent = MCPAgent(llm=llm, client=client, max_steps=5)

    # Run the query
    result = await agent.run(
        "How many robins are there?",
        max_steps=5,
    )
    print(f"\nResult: {result}")


if __name__ == "__main__":
    # Run the appropriate example
    asyncio.run(main())
```
5. Verified results 
    image

## Backwards Compatibility
They are

## Related Issues

Closes https://github.com/mcp-use/mcp-use/issues/70

### DIFF
--- a/mcp_use/config.py
+++ b/mcp_use/config.py
@@ -72,6 +72,7 @@ def create_connector_from_config(
             base_url=server_config["url"],
             headers=server_config.get("headers", None),
             auth_token=server_config.get("auth_token", None),
+            use_streamable_http=server_config.get("use_streamable_http", False),
         )
 
     # WebSocket connector
--- a/mcp_use/connectors/http.py
+++ b/mcp_use/connectors/http.py
@@ -8,14 +8,14 @@ through HTTP APIs with SSE for transport.
 from mcp import ClientSession
 
 from ..logging import logger
-from ..task_managers import SseConnectionManager
+from ..task_managers import SseConnectionManager, StreamableHttpConnectionManager
 from .base import BaseConnector
 
 
 class HttpConnector(BaseConnector):
-    """Connector for MCP implementations using HTTP transport with SSE.
+    """Connector for MCP implementations using HTTP transport with SSE or streamable HTTP.
 
-    This connector uses HTTP/SSE to communicate with remote MCP implementations,
+    This connector uses HTTP/SSE or streamable HTTP to communicate with remote MCP implementations,
     using a connection manager to handle the proper lifecycle management.
     """
 
@@ -26,6 +26,7 @@ class HttpConnector(BaseConnector):
         headers: dict[str, str] | None = None,
         timeout: float = 5,
         sse_read_timeout: float = 60 * 5,
+        use_streamable_http: bool = False,
     ):
         """Initialize a new HTTP connector.
 
@@ -35,6 +36,7 @@ class HttpConnector(BaseConnector):
             headers: Optional additional headers.
             timeout: Timeout for HTTP operations in seconds.
             sse_read_timeout: Timeout for SSE read operations in seconds.
+            use_streamable_http: Whether to use streamable HTTP instead of SSE.
         """
         super().__init__()
         self.base_url = base_url.rstrip("/")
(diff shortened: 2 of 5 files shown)


pietrozullo:
Hey @leonpham-mbh! Thank you for the contribution, looks great :hugs: :hugs:  !

I have one question, I think that if somebody passes in a URL to a streamableHTTP server and the use_streamable_http flag was set to False, the sse_client would fallback to the streamable http, and therefore it would work. I wonder if we completely removed sse_client and just assume that everything is streamablehttp,  would the mcp.client.streamable_http.streamablehttp_client also be able to handle the deprecated sse case ? 
If this was true I would remove the sse_client altogether, which would then in turn remove the need of a flag use_streamable_http and make the whole thing easier to use. 

What do you think ?

leonpham-mbh:
> Hey @leonpham-mbh! Thank you for the contribution, looks great 🤗 🤗 !
> 
> I have one question, I think that if somebody passes in a URL to a streamableHTTP server and the use_streamable_http flag was set to False, the sse_client would fallback to the streamable http, and therefore it would work. I wonder if we completely removed sse_client and just assume that everything is streamablehttp, would the mcp.client.streamable_http.streamablehttp_client also be able to handle the deprecated sse case ? If this was true I would remove the sse_client altogether, which would then in turn remove the need of a flag use_streamable_http and make the whole thing easier to use.
> 
> What do you think ?

@pietrozullo I like that idea! Unfortunately, I tested just using the streamablehttp client to see if it'll fall back to try to connect using SSE and it does not. I could try to test the connection inside the `HttpConnector.connect` method to avoid passing in the `use_streamable_http`.

leonpham-mbh:
@pietrozullo I pushed an update that will setup a connection using streamable http, ping the server, and if it fails fallback to sse. I updated the pull request description. Let me know if that's what you had in mind.

pietrozullo:
Thanks :hugs: Let me test :)

Pederzh (approved):
LGTM

pietrozullo:
Hey @leonpham-mbh thank you so much for you contribution, I fixed a few things and put some tests now should work.
Please Join discord https://discord.gg/XkNkSkMz3V we'll give you some contributor recognition :)

leonpham-mbh:
> Hey @leonpham-mbh thank you so much for you contribution, I fixed a few things and put some tests now should work. Please Join discord https://discord.gg/XkNkSkMz3V we'll give you some contributor recognition :)

Thank you @pietrozullo! Glad I could help 🙇

PR #117, opened 2025-06-08
Add codespell support (config, workflow to detect/not fix) and make it fix some typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

TODOs
- [x] Potentially remove a dedicated CI, since there is pre-commit?  Note that CI does nice annotation on PR diff files on where and what typo detected
- [x] Remove TEMP demonstrating typo detection

### DIFF
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,25 @@
+# Codespell configuration is within pyproject.toml
+---
+name: Codespell
+
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+
+permissions:
+  contents: read
+
+jobs:
+  codespell:
+    name: Check for spelling errors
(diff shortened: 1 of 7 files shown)


yarikoptic:
apparently pre-commit is not tested by CI, so leaving CI workflow in.

Here is how hinting from action looked like:
image

now removing that TEMP commit.

pietrozullo:
Is this going to GPL mcp-use ?

yarikoptic:
> Is this going to GPL mcp-use ?

AFAIK no, you aren't incorporating codespell into your product, it is just a tool you use and without modifications

yarikoptic:
pity

PR #133, opened 2025-06-18
Fix: prevent double async cleanup and event loop errors in astream 

### LINKED ISSUE #132: Double cleanup and event loop errors when using MCPAgent.astream
**Describe the bug**
When using the `MCPAgent.astream` generator, breaking out of the async for loop early (i.e., not fully exhausting the generator) causes resource cleanup errors and warnings. Specifically, warnings such as "Attempted to exit a cancel scope that isn't the current tasks's current cancel scope" and "RuntimeError: Event loop is closed" appear on shutdown.

**To Reproduce**
1. Create an `MCPAgent` and call `agent.astream()` in an async for loop.
2. Break out of the loop early (e.g., with a `break` statement or exception).
3. Allow the program to exit.
4. See resource cleanup warnings/errors in the logs.

**Expected behavior**
Cleanup should always be handled correctly, and no resource warnings or event loop errors should occur, even if the generator is not fully exhausted.

**sample log**
`mcp_use - WARNING - Error closing client session: Attempted to exit a cancel scope that isn't the current tasks's current cancel scope
2025-06-17 22:03:23,129 - mcp_use - WARNING - Encountered 1 errors during resource cleanup
success
Loop <_UnixSelectorEventLoop running=False closed=True debug=False> that handles pid 70420 is closed
Exception ignored in: <function BaseSubprocessTransport.__del__ at 0x1059e0860>
Traceback (most recent call last):
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_subprocess.py", line 126, in __del__
    self.close()
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_subprocess.py", line 104, in close
    proto.pipe.close()
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/unix_events.py", line 566, in close
    self._close(None)
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/unix_events.py", line 590, in _close
    self._loop.call_soon(self._call_connection_lost, exc)
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_events.py", line 761, in call_soon
    self._check_closed()
  File "/Users/hackarada/.pyenv/versions/3.11.6/lib/python3.11/asyncio/base_events.py", line 519, in _check_closed
    raise RuntimeError('Event loop is closed')
RuntimeError: Event loop is closed`

**Additional context**
This appears to be caused by double cleanup or cleanup from the wrong async context when the generator is not fully iterated. The issue is resolved by only performing cleanup when the generator is responsible for the agent's lifecycle (i.e., when `not self.client and initialised_here`).

hackarada:
# Pull Request Description

## Changes

This PR fixes improper async resource cleanup in the `MCPAgent` streaming generator (`_generate_response_chunks_async`).  
Cleanup is now only performed when the generator is responsible for the agent's lifecycle, preventing double cleanup and event loop errors.

## Implementation Details

1. Modified `_generate_response_chunks_async` to restrict agent cleanup:
    - Cleanup (`await self.close()`) now only occurs if the agent was initialized in this generator and is not client-managed (`not self.client and initialised_here`).
2. This prevents double cleanup and avoids closing async resources from the wrong context or after the event loop is closed.
3. No new dependencies were added or modified.

## Example Usage (Before)

```python
async for event in agent.astream("query"):
    if some_condition:
        break  # Could cause resource warnings/errors on shutdown
```

## Example Usage (After)

```python
async for event in agent.astream("query"):
    if some_condition:
        break  # No resource warnings/errors; cleanup is handled correctly
```

## Documentation Updates

* No documentation files were updated.
* The change is internal and does not affect the public API.

## Testing

- Manual testing: Ran streaming queries and broke out of the loop early; verified no resource warnings or event loop errors on shutdown.
- Observed that warnings like "Attempted to exit a cancel scope that isn't the current tasks's current cancel scope" and "RuntimeError: Event loop is closed" no longer appear.
- Existing tests for agent streaming and cleanup continue to pass.

## Backwards Compatibility

This change is fully backwards compatible.  
No changes are required for users; the fix only improves internal resource management and error handling.

## Related Issues

Closes #132


mcp_use/agents/mcpagent.py
```diff
@@ -365,9 +351,11 @@ async def _generate_response_chunks_async(
                         if not isinstance(message, ToolAgentAction):
                             self.add_to_history(message)
             yield event
-
         # 5. House-keeping -------------------------------------------------------
-        if initialised_here and manage_connector:
+        # Restrict agent cleanup in _generate_response_chunks_async to only occur when the agent was initialized in this generator and is not client-managed.
+        # No need to check manage_connector as its already checked when setting initialised_here
+        if not self.client and initialised_here:
+            logger.info("🧹 Closing agent after generator completion")
```
Pederzh: Thanks for taking the time to fix this @hackarada.
Looking at the cleanup logic, I noticed a small discrepancy between the comment and the actual behavior that might be worth discussing.

Looking at line 328:
  `if (manage_connector and not self._initialized) or (not self._initialized and self.auto_initialize):`

The initialised_here flag can be set to True in two scenarios:
  1. manage_connector and not self._initialized - Agent manages connector lifecycle
  2. not self._initialized and self.auto_initialize - Auto-initialization regardless of manage_connector

This means `initialised_here = True` can occur even when `manage_connector = False (if auto_initialize = True)`.

The correct cleanup condition should be:
  if not self.client and initialised_here and manage_connector:
```suggestion
        # Restrict agent cleanup in _generate_response_chunks_async to only occur when the agent was initialized in this generator and is not client-managed and the user does want us to manage the connection.
        if not self.client and initialised_here and manage_connector:
            logger.info("🧹 Closing agent after generator completion")
```

Why this matters:
  - If `manage_connector = False`, the caller expects to handle lifecycle management
  - The current fix could still cause unwanted cleanup in auto-initialize scenarios where the caller wants to manage the lifecycle
  
Let me include @pietrozullo in the convo.
`manage_connector` does need to be included to properly respect the user's intent about lifecycle management, right?

pietrozullo (approved):

mcp_use/agents/mcpagent.py
hackarada: I see your point there, That works.I though we are already checking for manage_connector/auto_initialize when setting initialised_here in # 1. Initialise on-demand.

pietrozullo:
@claude can you fix the linting error ? its line length on mcpagent.py, do not do anything else

PR #184, opened 2025-07-22
Retrieved the list of tools used post agent execution

NandhuKishore24:
# Pull Request Description

## Changes

Replaced local variable tools_used_names to class attribute.

## Implementation Details

The tools_used_names list was being constructed within the stream and _consume_and_return methods by appending tool names before each call. However, since it wasn't returned from the run method, it was inaccessible after the query execution.

Returning tools_used_names directly from run alongside the response would require consumers to unpack a tuple (e.g., response[0]), forcing all downstream code to be updated.

To avoid breaking existing integrations, I converted tools_used_names from a local variable to a class-level attribute. This makes the list accessible post-execution without altering the return signature of the run method, thus preserving backward compatibility.

## Example Usage (Before)
NA

## Example Usage (After)
NA

## Documentation Updates
NA

## Testing

- Manual testing performed

## Backwards Compatibility
yes the changes preserving backward compatibility.

## Related Issues

Closes #[183]


pietrozullo:
Thanks @NandhuKishore24 :hugs: merged!

PR #201, opened 2025-07-30
chore(example): multi_server_google_gemini_example.py

ayushxx7:
# Pull Request Description

## About
Google is giving out free use API keys, so it is useful to have this example file that uses the `gemini-2.5-pro` model


harshithvh:
Hi @ayushxx7, thank you for your contributions! It would be great if the examples could showcase some of the key functionalities of the mcp-use library.

pietrozullo:
Hey @ayushxx7 thanks for the contribution, indeed a multi server example already exists! If the documentation regarding how to use google is not enough please let me know, and in any case: keep contributing!

PR #208, opened 2025-08-03
Implementing logic to create sessions only for the specified servers

### LINKED ISSUE #82: server_name parameter in MCPAgent.run() missing
All available documentation still describes the use of server_name in MCPAgent.run().

As #52 says this is obsolete, because implementation changed. So what is now the right way to limit a run to specific servers?

NandhuKishore24:
# Pull Request Description
This PR introduces logic to allow the MCPClient to create sessions only for a specified subset of servers, based on the allowed_servers parameter. If allowed_servers is not provided, sessions will be created for all servers as before.
This enhancement enables more fine-grained control over session creation, which is especially useful in scenarios where the client should only connect to specific MCP servers depends on request comes from different sources, improving performance and flexibility.

Example:
I am maintaining a single config.json file which contains all the server details. But I am providing service to different applications. So based on the request coming from different applications I can add the server names in the allowed server list instead of having multiple config.json file.

## Changes
Added filtering logic to create sessions only for servers listed in allowed_servers.
Updated create_all_sessions method to conditionally create sessions.
Added safe handling for None or empty allowed_servers

## Implementation Details
1. MCPClient accepts an optional parameter called allowed_servers list.
2. This list is used to restrict session creation to only for the specified servers.
Assumption: All MCP server names and allowed_servers entries follow the lowercase naming convention as per usage standard.

## Example Usage (Before)

```python
# client = MCPClient(config=config)
```

## Example Usage (After)

```python
# allowed_servers is optional parameter
# client = MCPClient(config=config, allowed_servers=['server1']) 
```

## Documentation Updates
NA

## Testing

- Unit tests added in test_client.py (test_create_allowed_sessions)
- Manual testing performed

## Backwards Compatibility
Yes maintaining backward compatibility.

## Related Issues

Closes #82
Need to create one issue only to update the documentation.


NandhuKishore24:
Raised with conflicts

PR #217, opened 2025-08-05
fix: outdated create session from config functionality

kmontocam:
remove import and details of undefined method

# Pull Request Description

Removed undefined method on package init

## Changes

Removed import alias from `__all__` when import module was undefined, dropped from documentation at README


pietrozullo:
Thanks @kmontocam :)

PR #236, opened 2025-08-16
docs: Add chat loop examples

renvins:
The agent documentation was missing of examples on how to create chat loops, with streaming and not. Three examples will be added:

- A basic chat loop with clear and quit commands
- A streaming chat loop with the same commands
- An example with other commands like tools etc.

For any ideas, I'm open to discuss.


renvins:
@pietrozullo if you agree, it could be ready to be merged.

pietrozullo (approved):

PR #249, opened 2025-08-26
Chat ID inclusion to conversation history

THARUN-S-R:
# Chat ID inclusion to the conversation history

## Changes

We are passing the Chat Id variable into the conversation history to better identify the chats within the same session.

## Implementation Details

The BaseMessage of langchain has capabilities to accept the ID which we can use to identify the exact chat incase we need to use it for training or feedback purposes

This change will not be affected if chat id is not sent.

### DIFF
--- a/mcp_use/adapters/langchain_adapter.py
+++ b/mcp_use/adapters/langchain_adapter.py
@@ -39,7 +39,7 @@ class LangChainAdapter(BaseAdapter):
         self._connector_tool_map: dict[BaseConnector, list[BaseTool]] = {}
 
     def fix_schema(self, schema: dict) -> dict:
-        """Convert JSON Schema 'type': ['string', 'null'] to 'anyOf' format.
+        """Convert JSON Schema 'type': ['string', 'null'] to 'anyOf' format and fix enum handling.
 
         Args:
             schema: The JSON schema to fix.
@@ -51,6 +51,11 @@ class LangChainAdapter(BaseAdapter):
             if "type" in schema and isinstance(schema["type"], list):
                 schema["anyOf"] = [{"type": t} for t in schema["type"]]
                 del schema["type"]  # Remove 'type' and standardize to 'anyOf'
+
+            # Fix enum handling - ensure enum fields are properly typed as strings
+            if "enum" in schema and "type" not in schema:
+                schema["type"] = "string"
+
             for key, value in schema.items():
                 schema[key] = self.fix_schema(value)  # Apply recursively
         return schema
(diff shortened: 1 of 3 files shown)


THARUN-S-R:
hey @pietrozullo / @NandhuKishore24  , can you please review the changes for this request.

pietrozullo:
Hey @THARUN-S-R actually the chat id has no function on the local agent but only on the remote execution. I am not sure what are you trying to do here, could you explain ?

THARUN-S-R:
Hi @pietrozullo , I am using the Chat Id to pass a unique hash value to identify each agent run or Chat. This will be then added to the conversation history. I need to find each Chat to add a feedback to the response. This way i can identify the chat and improve prompting based on user feedback. For this langchain BaseMessage has optional id field which we can use to identify each chat.

pietrozullo:
Hey @THARUN-S-R I am still not sure I follow, have you see what the chat_id is used for inside the agent ? The chatID is a reference to the cloud hosted agent in the library context, it will not have anything to do with chat persistence locally, even after your changes. Also I see you are passing it in the langchain messages, which is fine, but that id is supposed to be unique for every message not a common one for a chat. If you want to persiste the messages in a chat you can create a Chat class where you persist them and you can have the Chat take an ID. 

Maybe I am not seeing something here but this PR does not seem to solve the problem you want to address, imo

THARUN-S-R:
Hi @pietrozullo , yes the chat id initially is used only for remote agent, but in my use case I want to mark each chat with a unique id to reference it back later. Hence, this field was already there which i can use to pass the unique id into the Messages. This way i dont need a new parameter to achieve this and can be assigned in Chat_id. About the uniqueness of the id, before triggering the run(query) func, I will assign the agent a new chat id as :
agent.chat_id = chat_id 
this assignment is not within the MCP-USE architecture and is at the users discretion. this way it doesnt affect the architecture if not passed but gives the user options to assign an id to the messages.:
image

This way I can easily identify the chats I want to refer back to. The user can have a feedback to the chat generated which we can use to adjust our prompt accordingly. This is just one of the use case of the Chat ID.

pietrozullo:
Hey @THARUN-S-R that is cool that you want to do this, though there is one change in your PR that would affect also users that do not need this feature and it is the fact that you pass the chat id as message id in the langchain messages, which is incorrect and could result in unexpected behaviour. I think we need to find a better solution to this. 

PS: I would love to know what you are building if you want to chat in discord.

THARUN-S-R:
Hi @pietrozullo  , Would love to have a chat with you regarding this and some more changes i have been working on. My discord ID is tharun_5972_67634. Kindly share your id so that we can connect on discord. Thanks...

pietrozullo:
Hey @THARUN-S-R you can join the discord server here, lets chat on the [questions](https://discord.com/channels/1359929700302061738/1361226988567920651) channel :) https://discord.gg/EyrfxnCc

harshithvh:
Hi @THARUN-S-R, this is a great use case! However, please note that if the user ends the session or if the Python process terminates, all in-memory chat history, including the chat ID, will be lost.
Have you implemented a storage mechanism for your use case?

Also, we currently have this feature (native storage mechanism) on our roadmap

Issue #262, opened 2025-09-08
💡 I wish mcp-use had ...


pietrozullo:
# 💡 I WISH MCP-USE HAD...
Hey yall,

This is a ticket to track a **wishlist of features, improvements, or ideas** you’d love to see in **mcp-use** 🚀  

---

### How to contribute:
- **COMMENT BELOW 👇** with your request 🔥  
- If we have any questions, we’ll follow up in comments / via DMs  
- **React with ❤️** on any request you’d also like to see  

---

P.S.: Come say hi 👋 in our [Discord](https://discord.com/invite/XkNkSkMz3V), we’d love to hear from you!

ver007:
I hope to display the number of times the MCP tool is called back by the large model and the actual number of calls, so that I can analyze whether there are any problems with the large model.

ver007:
I hope there will be an interface to get the current number of agents and their details, workflow process, so that I can create visualization effects separately.

sabapathi-1995:
It will be good to have Dictionary variable which will have tool name and their respective tools output for the mcp.run(#prompt#) . So it will easy to get any particular tool output. Currently AI is changing the  tool output in final answer .

tsotimus:
Oauth flow that uses [DCR](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#dynamic-client-registration) when using OAuth

dhruv-db:
I wish MCP-USE could allow us to plug and play MCP servers like [rube.app](https://rube.app/)
 making it easy to connect connectors and quickly build an MCP endpoint URL so users can integrate with any app using the MCP server.

pietrozullo:
@dhruv-db this is a feature request of the cloud product, not of the library. Right ?

dhruv-db:
Nope not just for cloud, we can do it on self hosted one as well. Then we can give more security and control to user.

pietrozullo:
@tsotimus 

> Oauth flow that uses [DCR](https://modelcontextprotocol.io/specification/2025-06-18/basic/authorization#dynamic-client-registration) when using OAuth

This is already available, where do you see it not being there ?

pietrozullo:
@sabapathi-1995 

> It will be good to have Dictionary variable which will have tool name and their respective tools output for the mcp.run(#prompt#) . So it will easy to get any particular tool output. Currently AI is changing the tool output in final answer .

[...]

PR #269, opened 2025-09-19
feat: Middleware

renvins:
# Pull Request Description

## Changes

This PR introduces a new middleware system for `mcp-use`. This system provides a flexible way to intercept and process MCP requests and responses, enabling features like logging, metrics, performance monitoring, and custom request/response handling.

A default logging middleware has been added and is enabled by default to provide basic request/response visibility. Additionally, a comprehensive set of metrics and analytics middleware is included for advanced monitoring.

## Implementation Details

1.  **Core Middleware Architecture (`mcp_use/middleware/middleware.py`)**:
    *   A new `Middleware` base class allows for creating custom middleware with strongly-typed hooks for different MCP methods (e.g., `on_call_tool`, `on_initialize`).
    *   `MiddlewareContext` provides a unified, typed context for all middleware operations.
    *   `MiddlewareManager` orchestrates the execution of the middleware chain.
    *   `CallbackClientSession` acts as an adapter to integrate the middleware system with the existing `ClientSession` without requiring changes to upstream callers.

2.  **Built-in Middleware**:
    *   `LoggingMiddleware` (`mcp_use/middleware/logging.py`): Provides default debug-level logging for all MCP requests and responses.
    *   `MetricsMiddleware`, `PerformanceMetricsMiddleware`, `ErrorTrackingMiddleware`, and `CombinedAnalyticsMiddleware` (`mcp_use/middleware/metrics.py`): A suite of middleware for collecting detailed metrics, performance data, and error analytics.

3.  **Integration**:
    *   `MCPClient` has been updated to accept a `middleware` argument in its constructor.
    *   The `default_logging_middleware` is automatically included.
    *   The middleware is passed down to the connectors and applied via the `CallbackClientSession`.

## Example Usage

```python
# After, you can pass custom middleware to the client.
# A default logging middleware is always active.
from mcp_use import MCPClient
from mcp_use.middleware import Middleware, MiddlewareContext, NextFunctionT
from mcp.types import CallToolRequestParams, CallToolResult

# Define a custom middleware
class CustomToolLogger(Middleware):
    async def on_call_tool(
        self, context: MiddlewareContext[CallToolRequestParams], call_next: NextFunctionT
    ) -> CallToolResult:
        params: CallToolRequestParams = context.params
        print(f"Intercepted tool call: {params.name} with args {params.arguments}")
        return await call_next(context)

# Initialize client with custom middleware
client = MCPClient(config=config, middleware=[CustomToolLogger()])

# ... now when a tool is called, the custom middleware will print the message.
```
The new `examples/example_middleware.py` provides a runnable example of this.

## Backwards Compatibility

These changes are fully backwards compatible. The `middleware` argument in `MCPClient` is optional. Existing code that does not use middleware will continue to work as before, with the only change being the addition of default debug logging for MCP requests.


pietrozullo:
This should allow this right https://github.com/mcp-use/mcp-use/issues/262#issuecomment-3268657790 ?

renvins:
> This should allow this right [#262 (comment)](https://github.com/mcp-use/mcp-use/issues/262#issuecomment-3268657790) ?

@pietrozullo yes! by intercepting the call. In theory with our performance middleware that should be possible!

mcp_use/__init__.py
```diff
@@ -35,6 +48,22 @@
...
+    "NextFunctionT",
+    # Default middleware
+    "default_logging_middleware",
+    # Middleware classes
+    "MetricsMiddleware",
+    "PerformanceMetricsMiddleware",
+    "ErrorTrackingMiddleware",
+    "CombinedAnalyticsMiddleware",
+    # Backwards compatibility
+    "create_metrics_middleware",
```
pietrozullo: outdated

mcp_use/__init__.py
pietrozullo, line 1: Why do we want to export this to the main module, do you see an advantage in doing 
`from mcp_use import Middleware` as opposed to `from mcp_use.middleware import Middleware` ?
At first sight, this looks like an advanced feature so one could do `mcp_use.middleware` and for the main module we just leave the very high level abstractions like MCPAgent and MCPClient. 

Anything against removing them from here ?  @renvins

renvins:
@pietrozullo If you approve, I merge!

mcp_use/__init__.py
renvins: No, I agree with you. It’s better do do mcp_use.middleware

renvins: @pietrozullo

pietrozullo: State if you are doing the change or if I should do it

renvins: Can you do it please? I’m busy right now

pietrozullo: Yes

PR #279, opened 2025-09-26
Fix http connector error handling

### LINKED ISSUE #278: Errors explanations are opaque and not clear when connecting or authenticating
**Describe the bug**
When there is an error during http connection (streamable or sse), errors were not clear. The problem could happen with failed authentication or failed connection.

**To Reproduce**
Just connect to an mcp server with http transport that is not started and you will see in console errors that are not clear

**Expected behavior**
Clear errors with correct explanation of why the error occurred.

**Screenshots**

Image

renvins:
# Pull Request Description

## Changes

This PR enhance errors' explanations that were opaque when encountering http errors during connection. That was happening because SSE runs behind a TaskGroup, and the MCP doesn't handle the exceptions as in the streamable and doesn't convert them to McpError, so exceptions are propagated to our code through ExceptionGroup (due to async TaskGroup). Now we're retrieving the last exception (in theory in our case is always one) and making it explicit.

It were also removed all the checks for fallback, because when Streamable HTTP transport fails, it always raise an McpError that doesn't contain further information about why the error happened and the status code (only custom status codes of MCP) and then the fallback is always tried (it was always tried also before). That's because in the MCP, streamable error handling is different: if something is wrong a JSONRPCError is generated and in the ClientSession after some checks, an McpError is raised.

Someone could ask why SSE doesn't raise McpError and directly http or connections errors. That's because of the nature of SSE and how it's implemented on MCP. Errors are not handled in try blocks so when for example httpx raises an HTTPStatusError, it is propagated to our code.

## Errors now
image

image

## Testing

*   **Unit Tests**: No new unit tests were added. The change corrects behavior that was not previously covered.
*   **Manual Testing**: I verified by running manual tests on the streamable transport with auth problems and connection problems.

## Backwards Compatibility

These changes are fully backwards compatible.

## Related Issues

Fixes #278

### DIFF
--- a/mcp_use/connectors/http.py
+++ b/mcp_use/connectors/http.py
@@ -203,6 +203,8 @@ class HttpConnector(BaseConnector):
                 else:
                     self._prompts = []
 
+            # Only McpError is raised from client's initialization because
+            # exceptions are handled internally.
             except McpError as mcp_error:
                 logger.error("MCP protocol error during initialization: %s", mcp_error.error)
                 # Clean up the test client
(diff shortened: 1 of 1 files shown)


renvins:
@pietrozullo Please review this fix and let me know. I had to put a lot of effort in it to fully understand the process behind!

PR #284, opened 2025-09-30
feat: Tools integrations for others SDKs [WIP]

renvins:
### Pull Request Description

This pull request introduces the first of our new integration adapters, allowing `mcp-use` to be used directly with other SDKs. This marks a significant step towards making `mcp-use` a universal tool provider that can plug into various LLM ecosystems beyond LangChain.

The initial part of this change is the `OpenAIMCPAdapter`, which handles the conversion of MCP tools into the format expected by OpenAI's tool-calling API. This allows developers to leverage the power of MCP servers while working directly with OpenAI's native tools.

### Key Changes:
*   **New `OpenAIMCPAdapter`**: Added `mcp_use/adapters/openai.py` to manage the conversion of MCP tools to the OpenAI format.
*   **Usage Example**: Included `examples/openai_integration_example.py` to demonstrate how to use the new adapter.
*   **Initial Documentation**: Created a basic documentation page at `docs/integration/openai.mdx`.

<details>
<summary><b>Usage Example (`examples/openai_integration_example.py`)</b></summary>

```python
import asyncio

from dotenv import load_dotenv
from openai import OpenAI

from mcp_use import MCPClient
from mcp_use.adapters import OpenAIMCPAdapter

# This example demonstrates how to use our integration
# adapaters to use MCP tools and convert to the right format.
# In particularly, this example uses the OpenAIMCPAdapter.

load_dotenv()


async def main():
    config = {
        "mcpServers": {
            "airbnb": {"command": "npx", "args": ["-y", "@openbnb/mcp-server-airbnb", "--ignore-robots-txt"]}
        }
    }

    try:
        client = MCPClient(config=config)

        # Creates the adapter for OpenAI's format
        adapter = OpenAIMCPAdapter()

        # Convert tools from active connectors to the OpenAI's format
        openai_tools = await adapter.create_tools(client)

        # Use tools with OpenAI's SDK (not agent in this case)
        openai = OpenAI()
        input_list = [{"role": "user", "content": "Search on Airbnb the cheapest hotel in Trapani for two nights."}]
        response = openai.chat.completions.create(model="gpt-4o", messages=input_list, tools=openai_tools)

        response_message = response.choices[0].message
        input_list.append(response_message)
        if not response_message.tool_calls:
            print("No tool call requested by the model")
            print(response_message.content)
            return

        for tool_call in response_message.tool_calls:
            import json

            function_name = tool_call.function.name
            arguments = json.loads(tool_call.function.arguments)

[...]

PR #288, opened 2025-10-05
fix: Return tool result in standard format

Pederzh:
# Pull Request Description

## Changes

This PR simplifies the tool result handling in the LangChain adapter by removing the complex `_parse_mcp_tool_result` method and directly returning the string representation of `tool_result.content`. 

**The main improvement is that tool results with multiple content elements (like resources) will now be properly formatted as JSON strings in MCP format, rather than being incorrectly concatenated or causing parsing errors.**

This change ensures proper MCP format compliance and fixes issues with resource display and MCP-UI compatibility.

## Implementation Details

1. **Removed complex parsing logic**: Eliminated the `_parse_mcp_tool_result` method that was manually handling different content types (text, image, resource) and their specific parsing logic
2. **Simplified return handling**: Changed from `return adapter_self._parse_mcp_tool_result(tool_result)` to `return str(tool_result.content)`
3. **Updated type annotations**: Changed return type from `Any` to `str | dict` for better type safety

## Example Usage (Before)

```python
# The old implementation had complex parsing logic
async def _arun(self, **kwargs: Any) -> Any:
    tool_result: CallToolResult = await self.tool_connector.call_tool(self.name, kwargs)
    try:
        # Complex parsing that handled different content types manually
        return adapter_self._parse_mcp_tool_result(tool_result)
    except Exception as e:
        return format_error(e, tool=self.name, tool_content=tool_result.content)

def _parse_mcp_tool_result(self, tool_result: CallToolResult) -> str:
    # 50+ lines of complex parsing logic for different content types
    # Handled text, image, and resource content separately
    # Could cause issues with multiple elements in content
```

## Example Usage (After)

```python
# The new implementation is much simpler and more reliable
async def _arun(self, **kwargs: Any) -> str | dict:
    tool_result: CallToolResult = await self.tool_connector.call_tool(self.name, kwargs)
    try:
        # Direct string conversion - handles all content types uniformly
        return str(tool_result.content)
    except Exception as e:
        return format_error(e, tool=self.name, tool_content=tool_result.content)
```

## Documentation Updates

* No documentation files were updated as this is an internal implementation change that maintains the same external API

## Testing

These changes were tested through:
- **Unit tests**: Existing tests continue to pass, confirming backward compatibility
- **Multi-element content**: Verified that tool results with multiple content elements are handled properly without the previous parsing issues

## Backwards Compatibility

These changes are **fully backwards compatible**:
- The external API remains unchanged. Tools still return string results
- Error handling behavior is preserved
- The change only affects internal implementation details
- Users will not need to modify their existing code

The main improvement is that tool results with multiple content elements (like resources) will now be properly formatted as JSON strings in MCP format, rather than being incorrectly concatenated or causing parsing errors.

## Related Issues

This change addresses issues with:
- Tool results containing multiple content elements not being properly formatted
- Resource display problems in MCP-UI compatible servers  

The simplified approach ensures that `tool_result.content` is always returned as a proper string representation, maintaining MCP protocol compliance while being more robust and maintainable.

### DIFF
--- a/mcp_use/adapters/langchain_adapter.py
+++ b/mcp_use/adapters/langchain_adapter.py
@@ -96,7 +96,7 @@ class LangChainAdapter(BaseAdapter):
                 """
                 raise NotImplementedError("MCP tools only support async operations")
 
-            async def _arun(self, **kwargs: Any) -> Any:
+            async def _arun(self, **kwargs: Any) -> str | dict:
                 """Asynchronously execute the tool with given arguments.
 
                 Args:
@@ -113,7 +113,6 @@ class LangChainAdapter(BaseAdapter):
                 try:
                     tool_result: CallToolResult = await self.tool_connector.call_tool(self.name, kwargs)
                     try:
-                        # Use the helper function to parse the result
                         return str(tool_result.content)
                     except Exception as e:
                         # Log the exception for debugging


Pederzh:
Same in the typescript library: https://github.com/mcp-use/mcp-use-ts/pull/52

pietrozullo:
FYI: @renvins this has conflicts with the integration PR https://github.com/mcp-use/mcp-use/pull/284

renvins:
@pietrozullo is that an update from MCP directly?

renvins:
@pietrozullo @Pederzh when you merge I have to update it in the integrations PR

PR #296, opened 2025-10-08
docs: update api reference

### LINKED ISSUE #290: docs: API reference is not populated
Would be nice to have these pages populated http://localhost:3000/api-reference/introduction, I am surprised mintlify does not have a way to do this automatically. One could use sphinx or something similar to do this.

harshithvh:
# Pull Request Description

## Documentation Updates

- Add missing MCPAgent methods (`stream_events`, `get_system_message`, `add_to_history`)
- Document remote agent support and structured output features
- Update MCPClient constructor and class methods with all parameters
- Add middleware support documentation

## Related Issues

Closes #290


pietrozullo:
Thanks for this, I think we need to fix the tests before merging anything anyway. Do you know what sphinx is ? It would be insanely cool to autogenerate the API docs based on the sphinx. I think this should be possible.

PR #311, opened 2025-10-16
Fix Edge Case Handling for Zapier MCP Servers 

### LINKED ISSUE #309: Zapier MCP Server “server-specific URL” triggers unexpected OAuth discovery
Kia ora koutou/G'Day everone,

Noticed a bug (or maybe a way on how we can handle Zapier MCP connections)

**Background**
Zapier has MCP server and anyone can pretty much tap into Zapier to do some stuff. Zapier offers two options - one via Direct/Specific URL and the other being oauth. 

Image

**Describe the bug**
When using [Zapier MCP Server](https://mcp.zapier.com/)
 with the server-specific URL (non-OAuth option), the MCP client (mcp-use) still performs OAuth discovery and attempts to register dynamically. This happens even when I explicitly set:

`auth: None
`
The result is that the client tries to open an OAuth browser flow, which fails in my setup. I’m deploying this on Google Cloud Run, so I want to avoid OAuth entirely and let the connection be seamless via the special URL Zapier provides.

As I was reading the pull request (yes..I did my homework) - https://github.com/mcp-use/mcp-use/pull/149 suggests that when  auth is None, mcp-use still tires its best to discover it (makes sense from an arhecitecture point of view). But in the case of Zapier, even if a direct URL is supplied that handles it, mcp-use still attempts to do it. Hence if there's a way to bypass it 

**Proposed Solution/Next Steps**

I’m happy to open a PR if this behaviour is indeed not supported yet. My thinking is:

If auth: None is explicitly set, the client should skip OAuth discovery entirely or similarily for how the issue for GitHub was being handled.....

Before I do that ......is there an existing way to configure mcp-use to bypass OAuth entirely for Zapier? Or is this currently unsupported and worth adding?

<<Other details are as below>>

**My Config**

```
MCP_SERVERS = {
    "weather": {
        "name": "🌤️ Weather Server",
        "description": "US Weather alerts and forecasts via NWS API",
        "url": os.getenv("MCP_SERVER_URL", "http://localhost:8000/mcp"),
        "tools": ["get_alerts", "get_forecast"],
    },
    "zapier": {
        "name": "⚡ Zapier Server",
        "description": "Zapier automation and CRM integration tools",
        "url": "https://mcp.zapier.com/api/mcp/s/XXXXXXXXXXXX==/mcp",
        "auth": None,
        "tools": ["xero_*", "activecampaign_*"],
    },
}
```

**Environment**

Client: mcp-use (latest from main) + streamlit 

Server: Zapier MCP (special URL mode, not OAuth)

Platform: Google Cloud Run

**Truncated Log from GCP**
```
2025-10-16 18:19:29.558 NZDT
2025-10-16 05:19:29,559 - mcp_use - INFO - Using default callback port 8080
2025-10-16 18:19:29.953 NZDT
2025-10-16 05:19:29,954 - mcp_use - INFO - OAuth authentication required
2025-10-16 18:19:29.954 NZDT
2025-10-16 05:19:29,954 - mcp_use - INFO - Attempting Dynamic Client Registration
2025-10-16 18:19:30.147 NZDT
GET2001.18 KB3 msChrome 141 https://XXXXXXapp/static/js/index.CMItVsFA.js
2025-10-16 18:19:30.455 NZDT
2025-10-16 05:19:30,455 - mcp_use - INFO - Dynamic Client Registration successful: nla-lcTIZrpk877XRidLEP3lgrlxbomCqs1LFZ0i
2025-10-16 18:19:30.571 NZDT
Opening browser for authorization: https://mcp.zapier.com/authorize?response_type=co.......
```

prasys:
# Pull Request Description

## Changes

Added special handling for Zapier MCP server endpoints to differentiate between OAuth-authenticated (`/a/`) and server-specific URL (`/s/`) connection types. 

(Types of connectivity with Zapier)
image

The OAuth discovery process now intelligently skips discovery for server-specific Zapier MCP server URLs while maintaining full OAuth flow support.

## Implementation Details

1. **Changes done on `_discover_metadata` method in `OAuth` class** (`libraries/python/mcp_use/auth/oauth.py`)
   - Added URL parsing logic to detect Zapier MCP server endpoints
   - `https://mcp.zapier.com/api/mcp/s/` endpoint (Pre-authenticated)**: Raises `OAuthDiscoveryError` so that it will be handled as no authentication  
   - `https://mcp.zapier.com/api/mcp/a/*` endpoint (OAuth)**: Proceeds with standard OAuth metadata discovery flow
   - Maintains existing edge case pattern (similar to GitHub API handling)

## Example Usage (Before)

```python
# Before: OAuth discovery would fail for Zapier /s/ endpoints with generic error
from mcp_use import MCPClient, MCPAgent

# Zapier
config = {
    "mcpServers": {
        "zapier": {
            "url": "https://mcp.zapier.com/api/mcp/s/XXXXXXXXX/mcp",
        }
    }
}
```
# This would attempt OAuth discovery

## Example Usage (After)
No change, the system will handle as if there is no authentication 

## Testing

I've performed manual testing on both the oauth and direct url with Zapier 

## Backwards Compatibility

N/A - it does not impact any existing functionality as it is mainly to handle Zapier edge case scenario 

## Related Issues

Closes #309


Pederzh:
@claude

tonxxd:
@claude

pietrozullo:
@prasys  thanks for the PR, sorry for taking so long but the authorization spec changes so fast that we cannot rely on hardcoding paths for specific clients, so I have been waiting for it (and our understand of it) to mature a bit. @renvins I think one pragmatic solution would be to allow people to send some of the auth configuration URLs from outside. Do you think it is feasible ?

prasys:
All good mate @pietrozullo - this is to make the app inclusive to all in one mcps like zapier and the likes of it. Just thinking out loud, wouldnt it be better for people to define their custom 'class' that they can put it in however they want it to be authorised or not...so instead of trying to auto-discover which may work for the majority of the cases, we can override it....

pietrozullo:
@prasys I agree!! Actually we have been talking about this but did not create an issue, i will do it now!
In fact this was the solution I proposed allowing customizations of this URLs so that it can accomodate all cases :) 

>one pragmatic solution would be to allow people to send some of the auth configuration URLs from outside.

PR #315, opened 2025-10-16
fix: use absolute path for changesets cwd parameter

tonxxd:
## Description

Fixes the TypeScript release workflow by using an absolute path for the changesets action `cwd` parameter.

## Problem
The changesets action was failing with:
```
Error: There is no .changeset directory in this project
```

## Solution
Changed the `cwd` parameter from relative path `libraries/typescript` to absolute path `${{ github.workspace }}/libraries/typescript`.

The `changesets/action` needs an absolute path to properly locate the `.changeset` directory in monorepo setups, as the action may run validation before switching to the specified working directory.

## Changes
- Updated `.github/workflows/typescript-release.yml` line 56

### DIFF
--- a/.github/workflows/typescript-release.yml
+++ b/.github/workflows/typescript-release.yml
@@ -53,7 +53,7 @@ jobs:
           version: pnpm run version
           commit: "chore(typescript): version packages"
           title: "chore(typescript): version packages"
-          cwd: libraries/typescript
+          cwd: ${{ github.workspace }}/libraries/typescript
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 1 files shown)


.github/workflows/typescript-release.yml
```diff
@@ -53,7 +53,7 @@ jobs:
           version: pnpm run version
           commit: "chore(typescript): version packages"
           title: "chore(typescript): version packages"
-          cwd: libraries/typescript
+          cwd: ${{ github.workspace }}/libraries/typescript
```
Pederzh, line 56: Great! Can you please add it also to `typescript-release-beta.yml`?

PR #283, opened 2025-09-30
feat: [WIP] pretty printer for agent run/stream 

harshithvh:
# Pull Request Description

## Changes

The Implementation includes visually appealing output formatting for agent responses (`run()` or `stream()`) using the `rich` library.

## Implementation Details

```
┌─────────────────────────────────────────────────────────────┐
│                      MCPAgent                               │
│                    (agents/mcpagent.py)                     │
└────────────┬────────────────────────────────────────────────┘
             │
             │ uses
             │
┌────────────▼────────────────────────────────────────────────┐
│                  Integration Layer                          │
│                (output/integration.py)                      │
│                                                             │
│  - wrap_stream_with_formatting()                           │
│  - format_and_print_result()                               │
│  - create_output_from_result()                             │
└────────────┬────────────────────────────────────────────────┘
             │
             │ uses
             │
┌────────────▼────────────────────────────────────────────────┐
│                    Formatter Layer                          │
│                 (output/formatter.py)                       │
│                                                             │
│  - format_output()                                         │
│  - format_stream()                                         │
│  - _format_* helpers                                       │
└──────┬─────────────────────────────────┬───────────────────┘
       │                                 │
       │ uses                            │ uses
       │                                 │
┌──────▼──────────────────┐   ┌─────────▼──────────────────┐
│   Printer Layer         │   │    Types Layer             │
│  (output/printer.py)    │   │  (output/types.py)         │
│                         │   │                            │
│  - OutputPrinter        │   │  - EventType               │
│  - StreamPrinter        │   │  - AgentOutputEvent        │
│                         │   │  - AgentOutput             │
└──────┬──────────────────┘   └────────────────────────────┘
       │
       │ uses
       │
┌──────▼──────────────────┐   ┌────────────────────────────┐
│   Rich Library          │   │   Configuration            │
│                         │   │  (output/config.py)        │
│  - Table                │   │                            │
│  - Console              │   │  - OutputConfig            │
│  - Live                 │   │                            │
│  - Markdown             │   └────────────────────────────┘
│  - JSON                 │
│  - Status               │
└─────────────────────────┘
```

### Directory Tree

```
mcp_use/
├── output/                          # output formatting module
│   ├── __init__.py                 # Public API exports
│   ├── types.py                    # Data structures (EventType, AgentOutputEvent, AgentOutput)
│   ├── formatter.py                # High-level formatting functions
│   ├── printer.py                  # Rich-based printer classes
│   ├── config.py                   # Configuration options
[...]

PR #358, opened 2025-10-21
feat: pretty printer for agent run/stream

harshithvh:
reopening: #283

### DIFF
--- a/libraries/python/mcp_use/agents/mcpagent.py
+++ b/libraries/python/mcp_use/agents/mcpagent.py
@@ -38,12 +38,12 @@ from mcp_use.agents.managers.server_manager import ServerManager
 
 # Import observability manager
 from mcp_use.agents.observability import ObservabilityManager
+from mcp_use.agents.output import OutputConfig, format_and_print_result, format_error, format_stream_with_panels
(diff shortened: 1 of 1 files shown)


libraries/python/mcp_use/output/formatter.py
```diff
@@ -0,0 +1,58 @@
+"""
+Output formatters for MCP agent responses.
+
+This module provides functions to format agent outputs
+using the rich library for terminal display.
+"""
+
+from typing import Any
+
+from .config import OutputConfig
```
pietrozullo: can we remove relative imports, I recently moved all imports to be absolute. (I know it makes sense here since it is only one layer above, but for consistency sake)

harshithvh:
> @harshithvh This was open before the langchain1.0.0 port, so there is a quite important conflict with MCPAgent from main. Could you resolve that ?

@pietrozullo resolved

pietrozullo:
Hey @harshithvh I have been using this printer and it seems to not be working properly 
image

pietrozullo:
I implemented a much simpler version of this which only adds one file to the package in #497, thanks @harshithvh for the inspiration though. If you think you can make the current system better happy to see it :)

PR #293, opened 2025-10-07
feat: interface for tool call management

harshithvh:
# Pull Request Description

## Changes

This PR implements Tool Call Management to enable conversation reconstruction by generating unique tool_call_id values that link Assistant Messages with Tool Messages. The implementation adds minimal functionality to track tool calls without introducing persistent memory management.

Message chain flow:
```
HumanMessage: "What's 2 + 3?"
AIMessage: [tool_call_id: "t1", tool_calls: [{"name": "calculator", "args": {"x": 2, "y": 3}}]]
ToolMessage: [tool_call_id: "t1", content: "5"]
AIMessage: "The answer is 5"
```

## Implementation Details

- Added `_generate_tool_call_id()` method using UUID for unique ID generation
- Added `_create_tool_message()` helper method for ToolMessage creation
- Updated `_arun()` method to generate `tool_call_id` before execution and create `ToolMessage` after execution
- `ToolMessage` will now be recorded into the existing conversation history

## Example Usage (Before)

```python
history = agent.get_conversation_history()
# [HumanMessage, AIMessage]
```

## Example Usage (After)

```python
history = agent.get_conversation_history()
# [ToolMessage(tool_call_id="call_abc12345"), HumanMessage, AIMessage]
```

## Documentation Updates

None

## Testing

- added tests in test_agent for tool call ID management
- added tests in test_adapter for LangChainAdapter integration

Validates ID generation, ToolMessage creation, and conversation history integration

## Backwards Compatibility

All existing code continues to work without modification, existing conversation history methods are unchanged

## Related Issues

Closes #276

PR #359, opened 2025-10-21
feat: interface for tool call management

harshithvh:
reopening: #293


renvins:
@harshithvh look at the comments of copilot, they seems good!

pietrozullo:
@harshithvh this is a good idea, but the current implementation isn't optimal, partially due to the library structure itself. this is why the PR is stale. doing this properly would require a deeper redesign, which isn't a priority right now since this feature wasn't requested by users as far as I know.

closing for now, but we might revisit this in the future if there's demand.

PR #382, opened 2025-10-23
Message ID inclusion to the conversation history

THARUN-S-R:
# Pull Request Description

## Changes

We are passing the Message Id variable into the conversation history to better identify the messages within the same session.

## Implementation Details

The BaseMessage of langchain has capabilities to accept the ID which we can use to identify the exact message incase we need to use it for training or feedback purposes

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -129,7 +129,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Fail if docs are outdated
-        if: steps.check_docs.outputs.docs_outdated == 'true' && github.event.pull_request.head.repo.full_name == github.repository
+        if: steps.check_docs.outputs.docs_outdated == 'true' && github.event.pull_request.head.repo.full_name == github.repository && github.event_name == 'pull_request' 
         run: |
           echo "Please update the API documentation and commit the changes."
           exit 1
--- a/libraries/python/mcp_use/agents/mcpagent.py
+++ b/libraries/python/mcp_use/agents/mcpagent.py
@@ -1037,7 +1037,11 @@ class MCPAgent:
                     for message in output:
                         if not isinstance(message, ToolAgentAction):
                             if self.message_id:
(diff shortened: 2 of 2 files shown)


.github/workflows/docs.yml
```diff
@@ -118,7 +129,7 @@ jobs:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Fail if docs are outdated
-        if: steps.check_docs.outputs.docs_outdated == 'true'
+        if: steps.check_docs.outputs.docs_outdated == 'true' && github.event.pull_request.head.repo.full_name == github.repository
```
harshithvh: I think this condition needs an extra check for `github.event_name == 'pull_request'` before accessing `github.event.pull_request.*` — otherwise it'll break on push events.

libraries/python/mcp_use/agents/mcpagent.py
```diff
@@ -1029,11 +1036,13 @@ async def _generate_response_chunks_async(
                 if isinstance(output, list):
                     for message in output:
                         if not isinstance(message, ToolAgentAction):
+                            if self.message_id:
+                                message["id"] = self.message_id
```
harshithvh: The `message["id"] = self.message_id` assumes message is always a dict, but it can be an object (like AIMessage). Add a quick `isinstance(message, dict)` check or use `setattr()` to avoid TypeErrors.

.github/workflows/docs.yml
pietrozullo: Remove all changes related to this file, they do not belong in the scope of the PR

pietrozullo: Are these changes related to the fact that this won't run correctly on PRs from forks ?

pietrozullo:
@THARUN-S-R, can we merge this ? Why is `message_id` and not `chat_id` or `session_id` ?

THARUN-S-R:
Hi @pietrozullo , will close this PR as the same changes are also included in the PR:
Feature: Changes for Runnable Config in Mcp-Use #395

PR #405, opened 2025-10-30
fix: Change default recursion limit

renvins:
# Pull Request Description

## Changes

A user issued a problem with default LangChain recursion limit. To solve it, I decided to set the default limit to max_steps * 2 that should be a value that fits most of the use-cases.

### DIFF
--- a/libraries/python/mcp_use/agents/mcpagent.py
+++ b/libraries/python/mcp_use/agents/mcpagent.py
@@ -21,7 +21,13 @@ from langchain.agents.middleware import ModelCallLimitMiddleware
 from langchain_core.agents import AgentAction
 from langchain_core.globals import set_debug
 from langchain_core.language_models import BaseLanguageModel
-from langchain_core.messages import AIMessage, BaseMessage, HumanMessage, SystemMessage, ToolMessage
+from langchain_core.messages import (
+    AIMessage,
+    BaseMessage,
+    HumanMessage,
+    SystemMessage,
+    ToolMessage,
+)
 from langchain_core.runnables.schema import StreamEvent
 from langchain_core.tools import BaseTool
 from pydantic import BaseModel
@@ -33,7 +39,10 @@ from mcp_use.agents.managers.server_manager import ServerManager
 # Import observability manager
 from mcp_use.agents.observability import ObservabilityManager
 from mcp_use.agents.prompts.system_prompt_builder import create_system_message
-from mcp_use.agents.prompts.templates import DEFAULT_SYSTEM_PROMPT_TEMPLATE, SERVER_MANAGER_SYSTEM_PROMPT_TEMPLATE
+from mcp_use.agents.prompts.templates import (
+    DEFAULT_SYSTEM_PROMPT_TEMPLATE,
+    SERVER_MANAGER_SYSTEM_PROMPT_TEMPLATE,
+)
 from mcp_use.agents.remote import RemoteAgent
 from mcp_use.client import MCPClient
 from mcp_use.client.connectors.base import BaseConnector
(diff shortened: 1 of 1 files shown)


renvins:
@pietrozullo let me know what you think!

pietrozullo:
How come the code is being reformmated like this ?

renvins:
@pietrozullo I just run ruff format before pushing and in nvim I have an automatic linter (when I save)

pietrozullo:
It does pass the linting checks in here, but this is not normal, do you have precommits ?
`uv pip install pre-commit`

renvins:
> It does pass the linting checks in here, but this is not normal, do you have precommits ? `uv pip install pre-commit`

Yes!

ilbertt:
I'm having issues with the `recursion_limit` parameter:

```python
async def main():
    # ... some setup code before

    client = MCPClient(config)

    # Create LLM
    llm = ChatOpenAI(model="gpt-5")

    agent = MCPAgent(
        llm=llm,
        client=client,
        max_steps=100,
    )

    # ... some other code after

if __name__ == "__main__":
    asyncio.run(main())
```

After a few tool calls, I get an error similar to this:
```
mcp_use - ERROR - ❌ Error running query: Recursion limit of 25 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.
For troubleshooting, visit: https://docs.langchain.com/oss/python/langgraph/errors/GRAPH_RECURSION_LIMIT

Traceback (most recent call last):
  File "./mcp-use-test/.venv/lib/python3.12/site-packages/mcp_use/agents/mcpagent.py", line 668, in stream
    async for chunk in self._agent_executor.astream(
  File "./mcp-use-test/.venv/lib/python3.12/site-packages/langgraph/pregel/main.py", line 3043, in astream
    raise GraphRecursionError(msg)
langgraph.errors.GraphRecursionError: Recursion limit of 25 reached without hitting a stop condition. You can increase the limit by setting the `recursion_limit` config key.
For troubleshooting, visit: https://docs.langchain.com/oss/python/langgraph/errors/GRAPH_RECURSION_LIMIT
```

According to:
https://github.com/mcp-use/mcp-use/blob/054084eedbdd79479ec3dadff99c9333b5e29797/libraries/python/mcp_use/agents/mcpagent.py#L129

the recursion limit should be `200`, since I'm configuring `MCPAgent` with `max_steps=100`, but apparently it's not picked up as the error shows `25`. Am I missing something?

ilbertt:
Oh, the reason may be that this fix has not been released yet

ilbertt:
I can confirm that installing the package from the GitHub `main` branch fixed the issue! It would be nice to have this patch released soon

virtualstaticvoid:
I experienced the `recursion_limit` issue today and found this ticket.

It would be nice if `MCPAgent` took an optional `config` parameter in it's `__init__` method, so any future config changes could just pass through to the underlying `langchain` `RunnableConfig` implementation.

humairshoukat:
@renvins @pietrozullo @Pederzh 

Can we reset the max_steps during agent execution?

pietrozullo:
Hey @humairshoukat  What do you mean reset ? You want to change it while executing ? Can you share a quick example of what you would like to do in pseudocode or similar ?

humairshoukat:
@pietrozullo  like I also get the above similar error even if I set max_steps to 100, because of long user chats. I want to know, how I can know the max_steps consumed of an agent during execution / allocation so that I can create a new one.

Currently I have created an agent pooling, where we create max 10 agents and assign them to users, using the app. but the issue is I get error "agent execution limit reached". We kill the unused agents after 5 mins window. But I want to know during agent execution, can I know how much max_steps are already being used. 

Because if you have asked 10 queries to an agent, behind the scenes it is calculating all the steps it did before for the prev queries. and when it reach the max step, we get the error. 

I'm in like this situation, can we have a call, where I can discuss this issue with you?

pietrozullo:
I think I understand that you run:
```python 
result1 = agent.run(user_message1) 
....
result2 = agent.run(user_message2)
```
and this is all happening in the same chat. So the cumulative steps can get as big as the length of the chat, and this creates issues. Right ? 

Originally we would track the `agent_steps` manually, this problem was introduced by the port to `langchain1.0.0` where we started using Middleware to limit the agent steps. We should fix this yeah.

humairshoukat:
@pietrozullo yes, but the thing is when multiple user come, we assign an agent from the pool and then run the agent async.

PR #408, opened 2025-10-31
fix(workflow): checkout step to handle PR refs correctly

harshithvh:
# Pull Request Description

## Changes

Updated the documentation workflow to fix checkout issues for pull requests.

## Implementation Details

Updated the Checkout code step to use a safer ref and repository pattern:
```yaml
fetch-depth: 0
ref: ${{ github.event.pull_request.head.ref || github.ref_name }}
```


pietrozullo:
Note for future readers: This is done so that it runs also on forks.
@harshithvh thanks 🫶 !

PR #422, opened 2025-11-03
fix: Add timeout option to stdio

### LINKED ISSUE #404: Unable to set timeout for connection to MCP Server
**Describe the bug**
For a remote MCP server the mcp_use API provides a timeout parameter (in http.py)
For a local MCP server the timeout parameter is missing (in stdio.py). 

**To Reproduce**
Connect to a local MCP server with takes a long time to load. In my case the server code is in Python and starts a very big program. It takes over a minute to start on the collegous computer. The MCP clients stops the connection after a while and reports a general error. The MCP server cannot be used then.

**Expected behavior**
No error message and proper connection to MCP server.

**Screenshots**

mcp{cli] 1.13.1
mcp-use 1.3.10
Python 3.12.8
Windows 11

**Additional context**
I searched in the code and found a timeout in
/mcp/shared/client.py:send_request(..)
which might be related. 

I would be very happy if someone can fix this.

renvins:
This pull request introduces a configurable `timeout` parameter for the `StdioConnector` and `StdioConnectionManager` to address issues with hardcoded process termination timeouts for stdio-based servers.

Previously, the timeout for terminating a stdio server process was not configurable through the `StdioConnector`, which could lead to premature process termination if a server required a longer-than-default time to shut down gracefully.

This change exposes the `timeout` setting in the `StdioConnector`, allowing developers to specify a custom duration to wait for the server process to exit. The provided timeout value is then used to override the global `PROCESS_TERMINATION_TIMEOUT` in the underlying `mcp` client library for the scope of the connection.

### Changes Made
*   Added a `timeout` parameter to `StdioConnector` with a default value imported from `mcp.client.stdio.PROCESS_TERMINATION_TIMEOUT`.
*   Propagated the `timeout` parameter from `StdioConnector` to `StdioConnectionManager`.
*   In `StdioConnectionManager`, the `timeout` is used to dynamically set the `mcp.client.stdio.PROCESS_TERMINATION_TIMEOUT` value, making the timeout configurable per connection.

This fix is a try to solve hardcoding... Fixes #404

### DIFF
--- a/libraries/python/mcp_use/client/connectors/stdio.py
+++ b/libraries/python/mcp_use/client/connectors/stdio.py
@@ -9,6 +9,7 @@ import sys
 
 from mcp import ClientSession, StdioServerParameters
 from mcp.client.session import ElicitationFnT, LoggingFnT, MessageHandlerFnT, SamplingFnT
+from mcp.client.stdio import PROCESS_TERMINATION_TIMEOUT
 
 from mcp_use.client.connectors.base import BaseConnector
 from mcp_use.client.middleware import CallbackClientSession, Middleware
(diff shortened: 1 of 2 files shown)


pietrozullo:
hey @renvins this PR is good but the solution we found for allowing this is quite hacky and since the user is not responding anymore I think we are better off not merging this code that might cause troubles if the official SDK changes the position/name of that variable.

renvins:
@pietrozullo Ok then let's close it for now.

PR #427, opened 2025-11-04
Feat/typescript/server side auth

pietrozullo:


docs/typescript/server/authentication.mdx
```diff
@@ -0,0 +1,654 @@
...
+When you call `server.use(middleware)`, the framework automatically wraps your middleware using a Proxy:
+
+**What gets wrapped:**
+- **MCP routes** (`/mcp`) → Runs through `AsyncLocalStorage.run()` to capture `req`/`res` context
+- **Your middleware** → Automatically skipped for excluded framework routes
+
+**What gets excluded:**
+- `/inspector` → Inspector UI (so you can debug without auth)
+- `/mcp-use/widgets` → Widget assets
+- `/__vite` → Vite HMR in development (dev mode only)
```
Pederzh, line 46: Here you mention that these routes are excluded but in `McpServer` we only exclude "/inspector".
In line 46: `private excludedMiddlewareRoutes = ["/inspector"];`
We might need to change the docs or the implementation.

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -39,6 +43,8 @@ export class McpServer {
   private serverPort?: number;
   private serverHost: string;
   private serverBaseUrl?: string;
+  private excludedMiddlewareRoutes = ["/inspector"];
+  private toolMiddleware: Map<string, any> = new Map();
```
Pederzh, line 46: `toolMiddleware` is not used anywhere.

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -1597,7 +1656,7 @@ if (container && Component) {
           zodType = z.object({});
           break;
         case "array":
-          zodType = z.array(z.any());
+          zodType = z.array(z.string());
```
Pederzh: This breaks support for arrays containing objects or other non-string types. 
Is there a reason we want to support only arrays of strings?

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -923,6 +970,12 @@ if (container && Component) {
       );
     });
 
+    widgets.forEach((widget) => {
+      console.log(
+        `[WIDGET] ${widget.name} mounted at ${baseRoute}/${widget.name}`
+      );
+    });
```
Pederzh: This is duplicated. Lines 967-971 and 973-977.

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -86,9 +92,37 @@ export class McpServer {
...
+          const value = (target as any)[prop];
+          return typeof value === "function" ? value.bind(target) : value;
         }
+
+        // Express methods - intercept 'use' to wrap middleware
+        if (prop === "use") {
+          return function (...args: any[]) {
+            // If the last argument is a middleware function (not a string path)
+            const lastArg = args[args.length - 1];
+            if (typeof lastArg === "function" && typeof args[0] !== "string") {
```
Pederzh: This condition excludes path-based middleware from wrapping:
```
// ❌ Won't be wrapped - args[0] is "/mcp" (string)
server.use('/mcp', passport.authenticate('jwt'))

// ✅ Will be wrapped - args[0] is the function
server.use(passport.authenticate('jwt'))
```
The documentation (authentication.mdx line 383) shows this usage pattern, which won't work with current implementation.
```
// Protect MCP endpoints
server.use('/mcp', passport.authenticate('jwt', { session: false }))
```

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -86,9 +92,37 @@ export class McpServer {
...
+              const originalMiddleware = lastArg;
```
Pederzh: The current code only wraps the last argument but app.use() it seems to support multiple middlewares as arguments in a single call per the Express middleware guide: https://expressjs.com/en/guide/using-middleware.html#middleware.application

Express supports:
1. Multiple middleware as arguments: app.use(mw1, mw2, mw3)
2. With path: app.use('/path', mw1, mw2, mw3)
3. Arrays: app.use([mw1, mw2])
4. Mixed: app.use('/path', mw1, [mw2, mw3])

There's this example where it shows how to use multiple middleware as an array (point 3 above).
```
function logOriginalUrl (req, res, next) {
  console.log('Request URL:', req.originalUrl)
  next()
}

function logMethod (req, res, next) {
  console.log('Request Type:', req.method)
  next()
}

const logStuff = [logOriginalUrl, logMethod]
app.get('/user/:id', logStuff, (req, res, next) => {
  res.send('User Info')
})
```

pietrozullo:
@Pederzh I think there is no super clean way to do that ignore path stuff, probably what we should do is let users specify they path manually (be clear about it in the docs) and if they want they can protect all paths and if not they can protect only the mcp path. What do you think ? 
To be more specific: 
`app.use` will intercept all paths with middleware. 
but one can do 
`app.use(/mcp,...)`
and this, for instance, will leave the inspector free.

pietrozullo:
PS: This might have unexpected consequences on widgets, so I will test this

tonxxd:
Closing in favor of https://github.com/mcp-use/mcp-use/pull/564

PR #478, opened 2025-11-18
feat: Add verify parameter to MCPClient and connectors

### LINKED ISSUE #452: MCP Use: Add Option to Disable SSL Verification (Like LangChain)
The Python script works perfectly with LangChain (SSL verification disabled via httpx.Client(verify=False)), but fails with MCP Use due to the same SSL error:
[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1010).
This error only occurs when using MCPAgent and blocks all interactions.

Run the script with MCPAgent and MCPClient.
The agent initializes and attempts to connect to the MCP server.
The SSL error appears, despite SSL being disabled for LangChain.

The script should allow disabling SSL verification for MCP Use, just like it does for LangChain, to enable execution in environments without valid certificates.

Desktop (please complete the following information):
OS: Windows
Python: 3.x
Libraries: langchain_mistralai, httpx, mcp_use

SSL verification is **successfully disabled** for LangChain using:
```
client_langchain = httpx.Client(
    verify=False,
    base_url=INSTANCE,
    headers={"Authorization": f"Bearer {API_KEY}"}
)
llm = ChatMistralAI(
    client=client_langchain,
    model=MODEL,
)
```

MCP Use **does not expose a parameter to disable SSL verification** in MCPClient or MCPAgent.
Question: Is there a way to disable SSL verification for MCP Use, or is this feature missing?

renvins:
## Summary
- Plumb the existing `verify` argument on `MCPClient` all the way through `HttpConnector`, so callers can turn TLS verification off (or supply a custom SSL context) when talking to development servers with self-signed certificates.
- Have `HttpConnector` build a custom `McpHttpClientFactory` that injects the client’s verification policy, and pass it into both the streamable HTTP and SSE connection managers.
- Update `StreamableHttpConnectionManager` and `SseConnectionManager` to accept the injected factory so every transport consistently instantiates `httpx.AsyncClient` with the caller’s TLS settings, including the OAuth discovery flow.

## Usage
Disable TLS verification for all HTTP servers loaded by the client:

```python
from mcp_use.client import MCPClient

client = MCPClient(
    config="/path/to/mcp.config.json",
    verify=False,
)
```

All streamable HTTP, SSE fallback, and OAuth discovery requests issued by this client will now inherit the same verification policy. Fixes #452

### DIFF
--- a/docs/python/api-reference/mcp_use_client_config.mdx
+++ b/docs/python/api-reference/mcp_use_client_config.mdx
@@ -37,7 +37,7 @@ from mcp_use.client.config import create_connector_from_config
 ><ParamField body="message_handler" type="mcp.client.session.MessageHandlerFnT | None" default="None" >   Parameter value </ParamField>
 ><ParamField body="logging_callback" type="mcp.client.session.LoggingFnT | None" default="None" >   Callback function </ParamField>
 ><ParamField body="middleware" type="list[mcp_use.client.middleware.middleware.Middleware] | None" default="None" >   Middleware instance </ParamField>
-><ParamField body="verify" type="bool | None" default="True" >   SSL certificate verification setting for HTTPS connections. Set to False to disable verification for development servers with self-signed certificates. </ParamField>
+><ParamField body="verify" type="bool | None" default="True" >   Boolean flag </ParamField>
 
 **Returns**
 ><ResponseField name="returns" type="mcp_use.client.connectors.base.BaseConnector" >A configured connector instance</ResponseField>


pietrozullo:
We use verify only as a boolean right ? @renvins

renvins:
Yes @pietrozullo

pyperanger:
The (no) verification isn't working correctly. Tested in an environment with version 1.5.0, it doesn't work properly. Could you please check this?

I'm still investigating, but it seems that something in the library's code is either overwriting or ignoring the value initially entered for Verify in some communications with the MCP servers.

PR #569, opened 2025-11-30
fix(inspector): use terser for smaller build

danstarns:
[NPM inspector](https://www.npmjs.com/package/@mcp-use/inspector) dist seemed unusually large and took a long time to 'instantly' use the `npx` command. 

Adding [terser](https://github.com/terser/terser) to compress the build further. Results locally show a 50% reduction in size in `packages/inspector/dist/`. 

Screenshot 2025-11-30 at 09 20 13

**Before** using `du -sh ./dist`: 

- 8.7mb

```sh
➜  inspector git:(main) ✗ du -sh ./dist 
8.7M    ./dist
```

**After** using `du -sh ./dist`: 

- 4.9mb

```sh
inspector git:(inspector-build-size) ✗ du -sh ./dist
4.9M    ./dist
```


tonxxd:
Thanks @danstarns that is a very good point 👍 
I ran some tests and the build setting `minify: true` as the same size of the one using `terser`. To avoid adding a new dep I changed from false to true instead. Merging into canary and releasing in main soon!

danstarns:
Screenshot 2025-12-01 at 06 33 17


@tonxxd much better 😅

PR #639, opened 2025-12-07
fix duplicate typescript codegroups in docs

Person-0:
# Fixed duplicate typescript codegroups in docs

The docs for TypeScript had duplicate codegroups for typescript in a lot of files with the same code appearing two times under the same name "TypeScript".
Fixed this be removing the duplicate codegroups.

### Before: 

image

### After: 

image


Person-0:
Some examples of the above mentioned issue:

https://mcp-use.com/docs/typescript/getting-started/quickstart#your-first-agent  >> FIXED (Replaced with *React*)
https://mcp-use.com/docs/typescript/agent/llm-integration#debug-model-behavior
https://mcp-use.com/docs/typescript/agent/building-custom-agents#basic-example

and so on..

tonxxd:
hey @Person-0 indeed this was an issue, thanks for reporting it! Since we did a bit of cleanup of the docs, i integrated the removal of the duplication in another pr so this should be solved now

PR #654, opened 2025-12-09
chore/typescript/package visibility

pietrozullo:
We want `typescript` packages to also link back to this repository so that we can see the number of dependents for them alongside the python package. This could be enough according to this [discussion](https://github.com/orgs/community/discussions/181457#discussioncomment-15205733)


pietrozullo:
Also @tonxxd we might need to do what I was suggesting yesterday about the releases https://github.com/orgs/community/discussions/181457#discussioncomment-15206739

PR #676, opened 2025-12-12
feat/typescript/release on publish

pietrozullo:


pietrozullo:
@tonxxd Can we merge this in canary to test if it works ? It will create tags and releases at each release (both canary and main) It should work, but to check it we need to merge in canary. It should be safe.

pietrozullo:
@tonxxd Particularly to test it we need a canary release, anything planned today you think ?

PR #679, opened 2025-12-12
chore(mcp-use): update package version and add client entry in package.json

tonxxd:
- Bumped version from 1.11.0-canary.7 to 1.11.0-canary.9 in package.json.
- Added client entry point in package.json for improved module exports.
- Updated tsup.config.ts to include src/client.ts in the build process.


pietrozullo:
ignore this comment

PR #721, opened 2025-12-23
feat: Add initialize request middleware

renvins:
## Summary
This PR extends the middleware system to support the `initialize` lifecycle event. This allows middleware to intercept, inspect, and potentially reject the initial MCP handshake (client capabilities, protocol version) before the session is fully established.

## Key Changes
- **Protocol Interception:** Implemented `MiddlewareServerSession` to override `_received_request`, capturing `initialize` calls at the protocol layer before they reach standard request handlers.
- **New Hook:** Added `on_initialize` support to the `Middleware` base class and dispatch logic.
- **Session Patching:** Patched `mcp.server.session.ServerSession` to inject the middleware manager during the handshake phase.

## Motivation
Standard request handlers in the MCP SDK do not expose the `initialize` event, as it is handled internally by the session manager. This limitation prevented middleware from:
- Validating client capabilities or metadata during the handshake.
- Logging connection attempts and protocol versions.
- Rejecting connections early based on initialization parameters.

## Example Usage

```python
class ConnectionGuard(Middleware):
    async def on_initialize(self, context, call_next):
        client_name = context.message.clientInfo.name
        print(f"Incoming connection from: {client_name}")
        
        # Example: Reject specific clients
        if client_name == "BannedClient":
            raise ValueError("Connection rejected")
            
        return await call_next(context)

### DIFF
--- a/docs/typescript/server/authentication/index.mdx
+++ b/docs/typescript/server/authentication/index.mdx
@@ -13,12 +13,12 @@ Add enterprise-grade OAuth 2.0/2.1 authentication to your MCP server with built-
 ### Basic OAuth Server
 
 ```typescript
-import { McpServer, auth0 } from 'mcp-use/server'
+import { MCPServer, oauthAuth0Provider } from 'mcp-use/server'
 
-const server = new McpServer({
+const server = new MCPServer({
   name: 'my-secure-server',
   version: '1.0.0',
-  oauth: auth0({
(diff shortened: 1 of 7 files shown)


pietrozullo:
@renvins hey, how come the tests are failing ?

renvins:
@pietrozullo I can't get that. What do you think?

PR #725, opened 2025-12-26
fix(python): resolve infinite Wait in ConnectionManager.stop()

kisharnath:


pietrozullo:
Hey @kisharnath you changed the lock file in the typescript library, remove that change please before I review.

pietrozullo:
In addition #726 solves the same, it is surprising you guys fixed the same issue so close 😄 
I will merge that PR which does not have the extraneous file modification and also is more complete. 

Thank you for contributing!

PR #797, opened 2026-01-12
feat: Add MCP Apps standard support for widgets

Pederzh:
## Summary

- Add MCP Apps standard (`text/html;profile=mcp-app`) support for widgets alongside existing Apps SDK support
- Introduce adaptor pattern for `useWidget` hook to support multiple host environments
- Add `mcp-apps` template to `create-mcp-use-app` for building MCP Apps compliant servers
- Remove deprecated `mcp-ui` template

### React Host Adaptors (`packages/mcp-use/src/react/host/`)
- **`mcp-app-adaptor.ts`**: Adaptor using `@modelcontextprotocol/ext-apps` for MCP Apps hosts
- **`apps-sdk-adaptor.ts`**: Adaptor for OpenAI Apps SDK hosts
- **`standalone-adaptor.ts`**: Adaptor for standalone/inspector environments
- **`types.ts`**: `WidgetHostAdaptor` interface with `getRevision()` for stable React state

### Server Widget Registration (`packages/mcp-use/src/server/`)
- Enhanced `mcp-server.ts` to differentiate between MCP Apps and Apps SDK metadata formats
- MCP Apps uses `ui.resourceUri` instead of OpenAI-specific `openai/*` fields
- Widget type stored in `mcp-use/widgetType` metadata for runtime differentiation

### Template Changes (`packages/create-mcp-use-app/`)
- **Added**: `mcp-apps` template with task-manager widget example
- **Removed**: `mcp-ui` template (deprecated in favor of `apps-sdk` and `mcp-apps`)

## Key Architectural Decisions

1. **Adaptor Pattern**: `useWidget` now detects the host environment and uses the appropriate adaptor, allowing widgets to work across different MCP-compatible hosts without code changes

2. **Metadata Separation**: Tool responses use different metadata formats based on widget type:
   - MCP Apps: `mimeType: "text/html;profile=mcp-app"`, `ui: { resourceUri: "..." }`
   - Apps SDK: `openai/outputTemplate`, `openai/widgetDescription`, etc.

3. **Widget-only metadata in widget files**: Widget `.tsx` files only define UI metadata (type, description, props). Tools are registered separately in `index.ts` using the `widget()` helper with `output` parameter.

### Creating an MCP Apps widget

```typescript
// resources/my-widget/widget.tsx
export const widgetMetadata: WidgetMetadata = {
  type: "mcpApp",
  description: "My widget using MCP Apps standard",
  props: propSchema,
};

// index.ts
server.tool(
  {
    name: "my-widget",
    schema: z.object({ data: z.string() }),
    widget: { name: "my-widget", invoking: "Loading...", invoked: "Ready" },
  },
  async (params) => {
    return widget({
      props: params,
      output: text(`Widget loaded with: ${params.data}`),
    });
  }
);
```

## Testing

- [x] `pnpm build` succeeds
- [x] `pnpm test_app:mcp-apps` - verified widget displays correctly
- [x] Verified tool responses include correct metadata for each widget type

## Backwards Compatibility

- **Breaking**: Removed `mcp-ui` template from `create-mcp-use-app`
- **Non-breaking**: Existing Apps SDK widgets continue to work unchanged
- **New**: `mcpApp` widget type for MCP Apps standard compliance

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **New Features**
  * MCP Apps template and Task Manager widget; unified multi-host support (Apps SDK, MCP App, Standalone); expanded host adaptor, widget, session and client APIs.

* **Documentation**
  * Added comprehensive MCP Apps template docs and a "Manual Testing with test_app" section with command examples.

* **Deprecations**
  * Removed legacy MCP UI template and its example Kanban widget.

* **Chores**
  * Renamed test script to test_app:mcp-apps and added ext-apps dependency.

* **Tests**
  * Added unit tests for host adaptor detection and behavior.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

---

<a href="https://app.devin.ai/review/mcp-use/mcp-use/pull/797">
  <picture>
    <source media="(prefers-color-scheme: dark)" srcset="https://static.devin.ai/assets/gh-open-in-devin-review-dark.svg?v=1">
    Open with Devin
  </picture>
</a>

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/widgets/adapters/apps-sdk-adapter.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/widgets/adapters/apps-sdk-adapter.ts
@@ -1,0 +1,362 @@
+/**
+ * Apps SDK Protocol Adapter
+ *
+ * Handles UI resource registration for OpenAI Apps SDK widgets.
+ * Uses text/html+skybridge MIME type and supports dual registration
+ * for both Apps SDK and MCP App hosts.
+ *
+ * @see https://developers.openai.com/apps-sdk/build/mcp-server
+ */
+
+import type {
+  AppsSdkUIResource,
+  UIResourceDefinition,
+} from "../../types/index.js";
+import type { UIResourceServer } from "../ui-resource-registration.js";
+import type {
+  UIResourceProtocolAdapter,
+  AdapterServerConfig,
+  UIResourceMeta,
+  AdapterToolResponse,
+} from "./types.js";
+import {
+  generateWidgetUri,
+  createWidgetUIResource,
+  type WidgetServerConfig,
+} from "../widget-helpers.js";
+
+/**
+ * Apps SDK Protocol Adapter
+ *
+ * Implements the UIResourceProtocolAdapter interface for OpenAI Apps SDK widgets.
+ * Handles:
+ * - text/html+skybridge MIME type
+ * - Dual registration (Apps SDK + MCP App)
+ * - Dynamic URI generation with random IDs
+ * - OpenAI-specific tool metadata
+ */
+export class AppsSdkProtocolAdapter implements UIResourceProtocolAdapter<AppsSdkUIResource> {
+  readonly type = "appsSdk" as const;
+  readonly mimeType = "text/html+skybridge";
+
+  /**
+   * Build tool metadata for Apps SDK widgets
+   *
+   * Includes OpenAI-specific fields like outputTemplate and invocation status.
+   */
+  buildToolMeta(
+    def: UIResourceDefinition,
+    serverConfig: AdapterServerConfig
+  ): Record<string, unknown> {
+    const definition = def as AppsSdkUIResource;
+    const resourceUri = this.generateResourceUri(
+      definition,
+      serverConfig.buildId
+    );
+    const toolMetadata: Record<string, unknown> = definition._meta || {};
+
+    // Add Apps SDK tool metadata
+    toolMetadata["openai/outputTemplate"] = resourceUri;
+
+    // Copy over tool-relevant metadata fields from appsSdkMetadata
+    if (definition.appsSdkMetadata) {
+      const toolMetadataFields = [
+        "openai/toolInvocation/invoking",
+        "openai/toolInvocation/invoked",
+        "openai/widgetAccessible",
+        "openai/resultCanProduceWidget",
+      ] as const;
+
+      for (const field of toolMetadataFields) {
+        if (definition.appsSdkMetadata[field] !== undefined) {
+          toolMetadata[field] = definition.appsSdkMetadata[field];
+        }
+      }
+    }
+
+    return toolMetadata;
+  }
+
+  /**
+   * Build UI resource metadata for storage in widgetDefinitions
+   *
+   * Includes contentHash if available for deterministic URI generation.
+   */
+  buildUIResourceMeta(
+    def: UIResourceDefinition,
+    _serverConfig: AdapterServerConfig
+  ): UIResourceMeta {
+    const definition = def as AppsSdkUIResource;
+    const widgetMeta = definition._meta?.["mcp-use/widget"] as
+      | { contentHash?: string }
+      | undefined;
+
+    if (definition._meta) {
+      return {
+        ...definition._meta,
+        "mcp-use/widgetType": this.type,
+        // Include contentHash at top level for easy access during tool invocation
+        "mcp-use/contentHash": widgetMeta?.contentHash,
+      };
+    }
+    return {
+      "mcp-use/widgetType": this.type,
+    };
+  }
+
+  /**
+   * Register resources and templates for Apps SDK widgets
+   *
+   * Registers:
+   * 1. Static resource (main widget)
+   * 2. Resource template (for dynamic URIs with random IDs)
+   * 3. MCP App resource (dual registration for compatibility)
+   * 4. MCP App resource template (dual registration)
+   */
+  registerResources(
+    server: UIResourceServer,
+    def: UIResourceDefinition,
+    serverConfig: AdapterServerConfig
+  ): void {
+    const definition = def as AppsSdkUIResource;
+    const resourceUri = this.generateResourceUri(
+      definition,
+      serverConfig.buildId
+    );
+    const widgetServerConfig = this.toWidgetServerConfig(serverConfig);
+
+    // 1. Register the main static resource
+    server.resource({
+      name: definition.name,
+      uri: resourceUri,
+      title: definition.title,
+      description: definition.description,
+      mimeType: this.mimeType,
+      _meta: definition._meta,
+      annotations: definition.annotations,
+      readCallback: async () => {
+        const uiResource = await createWidgetUIResource(
+          definition,
+          {},
+          widgetServerConfig
+        );
+        uiResource.resource.uri = resourceUri;
+        return { contents: [uiResource.resource] };
+      },
+    });
+
+    // 2. Register resource template for dynamic URIs
+    const buildIdPart = serverConfig.buildId ? `-${serverConfig.buildId}` : "";
+    const uriTemplate = `ui://widget/${definition.name}${buildIdPart}-{id}.html`;
+
+    server.resourceTemplate({
+      name: `${definition.name}-dynamic`,
+      resourceTemplate: {
+        uriTemplate,
+        name: definition.title || definition.name,
+        description: definition.description,
+        mimeType: this.mimeType,
+      },
+      _meta: definition._meta,
+      title: definition.title,
+      description: definition.description,
+      annotations: definition.annotations,
+      readCallback: async (uri: URL, _params: Record<string, string>) => {
+        const uiResource = await createWidgetUIResource(
+          definition,
+          {},
+          widgetServerConfig
+        );
+        uiResource.resource.uri = uri.toString();
+        return { contents: [uiResource.resource] };
+      },
+    });
+
+    // 3. DUAL REGISTRATION: Register MCP App resource for compatibility
+    this.registerMcpAppDualResources(server, definition, serverConfig);
+  }
+
+  /**
+   * Register dual MCP App resources for Apps SDK widgets
+   *
+   * This allows the same widget to work with both OpenAI Apps SDK
+   * and MCP Apps compliant hosts.
+   */
+  private registerMcpAppDualResources(
+    server: UIResourceServer,
+    definition: AppsSdkUIResource,
+    serverConfig: AdapterServerConfig
+  ): void {
+    const mcpAppUri = generateWidgetUri(
+      definition.name,
+      serverConfig.buildId,
+      "-mcp.html"
+    );
+    const mcpAppMimeType = "text/html;profile=mcp-app";
+    const widgetServerConfig = this.toWidgetServerConfig(serverConfig);
+
+    // Build MCP App specific metadata with CSP
+    const serverOrigin =
+      serverConfig.serverBaseUrl ||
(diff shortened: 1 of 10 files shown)


libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/README.md
```diff
@@ -0,0 +1,168 @@
...
+}
+```
+
+### MCP Apps Communication
+
+The widget uses `@modelcontextprotocol/ext-apps` for host communication:
+
+| Method | MCP Apps API |
+|--------|--------------|
+| `callTool(name, args)` | `app.callServerTool({ name, arguments })` |
```
gabrypavanello: Might worth to keep it as callServerTool for now, considering that we might have in the future something like callWidgetTool, so Widget that exposed to ChatGPT tools to interact with the UI. WDYT?

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
```diff
@@ -293,25 +277,46 @@ class MCPServerClass<HasOAuth extends boolean = false> {
...
+              responseMeta = {
+                mimeType: "text/html;profile=mcp-app",
+                "mcp-use/props": (result as any)._meta?.["mcp-use/props"] || {},
+                ui: {
+                  resourceUri,
+                },
+              };
+            } else {
+              // Apps SDK (default): use OpenAI-specific metadata
+              const uniqueUri = `ui://widget/${widgetName}${buildIdPart}-${randomId}.html`;
```
gabrypavanello: I would suggest to calculate an hash on the whole resource, so that you have a real versioning. 

**On operational POV:**
In case you deploy only server side logic, It's true that you wont have any issue, because the UI cached on OpenAI side it's exactly the same as the new deploy (just different resource URI).

**On DX POV**
You won't be able to recognize which UI version you're testing or you have deployed. At each run, it changes.

With the hash, this remains the same until you change something on the UI side.

**Performance Considerations:**
- The performance impact is on startup only if we run on containers. 
- On lambda/functions might have greater impact to measure

WDYT?

libraries/typescript/packages/mcp-use/src/server/widgets/ui-resource-registration.ts
gabrypavanello, line 1: I think all of this resource registration and its "binding" can be split in their own adapters, keeping all the composition logics of the different protocols truly separated (like done for the react adapters).
Conforming them to an interface where we expose the build of tool and resources metas. 
Like:`buildToolMeta` and `buildUIResourceMeta`

This makes it easy to extend, it reduces the cognitive load and it's better testable. WDYT?

libraries/typescript/packages/mcp-use/src/react/host/types.ts
```diff
@@ -0,0 +1,152 @@
...
+  // ─────────────────────────────────────────────────────────────────
+  // Subscriptions
+  // ─────────────────────────────────────────────────────────────────
+
+  /**
+   * Subscribe to host state changes
+   * @param callback - Called when any state changes
+   * @returns Unsubscribe function
+   */
+  subscribe(callback: () => void): () => void;
```
gabrypavanello, line 138: Maybe for another PR, but might worth no allow subscription for specific state changes (for instance: i might need in some parts to just observe the theme change). In this way i don't get notified for events that for the context are useless. WDYT?

libraries/typescript/packages/mcp-use/package.json
```diff
@@ -162,6 +162,7 @@
     "@langchain/anthropic": "^1.0.0",
     "@langchain/core": "^1.0.0",
     "@langchain/openai": "^1.0.0",
+    "@modelcontextprotocol/ext-apps": "^0.3.1",
```
gabrypavanello: There is already the 0.4.0 out :(

libraries/typescript/packages/mcp-use/src/react/host/types.ts
```diff
@@ -0,0 +1,152 @@
...
+  publicUrl?: string;
+}
+
+/**
+ * Common interface for host adaptors
+ *
+ * Each adaptor implements this interface to provide a unified API
+ * for widget communication regardless of the host environment.
+ */
+export interface WidgetHostAdaptor {
```
gabrypavanello, line 41: Nice! 🚀 ;)

libraries/typescript/packages/mcp-use/src/server/widgets/ui-resource-registration.ts
```diff
@@ -203,6 +211,153 @@ export function uiResourceRegistration<T extends UIResourceServer>(
         };
       },
     });
+
+    // DUAL REGISTRATION: Also register MCP App resource (text/html;profile=mcp-app)
```
gabrypavanello: **On Dual Registration Architecture:**
I understand the ergonomics of supporting both ChatGPT (text/html+skybridge) and MCP Apps (text/html;profile=mcp-app) from the same server instance simultaneously. However, I'm curious about the real-world advantage of this approach.

**Operational concerns:**
In production environments, when scaling solutions with multiple host integrations (VSCode only for now), teams typically want to keep connectors/partners separated for operational reasons:

- Independent deployments - Roll out changes to one integration without affecting others
- Targeted rollbacks - Revert a broken ChatGPT integration without touching MCP Apps
- Isolated blast radius - A bug in one adapter doesn't take down all integrations
- Different SLAs/MTTR - Partners may have different uptime requirements

Simpler alternative:
A configuration-based approach seems sufficient.

**Asymmetric ecosystem:**
Also worth noting: while MCP Apps can be supported by many hosts (VSCode only for now), the OpenAI Apps SDK is proprietary to ChatGPT. So in practice:

- text/html;profile=mcp-app → Many potential hosts
- text/html+skybridge → Only ChatGPT

This asymmetry suggests most production deployments will target one or the other, not both simultaneously.

**Question**: What use cases drove the dual registration design? Is there a real customer scenario where serving both from one server instance provides meaningful value over separate deployments?

libraries/typescript/packages/mcp-use/src/react/host/index.ts
```diff
@@ -0,0 +1,136 @@
...
+ * Detection priority:
+ * 1. Explicit override via window.mcpUse.hostType (injected by server)
+ * 2. window.openai presence -> apps-sdk
+ * 3. In iframe (window.parent !== window) -> mcp-app
+ * 4. Default -> standalone
+ *
+ * @returns The detected host type
+ */
+// Valid host type values for validation
+const VALID_HOST_TYPES: readonly HostType[] = [
```
gabrypavanello: In which scenario should i need to inject an hostType?

libraries/typescript/packages/mcp-use/src/server/widgets/mcp-ui-adapter.ts
```diff
@@ -176,14 +195,40 @@ export async function createRemoteDomResource(
...
 ): UIResourceContent {
   // For Apps SDK, we create the resource structure manually following the official pattern
   // from https://developers.openai.com/apps-sdk/build/mcp-server
-  const resource: any = {
+  let processedHtml = htmlTemplate;
+
+  // Inject host type if specified (for MCP Apps mode)
+  // Use JSON.stringify to escape values and prevent XSS/injection
+  if (hostType) {
+    const hostTypeScript = `<script>window.mcpUse = window.mcpUse || {}; window.mcpUse.hostType = ${JSON.stringify(hostType)};</script>`;
```
gabrypavanello: It would be nice to allow the dev to inject custom values.
**Reason**: runtime env vars needed by the widget (e.g. Google Maps API Key), since we don't want to persist secrets in the codebase
WDYT?

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
tonxxd: @gabrypavanello this makes a lot of sense, let me fix this

libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/README.md
tonxxd: Not sure i follow, do you have a pointer to the difference between these types of tools?

libraries/typescript/packages/mcp-use/src/server/widgets/mcp-ui-adapter.ts
tonxxd: love this, will add and document

libraries/typescript/packages/mcp-use/src/react/host/index.ts
tonxxd: @Pederzh ?

libraries/typescript/packages/mcp-use/src/server/mcp-server.ts
Pederzh: That's a really great suggestion! thanks @gabrypavanello

libraries/typescript/packages/mcp-use/src/server/widgets/ui-resource-registration.ts
Pederzh: Indeed! So instead of having all those select cases we'd have different adapters, one for **appsSdk** and one for **mcpApp**.

I think we also have to remove legacy `definition.types` like:
 ```
     case "externalUrl":
      resourceUri = generateWidgetUri(definition.widget, server.buildId);
      mimeType = "text/uri-list";
      break;
    case "rawHtml":
      resourceUri = generateWidgetUri(definition.name, server.buildId);
      mimeType = "text/html";
      break;
    case "remoteDom":
      resourceUri = generateWidgetUri(definition.name, server.buildId);
      mimeType = "application/vnd.mcp-ui.remote-dom+javascript";
      break;
 ```
 What do you think @gabrypavanello?

libraries/typescript/packages/mcp-use/src/react/host/index.ts
Pederzh: Good question! In fact `window.mcpUse `was introduced specifically for this feature. @gabrypavanello @tonxxd 

I've addressed it.
Rather than rely on the global injection, I've updated `createHostAdaptor()` to accept an explicit `hostType` option:

This is more explicit, type-safe, and testable than the global. The window.mcpUse.hostType injection remains as a fallback for server-injected HTML scenarios.

See commit e712491a.

libraries/typescript/packages/mcp-use/src/server/widgets/ui-resource-registration.ts
Pederzh: You're right that this doesn't match prod envs where isolation is preferred @gabrypavanello.
The implicit dual registration when using type: "appsSdk" could also surprise users who only want ChatGPT support.

Options to consider:
- 1. Remove dual registration: appsSdk = ChatGPT only, mcpApp = MCP Apps only. Users who want both can register the widget twice explicitly.
- 2. Make it explicit: Add dualRegistration: true option, default to false.

WDYT? @tonxxd

libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/README.md
gabrypavanello: It's not something existing yet (need to check), but I remember that I read somewhere a guy proposing to expose on widget side tools visible to the agent to act on the UI.
CallServerTool is very specific, and probably is named like this for that reason. I'll try to find a reference

libraries/typescript/packages/mcp-use/src/server/widgets/ui-resource-registration.ts
gabrypavanello: Sounds good!👌

PR #841, opened 2026-01-19
feat(server): Bearer Token authentication for Python MCP server

### LINKED ISSUE #840: [HELP] any way to control authorization for MCP server?
Hello  Team,
    is there any way to control authorization for MCP server?
    for example, enable API key for MCP server like http://127.0.0.1:8080/mcp?api_key=secret123

Thanks.

Best Regards,
Felix

pietrozullo:
## Summary

Add HTTP-layer bearer token authentication to the Python MCP server, allowing server developers to protect their endpoints with proper HTTP 401 responses.

## Changes

- `BearerAuthProvider` - Abstract base class with single `verify_token()` method
- `AccessToken` - Simple model with `token`, `claims`, and `scopes` fields  
- `AuthMiddleware` - Starlette middleware returning proper HTTP 401 for invalid/missing tokens
- `get_access_token()` / `require_auth()` - Context helpers to access auth in tools
- Documentation at `docs/python/server/auth/bearer.mdx`

## Usage

```python
from mcp_use.server import MCPServer, BearerAuthProvider, AccessToken, get_access_token

class MyAuthProvider(BearerAuthProvider):
    async def verify_token(self, token: str) -> AccessToken | None:
        if token != "secret":
            return None
        return AccessToken(token=token, claims={"user": "alice"})

server = MCPServer(name="my-server", auth=MyAuthProvider())

@server.tool()
def whoami() -> str:
    token = get_access_token()
    return f"Hello {token.claims.get('user')}" if token else "Not authenticated"
```

## Test Plan

- [ ] Test with valid bearer token returns 200
- [ ] Test with invalid token returns 401
- [ ] Test with missing token returns 401
- [ ] Test `get_access_token()` returns claims in tools
- [ ] Test excluded paths (`/docs`, `/inspector`) don't require auth

## Related

- Closes #840
- Linear: [MCP-1191](https://linear.app/mcp-use/issue/MCP-1191)

🤖 Generated with [Claude Code](https://claude.ai/code)

## Summary by CodeRabbit

* **New Features**
  * Optional Bearer token authentication for the Python MCP server (pluggable provider, middleware, context helpers) and MCPServer accepts an auth provider plus host/port configuration.

* **Documentation**
  * New docs and API reference pages for auth provider, middleware, token model, helpers, and updated host/port deployment guidance (default host 0.0.0.0, port 8000).

* **Tests**
  * Added unit and integration tests covering auth flows, middleware, helpers, and host/port behavior.

* **Chores**
  * Python dependency range for the MCP client library updated.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/docs/python/server/auth/bearer.mdx
+++ b/docs/python/server/auth/bearer.mdx
@@ -19,7 +19,8 @@ No additional code is needed in individual tools. The middleware validates the t
 ## Quick Start
 
 ```python
-from mcp_use.server import MCPServer, BearerAuthProvider, AccessToken, get_access_token
+from mcp_use.server import MCPServer
+from mcp_use.server.auth import BearerAuthProvider, AccessToken, get_access_token
 
 API_KEYS = {
     "sk-abc123": {"email": "alice@example.com"},
@@ -71,7 +72,7 @@ sequenceDiagram
 Extend this class and implement `verify_token()`:
 
 ```python
-from mcp_use.server import BearerAuthProvider, AccessToken
+from mcp_use.server.auth import BearerAuthProvider, AccessToken
 
 
 class MyAuthProvider(BearerAuthProvider):
@@ -148,7 +149,7 @@ Since the middleware already validates authentication, these helpers are for **a
 | `require_auth()` | Returns `AccessToken` or raises `AuthenticationError`. Use as a safeguard. |
 
 ```python
-from mcp_use.server import get_access_token, require_auth
+from mcp_use.server.auth import get_access_token, require_auth
 
 
 @server.tool()
--- a/libraries/python/mcp_use/server/auth/bearer.py
+++ b/libraries/python/mcp_use/server/auth/bearer.py
@@ -17,7 +17,8 @@ class BearerAuthProvider(ABC):
 
     Example:
         ```python
-        from mcp_use.server import MCPServer, BearerAuthProvider, AccessToken
+        from mcp_use.server import MCPServer
+        from mcp_use.server.auth import BearerAuthProvider, AccessToken
 
         class MyAuthProvider(BearerAuthProvider):
             async def verify_token(self, token: str) -> AccessToken | None:
--- a/libraries/python/mcp_use/server/auth/dependencies.py
+++ b/libraries/python/mcp_use/server/auth/dependencies.py
@@ -20,7 +20,7 @@ def get_access_token() -> AccessToken | None:
 
     Example:
         ```python
-        from mcp_use.server import get_access_token
+        from mcp_use.server.auth import get_access_token
 
         @server.tool()
         def whoami() -> str:
@@ -43,7 +43,7 @@ def require_auth() -> AccessToken:
 
     Example:
         ```python
-        from mcp_use.server import require_auth
+        from mcp_use.server.auth import require_auth
 
         @server.tool()
         def admin_action() -> str:
--- a/libraries/python/mcp_use/server/auth/models.py
+++ b/libraries/python/mcp_use/server/auth/models.py
@@ -23,7 +23,7 @@ class AccessToken(BaseModel):
 
     Example:
         ```python
-        from mcp_use.server import get_access_token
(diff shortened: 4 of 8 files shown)


docs/python/server/auth/bearer.mdx
```diff
@@ -0,0 +1,162 @@
...
+    claims={
+        "sub": "user-123",           # User ID
+        "email": "alice@example.com",
+        "name": "Alice",
+        "plan": "pro",               # Custom fields
+    },
+)
+```
+
+Access in tools via `token.claims.get("email")`.
```
renvins: Documentation is not totally clear. Especially this part should be self-explanatory. Think about improving it

libraries/python/mcp_use/server/auth/bearer.py
```diff
@@ -0,0 +1,46 @@
...
+    async def verify_token(self, token: str) -> AccessToken | None:
+        """Verify a bearer token and return user information.
+
+        Args:
+            token: The Bearer token from the Authorization header
+
+        Returns:
+            AccessToken with user claims if valid, None if invalid/expired
+        """
+        pass
```
renvins, line 47: Can be useful to have a default function, not only an abstract one.

libraries/python/mcp_use/server/auth/dependencies.py
renvins, line 1: Think about renaming this file, it's not totally related to what it does.

libraries/python/mcp_use/server/auth/models.py
renvins, line 1: Think also about renaming this file...

libraries/python/mcp_use/server/server.py
```diff
@@ -263,6 +271,25 @@ def streamable_http_app(self):
             pretty_print_jsonrpc=self.pretty_print_jsonrpc,
         )
 
+        # Add auth middleware if provider is configured
+        if self._auth:
+            app.add_middleware(
+                cast(type, AuthMiddleware),
```
renvins: Why this cast here? Is that needed?

libraries/python/tests/unit/server/auth/test_bearer_auth.py
renvins, line 1: I don't like very much this tests, it seems that some operations are repeated in multiple tests.

docs/python/server/auth/bearer.mdx
pietrozullo: Fixed in ab3350a2 - improved the claims section with explanation and code example.

libraries/python/mcp_use/server/auth/bearer.py
pietrozullo: Discussed - keeping as abstract only. A default implementation doesn't make sense since there's no sensible default behavior for token validation - users must implement their own logic.

libraries/python/mcp_use/server/auth/dependencies.py
pietrozullo: Discussed - keeping as `dependencies.py`. The name fits the dependency pattern (similar to FastAPI), and will be even more appropriate when we add `Depends()` support (tracked in MCP-1205).

libraries/python/mcp_use/server/auth/models.py
pietrozullo: Discussed - keeping as `models.py` for now. Consistent with Pydantic model convention. Could revisit in a broader refactor.

libraries/python/mcp_use/server/server.py
pietrozullo: Fixed in ab3350a2 - removed the unnecessary `cast()` since the import is now unconditional.

libraries/python/tests/unit/server/auth/test_bearer_auth.py
pietrozullo: Fixed in ab3350a2 - refactored tests to use pytest fixtures (`client`, `auth_provider`, `sample_token`, `reset_access_token`) to reduce repetition.

PR #914, opened 2026-01-29
fix(python): propagate verify parameter to HTTPConnector

### LINKED ISSUE #913: Verify parameter not always propagated to HTTPConnector
Related to PR https://github.com/mcp-use/mcp-use/pull/478 

I came across the same issue from this comment https://github.com/mcp-use/mcp-use/pull/478#issuecomment-3648093091

The `create_connector_from_config` is missing a `verify=self.verify` kwarg to propagate that to the  HTTPConnector. 

https://github.com/mcp-use/mcp-use/blob/bf8b806b7ee89b82403f131d439fb82fb0980053/libraries/python/mcp_use/client/client.py#L306

pietrozullo:
## Summary
- Fix the `verify` parameter not being propagated from `MCPClient` to `HTTPConnector` when creating sessions

## Implementation Details
1. Added `verify=self.verify` to the `create_connector_from_config` call in `MCPClient.create_session`
2. Updated test assertions to expect the `verify` parameter

## Testing
- All unit tests pass
- The fix ensures `verify=False` properly disables SSL certificate verification for HTTP-based MCP servers

## Backwards Compatibility
This is a bug fix with no breaking changes. The `verify` parameter already existed on `MCPClient` but was not being used.

## Related Issues
Closes #913

### DIFF
--- a/libraries/python/mcp_use/client/client.py
+++ b/libraries/python/mcp_use/client/client.py
@@ -312,6 +312,7 @@ class MCPClient:
             message_handler=self.message_handler,
             logging_callback=self.logging_callback,
             middleware=self.middleware,
+            verify=self.verify,
             roots=self.roots,
             list_roots_callback=self.list_roots_callback,
         )
--- a/libraries/python/tests/unit/test_client.py
+++ b/libraries/python/tests/unit/test_client.py
@@ -229,6 +229,7 @@ class TestMCPClientSessionManagement:
             message_handler=None,
             logging_callback=None,
             middleware=[default_logging_middleware],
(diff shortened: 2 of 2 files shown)


woprd:
Thank you @pietrozullo . I notice that the version number in `pyproject.toml` wasn't bumped (here https://github.com/mcp-use/mcp-use/pull/914/changes ), which meant the Github workflow for python release only created a draft release (here https://github.com/mcp-use/mcp-use/blob/main/.github/workflows/python-release.yml ). At what point will the fix be available for users of pypi?

PR #937, opened 2026-02-02
fix: prefix resource and prompt tool names to avoid collisions

### LINKED ISSUE #910: Inspector Chat fails with 400 when prompt and resource share the same name (tool name collision)
**Describe the bug**
When a resource and a prompt share the same name, the inspector's chat interface throws a 400 Bad Request error. The agent treats both as tools, causing a naming collision.

**To Reproduce**
1. Create an MCP server with a resource and a prompt that share the same name. (you can use `examples/typescript/server/basic/simple` to reproduce this since the prompt and resource are both called `greeting`)
2. Start the server.
3. Open the inspector, connect to the server, and navigate to the Chat tab.
4. Send any message in the chat.
5. See error: POST https://api.anthropic.com/v1/messages 400 (Bad Request)

**Expected behavior**
The chat should work normally. Resources and prompts should be able to share names without conflicts, or the server should validate and warn about duplicate names across different types.

**Screenshots**
Errors:
Image

Payload:
Image

 - OS: Any
 - Browser: Any
 - Version: Any

**Additional context**
The core issue is in libraries/typescript/packages/mcp-use/src/agents/mcp_agent.ts where tools, resources, and prompts are all combined into a single _tools array:
Line 453 (when using MCPClient): this._tools = [...tools, ...resources, ...prompts];
Line 481 (when using direct connectors): this._tools = [...tools, ...resources, ...prompts];
This merges all three types into one array, so duplicate names across types cause collisions when LangChain registers them.

gabrielekarra:
## Changes
Fix 400 Bad Request error in Inspector Chat when a resource and a prompt share the same name.

## Implementation Details
1. Prefix resource tool names with `resource_` in `LangChainAdapter.convertResource()`
2. Prefix prompt tool names with `prompt_` in `LangChainAdapter.convertPrompt()`
3. This ensures unique tool names when all three types are merged into a single array in `MCPAgent.initialize()`

## Backwards Compatibility
This is a breaking change for users referencing resource/prompt tool names directly — resource tools are now prefixed with `resource_` and prompt tools with `prompt_`.

## Related Issues
Closes #910

### DIFF
--- a/libraries/typescript/.changeset/fix-tool-name-collision-reserve.md
+++ b/libraries/typescript/.changeset/fix-tool-name-collision-reserve.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Fix tool name collisions between resources, prompts, and regular tools in LangChainAdapter. The `reserveName` method now checks whether the prefixed fallback name (`resource_<name>` / `prompt_<name>`) is itself already taken, falling back to a numeric suffix when needed. Prompt names are also now sanitized consistently with resource names.


tonxxd:
Thanks @gabrielekarra what do you think about making it dynamic? e.g. only prefix if there is a conflict?

PR #1126, opened 2026-03-03
fix(python): register debug routes when debug=True passed to run()

### LINKED ISSUE #1099: debug=True in run() does not register debug routes when server was created with debug=False
# Bug: `debug=True` in `run()` does not register debug routes when `MCPServer` was created with default `debug=False`

## Describe the bug

When calling `server.run(debug=True)` on an `MCPServer` instance that was created with `debug=False` (the default), the debug routes (`/docs`, `/inspector`, `/openmcp.json`) are never registered. The server starts without them despite `debug=True` being passed to `run()`.

## To Reproduce

```python
from mcp_use.server import MCPServer

# Server created WITHOUT debug=True
server = MCPServer(name="my-server")

# debug=True passed at run time — routes are NOT added
server.run(debug=True)
```

Attempting to access `/docs`, `/inspector`, or `/openmcp.json` returns **404**.

## Expected behavior

Passing `debug=True` to `run()` should register the debug routes (as documented), regardless of the value used when constructing the `MCPServer` instance.

## Root cause (for maintainers)

In `MCPServer.run()` (`server.py:474-476`), `self.debug_level` is set to `1` before `runner.run()` is called.

In `ServerRunner.run()` (`runner.py:91-93`), the branch that calls `_add_dev_routes()` checks `not self.server.debug` — but since `debug_level` was already promoted to `1`, `self.server.debug` returns `True` and the condition is **never satisfied**.

The route-adding code is effectively **unreachable dead code**.

## Additional context

- Affects `streamable-http` transport (the default).
- The `stdio` transport is **not** affected since it doesn't use HTTP routes.
- Setting `debug=True` at **construction time** works correctly.
- The `DEBUG` environment variable also works correctly as a workaround.

## Workaround

Pass `debug=True` at construction time instead of at `run()` time:

```python
server = MCPServer(name="my-server", debug=True)
server.run()
```

pietrozullo:
## Changes

Fixes debug routes (`/docs`, `/inspector`, `/openmcp.json`) not being registered when `debug=True` is passed to `run()` on a server created with `debug=False` (the default).

## Implementation Details

1. Moved the debug route registration logic into `MCPServer.run()` — when `debug=True` is passed, `_add_dev_routes()` is called and the Starlette app is rebuilt so the routes are included
2. Removed dead code in `ServerRunner.run()` that attempted the same fix but could never execute (the `debug_level` was already promoted before `ServerRunner.run()` was called, so the condition was always false)
3. Removed the `debug` parameter from `ServerRunner.run()` since it's no longer needed there

## Example Usage (Before — broken)
```python
server = MCPServer(name="my-server")
server.run(debug=True)  # /docs, /inspector, /openmcp.json returned 404
```

## Example Usage (After — fixed)
```python
server = MCPServer(name="my-server")
server.run(debug=True)  # debug routes now work correctly
```

## Documentation Updates
- Regenerated API reference for `ServerRunner.run()` (removed `debug` param)

## Testing
- Added `TestDebugRoutes` test class with 3 tests:
  - `debug=False` does not register dev routes
  - `debug=True` at init registers routes
  - `debug=True` at `run()` registers routes (regression test)

## Backwards Compatibility
- Removing `debug` from `ServerRunner.run()` is technically breaking if anyone called it directly, but `ServerRunner` is an internal API not intended for direct use

Closes #1099

### DIFF
--- a/docs/python/api-reference/mcp_use_server_runner.mdx
+++ b/docs/python/api-reference/mcp_use_server_runner.mdx
@@ -52,7 +52,6 @@ Run the MCP server.
 ><ParamField body="host" type="str" default='127.0.0.1' >   Host to bind to </ParamField>
 ><ParamField body="port" type="int" default="8000" >   Port to bind to </ParamField>
 ><ParamField body="reload" type="bool" default="False" >   Whether to enable auto-reload </ParamField>
-><ParamField body="debug" type="bool" default="False" >   Whether to enable debug mode </ParamField>
 
 **Signature**
 ```python wrap
@@ -60,8 +59,7 @@ def run(
 transport: Literal = "streamable-http",
     host: str = "127.0.0.1",
     port: int = 8000,
-    reload: bool = False,
-    debug: bool = False
+    reload: bool = False
 ):
 ```
 


pietrozullo:
Closing — this fix was already merged via #1114. This branch was stale.

Issue #1107, opened 2026-02-25
Add custom middleware hooks


pldesch-chift:
A very handy feature in Python's FastMCP package is to allow to add [middleware hooks](https://gofastmcp.com/servers/middleware) to change some specifics parts of the answer conditionally. Would love to see such a feature coming to MCP-use TS package. See this Discord [thread](https://discord.com/channels/1359929700302061738/1475832119791063190) for more context

PR #1137, opened 2026-03-04
RFC: MCP Operation-Level Middleware for TypeScript SDK

tonxxd:
> **This PR is open for discussion.** we'd love feedback on the API design before merging. Please comment with your thoughts, use cases, or alternative approaches.

---

### Motivation

A recurring request from the community — including a workaround shared in the [mcp-use Discord](https://discord.gg/XkNkSkMz3V) — is the ability to intercept MCP operations (tool calls, resource reads, prompt fetches) before they reach the handler. The current workaround involves reaching into the SDK's internal `_requestHandlers` map directly:

```typescript
// Community workaround — reaches into SDK internals
const handlerMap = (mcpServer.server as any)._requestHandlers;
const previous = handlerMap.get("tools/call");
handlerMap.set("tools/call", (req, extra) => middleware(req, extra, previous));
```

The use cases cited were: logging, scope-based auth guards, per-session rate limiting, and filtering tool lists dynamically.

Python's FastMCP addresses this with `on_call_tool` / `on_message` hooks. This PR proposes a TypeScript-native equivalent.

---

## Proposed API

A Hono-style `server.use()` call with an explicit `mcp:` prefix to distinguish MCP operation middleware from HTTP middleware:

```typescript
server.use("mcp:*", async (ctx, next) => { ... })          // catch-all
server.use("mcp:tools/call", async (ctx, next) => { ... }) // exact method
server.use("mcp:tools/list", async (ctx, next) => { ... }) // list filtering
```

The `mcp:` prefix routes the handler to the MCP middleware registry instead of Hono, so existing HTTP middleware usage is completely unaffected. `ctx` and `next` are **automatically typed** — no explicit type annotations needed.

### `MiddlewareContext`

```typescript
interface MiddlewareContext {
  method: string;                   // e.g. "tools/call"
  params: Record<string, unknown>;  // raw JSON-RPC params
  session?: { sessionId: string };  // active session, if any
  auth?: AuthInfo;                  // OAuth scopes/claims, if OAuth is configured
  state: Map<string, unknown>;      // per-request scratch space
}
```

### Supported patterns

| Pattern | Matches |
|---|---|
| `mcp:*` | Every MCP operation |
| `mcp:tools/call` | Tool calls |
| `mcp:tools/list` | Tool listing |
| `mcp:resources/read` | Resource reads |
| `mcp:resources/list` | Resource listing |
| `mcp:prompts/get` | Prompt fetches |
| `mcp:prompts/list` | Prompt listing |

---

## Implementation Details

1. **Pattern matching & composition** — `src/server/middleware/mcp-middleware.ts` — pure functions, no classes. `composeMiddleware()` builds the onion chain dynamically per-request.
2. **Proxy intercept** — `createHonoProxy` detects the `mcp:` prefix and routes to `_registerMcpMiddleware()` on `MCPServerClass` instead of Hono.
3. **Handler wrapping** — In `getServerForSession()`, every tool/resource/prompt callback is wrapped with `composeMiddleware`. List handlers (`tools/list`, etc.) are wrapped via `_wrapListHandlers`.
4. **HMR** — `this.mcpMiddlewares` is read dynamically inside closures, so hot-reloaded middleware is picked up without re-registering handlers.
5. **OAuth integration** — `ctx.auth` is populated from the existing `AsyncLocalStorage` request context, making scopes available to middleware with zero additional setup.
6. **Automatic typing** — `McpServerInstance` is intersected with `WithMcpUse`, which adds a template-literal overload `` `mcp:${string}` `` so TypeScript infers `ctx: MiddlewareContext` automatically.

---

## Open Questions for Discussion

These are the areas where we'd most value community input:

**1. Should `mcp:*` be the wildcard, or should we support method-prefix wildcards like `mcp:tools/*`?**
The current implementation supports only exact matches and `*`. A glob-style `tools/*` would cover `tools/call` + `tools/list` — useful, but adds complexity.

**2. Should middleware be able to mutate `ctx.params` before the handler sees them?**
Currently `ctx.params` is readable but mutation propagates to the handler. This is powerful (e.g., param sanitization) but may be surprising. Should we freeze params, or document mutation as intentional?

**3. Error handling behaviour**
Currently, if middleware throws, the MCP protocol converts it to `{ isError: true, content: [{ text: message }] }`. Should we expose a way to throw typed `McpError` with custom JSON-RPC error codes instead?

---

## Changes

- `src/server/middleware/mcp-middleware.ts` — new: `MiddlewareContext`, `McpMiddlewareFn`, `matchesPattern`, `composeMiddleware`
- `src/server/utils/hono-proxy.ts` — intercept `mcp:` prefix; export `WithMcpUse` for automatic typing
- `src/server/mcp-server.ts` — `mcpMiddlewares` array, `_registerMcpMiddleware()`, handler wrapping, HMR sync
- `src/server/index.ts` — exports for all new public types
- `tests/unit/server/middleware.test.ts` — 20 unit tests (pattern matching, composition, context, short-circuit, errors)
- `tests/integration/middleware.test.ts` — 11 end-to-end tests (tools, resources, prompts, list filtering, rejection)
- `examples/server/features/middleware/` — example server: logging, scope guard, rate limiter, tool filter
- `docs/typescript/server/middleware.mdx` — new documentation page

## Backwards Compatibility

Fully backward compatible. The `mcp:` prefix is only active when explicitly used; all existing `server.use(path, handler)` calls continue to route to Hono as before.

### DIFF
--- a/docs/docs.json
+++ b/docs/docs.json
@@ -596,6 +596,7 @@
                     "group": "Advanced Features",
                     "icon": "sparkle",
                     "pages": [
+                      "typescript/server/middleware",
                       "typescript/server/proxy",
                       "typescript/server/elicitation",
                       "typescript/server/sampling",
--- a/docs/typescript/server/middleware.mdx
+++ b/docs/typescript/server/middleware.mdx
@@ -1,0 +1,225 @@
+---
+title: Middleware
+description: Intercept and process MCP requests with middleware
+icon: "layers"
+---
+
+Middleware lets you intercept MCP operations to add logging, authentication, rate limiting, or any cross-cutting logic — without touching individual tool handlers.
+
+## Quick Start
+
+```typescript
+import { MCPServer, text } from "mcp-use/server";
+
+const server = new MCPServer({ name: "my-server", version: "1.0.0" });
+
+// Log every tool call
+server.use("mcp:tools/call", async (ctx, next) => {
+  console.log(`→ ${ctx.params.name}`);
+  const result = await next();
+  console.log(`← ${ctx.params.name}`);
+  return result;
+});
+
+server.tool(
+  { name: "greet", description: "Say hello", schema: z.object({ name: z.string() }) },
+  async ({ name }) => text(`Hello, ${name}!`)
+);
+
+await server.listen();
+```
+
+## How It Works
+
+MCP middleware uses an **onion model** — each middleware wraps the next, with the handler at the center.
+
+```
+Request
+  → Middleware A (before)
+    → Middleware B (before)
+      → Handler
+    ← Middleware B (after)
+  ← Middleware A (after)
+Response
+```
+
+Middleware is registered with `server.use('mcp:pattern', handler)` where the `mcp:` prefix distinguishes MCP middleware from HTTP middleware. Multiple middleware functions are composed in registration order (first registered = outermost).
+
+Each middleware can:
+- Inspect or modify the request before calling `next()`
+- Inspect or modify the response after `next()` returns
+- **Short-circuit** by returning early without calling `next()`
+- **Reject** by throwing an error
+
+## Patterns
+
+The pattern string (after `mcp:`) determines which operations trigger the middleware:
+
+| Pattern | Matches |
+|---------|---------|
+| `mcp:tools/call` | Tool executions only |
+| `mcp:tools/list` | Requests to list available tools |
+| `mcp:tools/*` | All tool operations |
+| `mcp:resources/read` | Resource reads |
+| `mcp:resources/list` | Requests to list resources |
+| `mcp:resources/*` | All resource operations |
+| `mcp:prompts/get` | Prompt fetches |
+| `mcp:prompts/list` | Requests to list prompts |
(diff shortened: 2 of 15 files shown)


tonxxd:
Related issue: https://github.com/mcp-use/mcp-use/issues/1107

tomjwxf:
I saw this RFC and built something in the same direction — a local proxy wrapper that adds per-tool policy and signed evidence without modifying the underlying MCP server.

The fast path:

    npx protect-mcp -- node your-server.js

Shadow mode logs every tool call. Add `--enforce` with a policy file to block, rate-limit, or require approval for specific tools. `protect-mcp init` generates local Ed25519 signing keys so every decision becomes an independently verifiable receipt.

The question this PR raises — middleware inside the runtime vs. proxy outside it — is one I've been testing. The proxy approach is simpler to adopt (zero code changes, works with any stdio server), but loses visibility into internal runtime state. For most operators, "what tools were called and what was the decision" is enough.

npm: https://www.npmjs.com/package/protect-mcp

Would be interested in whether the mcp-use team sees external proxy and internal middleware as complementary or competing approaches.

PR #1160, opened 2026-03-06
fix: replace examples symlinks with navigable READMEs

pietrozullo:
## Summary
- Replace broken symlinks in `examples/python` and `examples/typescript` with directories containing redirect READMEs
- Add comprehensive README files with example descriptions to the actual library example directories

## Problem
GitHub's web UI does not follow symlinks. Instead of navigating to the linked directory, it renders the symlink as a plain text file showing the target path (e.g., `../libraries/python/examples`). This is a [known GitHub limitation](https://github.com/dear-github/dear-github/issues/156).

## Changes
1. **Removed symlinks** (`examples/python` → `../libraries/python/examples`, `examples/typescript` → `../libraries/typescript/packages/mcp-use/examples`)
2. **Added redirect READMEs** in `examples/python/README.md` and `examples/typescript/README.md` with clickable links to the actual directories
3. **Added example index READMEs** to `libraries/python/examples/README.md` and `libraries/typescript/packages/mcp-use/examples/README.md` listing all examples with short descriptions in table format — GitHub renders these automatically when browsing

## Testing
- All 68 links in the READMEs validated by pre-commit hook
- Clicking `examples/python` or `examples/typescript` on GitHub will now show a README with a working link instead of a dead symlink path

## Backwards Compatibility
Non-breaking. Local clones that relied on the symlinks will see directories with READMEs instead, but the actual examples haven't moved.

Closes MCP-865

### DIFF
--- a/examples/python/README.md
+++ b/examples/python/README.md
@@ -3,5 +3,3 @@
 The Python examples live in the library directory:
 
 **[Browse Python Examples](../../libraries/python/examples/)**
-
-This includes client, server, and agent examples.
--- a/examples/typescript/README.md
+++ b/examples/typescript/README.md
@@ -3,5 +3,3 @@
 The TypeScript examples live in the library directory:
 
 **[Browse TypeScript Examples](../../libraries/typescript/packages/mcp-use/examples/)**
-
-This includes client, server, and agent examples.
--- a/libraries/python/examples/README.md
+++ b/libraries/python/examples/README.md
@@ -1,0 +1,57 @@
+# Python Examples
+
+## Client Examples
+
+| Example | Description |
+|---------|-------------|
+| [http_example.py](http_example.py) | HTTP client usage |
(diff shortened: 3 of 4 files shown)


pietrozullo:
## Root Cause

The `examples/python` and `examples/typescript` entries are **git symlinks** pointing to `../libraries/python/examples` and `../libraries/typescript/packages/mcp-use/examples` respectively.

GitHub's web UI **does not follow symlinks**. Instead of navigating to the target directory, it renders the symlink as a plain text file showing the path string. This is a [known limitation](https://github.com/dear-github/dear-github/issues/156) that has been requested since 2016 but never implemented.

The GitHub API confirms this — it reports `"type": "symlink"` with `"html_url"` pointing to `/blob/` (file view) rather than `/tree/` (directory view), so there is no workaround on GitHub's side.

**Fix:** Replace symlinks with directories containing redirect READMEs, and add proper example index READMEs to the actual library directories.

PR #1172, opened 2026-03-08
fix(cli): use serverFileUrl instead of serverFilePath in tsImport call

### LINKED ISSUE #1171: Bug: `mcp-use dev` crashes on Windows — `serverFilePath` used instead of `serverFileUrl` in `tsImport` call
### Environment
- **OS:** Windows (all drives — `C:`, `D:`, `A:`, etc.)
- **Node.js version:** v24.0.0
- **mcp-use CLI version:** `@mcp-use/cli v2.18.1`

---

### Description

`mcp-use dev` crashes immediately on all Windows machines. The server never starts. The same error also occurs during `npm install` via the `postinstall` script (`mcp-use generate-types`).

---

### Steps to Reproduce

1. On any Windows machine, run `npx create-mcp-use-app my-server`
2. `cd my-server && npm install`
3. `npm run dev`

---

### Error Output

```
Failed to start server: Only URLs with a scheme in: file, data, and node are supported
by the default ESM loader. On Windows, absolute paths must be valid file:// URLs.
Received protocol 'c:'

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:208:11)
    at defaultLoad (node:internal/modules/esm/load:103:3)
    ...
```

---

### Root Cause

In `dev.ts`, `serverFileUrl` is **correctly computed** using `pathToFileURL` right before `importServerModule` is defined:

```ts
const serverFilePath = path.join(projectPath, serverFile); // raw OS path
const serverFileUrl = pathToFileURL(serverFilePath).href; // ✅ file:///C:/Projects/...
```

However, inside `importServerModule`, the `tsImport` branch incorrectly uses the raw `serverFilePath` instead of the already-computed `serverFileUrl`:

```ts
// ❌ BUG: passes raw Windows path to tsImport
await tsImport(`${serverFilePath}?t=${Date.now()}`, {
  parentURL: import.meta.url,
  ...
});

// ✅ CORRECT: the else-branch already does it right
await import(`${serverFileUrl}?t=${Date.now()}`);
```

On Windows, `serverFilePath` is something like `C:\Projects\my-server\index.ts`. Node's ESM loader sees `c:` as the URL scheme and throws `ERR_UNSUPPORTED_ESM_URL_SCHEME`. This affects every Windows drive letter, not just non-`C:` drives.

The variable `serverFileUrl` containing the correct `file:///C:/...` format already exists in scope — it is just not used in the `tsImport` branch.

---

### Expected Behavior

`mcp-use dev` starts successfully on Windows.

### Actual Behavior

Immediate crash with `ERR_UNSUPPORTED_ESM_URL_SCHEME` on all Windows machines.

---

### Fix

One-line change in the `tsImport` branch of `importServerModule` — replace `serverFilePath` with `serverFileUrl`:

```ts
// Before (broken on Windows):
await tsImport(`${serverFilePath}?t=${Date.now()}`, {

// After (fix):
await tsImport(`${serverFileUrl}?t=${Date.now()}`, {
```

`serverFileUrl` is already in scope and already correctly formatted as a `file://` URL via `pathToFileURL`. No other changes needed.

CloveSVG:
## Problem

mcp-use dev crashes on Windows with ERR_UNSUPPORTED_ESM_URL_SCHEME. On Windows, serverFilePath is an absolute path like C:\Projects\my-server\index.ts. Node's ESM loader interprets the drive letter c: as a URL scheme and throws.

Fixes #1171

## Solution

In the importServerModule helper, the tsImport branch passes serverFilePath, but serverFileUrl (already computed via pathToFileURL) is the correct argument. The else branch already uses serverFileUrl correctly.

Before:
await tsImport(`${serverFilePath}?t=${Date.now()}`, { ... })

After:
await tsImport(`${serverFileUrl}?t=${Date.now()}`, { ... })

serverFileUrl is already in scope as pathToFileURL(serverFilePath).href, producing the correct file:///C:/... format on Windows.

## Testing

The else branch (native import) already uses serverFileUrl and works on Windows. This fix applies the same pattern to the tsImport branch.

### DIFF
--- a/libraries/typescript/packages/cli/src/index.ts
+++ b/libraries/typescript/packages/cli/src/index.ts
@@ -342,8 +342,10 @@ async function generateToolRegistryTypesForServer(
     (globalThis as any).__mcpUseHmrMode = true;
     (globalThis as any).__mcpUseLastServer = undefined;
 
+    const { pathToFileURL: pathToFileURL2 } = await import("node:url");
+    const serverFileUrl2 = pathToFileURL2(serverFile).href;
     const { tsImport } = await import("tsx/esm/api");
-    await tsImport(serverFile, {
+    await tsImport(serverFileUrl2, {
       parentURL: import.meta.url,
       tsconfig: path.join(projectPath, "tsconfig.json"),
     });


Pederzh:
Hey @Jaredw2289-svg thanks for opening a PR. 
The fix looks correct!

One thing: I think there's the same bug in `generateToolRegistryTypesForServer()` (line ~346). 

It also passes a raw OS path to `tsImport`:
```
  const serverFile = path.join(projectPath, serverFileRelative);
  // ...
  await tsImport(serverFile, {
```

This would also crash on Windows, I think. Could you apply the same fix for pathToFileURL() there as well?

PR #1191, opened 2026-03-13
feat(inspector): add editable MCP server aliases in connection settings

AielloChan:
## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

This PR adds editable MCP server aliases to the Inspector connection settings flow so users can rename saved servers for easier management. It also fixes an IME issue where pressing `Enter` during Chinese input composition could submit the connection form unexpectedly.

## Implementation Details

1. Added an `Alias` field to `ConnectionSettingsForm` and wired it through both the new-connection flow and the `Edit Connection Settings` modal.
2. Reused the existing saved `name` field for persistence instead of introducing a new storage schema, so aliases are preserved across add, edit, copy-config, paste-config, and refresh flows.
3. Introduced a shared `serverNames` utility to centralize display-name resolution and prefer user-defined aliases over server-reported metadata across Inspector surfaces.
4. Updated connected server rendering in dashboard tiles, server dropdowns, header client export actions, command palette entries, server info modal, server icon alt text, and the server list view.
5. Added an end-to-end test covering alias save and refresh persistence.
6. Updated keyboard handling in the connection form to ignore `Enter` while an IME composition session is active.

---

## TypeScript Checklist

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [x] `inspector`

### Pre-commit Checklist

- [ ] Ran `pnpm build` and build succeeds without errors

---

## Python Checklist

### Pre-commit Checklist

- [ ] Code formatted with `ruff format`
- [ ] Linting passes with `ruff check`
- [ ] Added or updated tests if needed

---

## Example Usage (Before)

```json
{
  "url": "http://localhost:3002/mcp",
  "transportType": "http"
}
```

In the Inspector UI, connected servers were effectively labeled by URL or server-reported metadata, and users could not assign a custom management name in `Edit Connection Settings`.

## Example Usage (After)

```json
{
  "url": "http://localhost:3002/mcp",
  "name": "QA Conformance",
  "transportType": "http"
}
```

In the Inspector UI, users can set an alias in `Edit Connection Settings`, and that alias is used consistently across connected server management views.

## Documentation Updates

No documentation files were updated in this PR.

## Testing

- Added an e2e test in `libraries/typescript/packages/inspector/tests/e2e/connection.test.ts` for alias save and refresh persistence.
- Ran `pnpm --filter mcp-use build`.
- Ran `pnpm --filter @mcp-use/cli build`.
- Ran `pnpm --filter @mcp-use/inspector type-check`.
- Verified that the alias-related changes do not add new type errors.
- Confirmed the remaining `type-check` failures are pre-existing Inspector issues unrelated to this PR, including unused parameters and an existing prompt typing mismatch in other files.
- Manually updated the connection form keyboard handling to guard against IME composition submission.

## Backwards Compatibility

These changes are backwards compatible. Existing saved connections continue to work unchanged. Aliases reuse the existing `name` field, and when no alias is configured the Inspector falls back to the previous metadata-based display behavior.

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/ConnectionSettingsForm.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/ConnectionSettingsForm.tsx
@@ -28,8 +28,6 @@ interface ConnectionSettingsFormProps {
   // Form state
   alias: string;
   setAlias: (value: string) => void;
-  transportType: string;
-  setTransportType: (value: string) => void;
   url: string;
   setUrl: (value: string) => void;
   connectionType: string;
@@ -89,8 +87,6 @@ interface ConnectionSettingsFormProps {
 export function ConnectionSettingsForm({
   alias,
   setAlias,
-  transportType: _transportType,
-  setTransportType: _setTransportType,
   url,
   setUrl,
   connectionType,
--- a/libraries/typescript/packages/inspector/src/client/components/InspectorDashboard.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/InspectorDashboard.tsx
@@ -969,8 +969,6 @@ export function InspectorDashboard() {
           <ConnectionSettingsForm
             alias={alias}
             setAlias={setAlias}
-            transportType="SSE"
-            setTransportType={() => {}}
             url={url}
             setUrl={setUrl}
             connectionType={connectionType}
--- a/libraries/typescript/packages/inspector/src/client/components/ServerConnectionModal.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/ServerConnectionModal.tsx
@@ -220,8 +220,6 @@ export function ServerConnectionModal({
         <ConnectionSettingsForm
           alias={alias}
           setAlias={setAlias}
-          transportType="SSE"
-          setTransportType={() => {}}
           url={url}
           setUrl={setUrl}
           connectionType={connectionType}


tonxxd:
Hey @AielloChan thanks for this PR! Makes sense to me, the only thing i would change is that right now changing the alias in the edit connection modal triggers a full reconnect. It might be better ux to simply change the name without triggering the reconnection, what do you think?

AielloChan:
> Hey @AielloChan thanks for this PR! Makes sense to me, the only thing i would change is that right now changing the alias in the edit connection modal triggers a full reconnect. It might be better ux to simply change the name without triggering the reconnection, what do you think?

I checked the current flow and you're right: editing the alias currently goes through `updateServer`, which always disconnects and remounts the connection, so an alias-only change triggers a full reconnect today.

I agree that alias changes should be treated as metadata-only and should not reconnect. I'll split that path so reconnect only happens when the actual connection config changes.

AielloChan:
> Hey @AielloChan thanks for this PR! Makes sense to me, the only thing i would change is that right now changing the alias in the edit connection modal triggers a full reconnect. It might be better ux to simply change the name without triggering the reconnection, what do you think?

I looked into the current implementation, and the scope is a bit broader than just changing the modal behavior.

Right now, editing the connection in the Inspector goes through `updateServer(...)`, and in `McpClientProvider` that update path always disconnects the existing server and remounts it. So alias edits reconnect not because the Inspector explicitly wants that, but because “update config” and “reconnect” are currently coupled at the provider level.

From a behavior perspective, I think alias-only changes should be treated as metadata-only updates, while URL / proxy / headers / transport changes should continue to reconnect.

There are two reasonable ways to implement that:

1. Inspector-local solution
   Detect when only the alias changed, and handle that separately inside the Inspector without calling the reconnecting update path.
   This is the smaller change, but it means the Inspector would need its own way to update persisted connection metadata.

2. Provider-level solution
   Add a separate non-reconnecting update path in `McpClientProvider` for metadata-only changes, and let the Inspector use that for alias updates.
   This is a cleaner model long-term because it explicitly separates:
   - metadata updates
   - connection-affecting updates
   But it does slightly expand the public React API surface.

So the main decision is whether you want the minimal Inspector-only fix, or the cleaner provider-level API split.

My preference is the provider-level split if you want the model to stay clean, and the Inspector-local fix if you want to keep this PR narrowly scoped.

tonxxd:
Indeed, i would also be in favor of the change on McpClientProvider

AielloChan:
> Indeed, i would also be in favor of the change on McpClientProvider

Makes sense. I’ll take the provider-level approach and split metadata-only updates from connection-affecting updates in `McpClientProvider`.

Then I’ll wire alias edits to the non-reconnecting path, while keeping reconnect behavior for URL / proxy / headers / transport changes, and add coverage for both cases.

AielloChan:
Implemented this with a provider-level split.

I added a new `updateServerMetadata(id, { name })` path in `McpClientProvider` for metadata-only updates, and kept `updateServer(...)` as the reconnecting path for connection-affecting changes. Then I wired the Inspector alias edit flow to use the metadata-only path when only the alias changes.

So after this change:
- alias-only edits update the display name without reconnecting
- URL / proxy / headers / transport changes still reconnect as before

I also added provider-level regression coverage for both cases, and updated the Inspector connection test to verify the renamed connection remains usable after the alias change.

PR #1226, opened 2026-03-23
fix(python): prevent double-close race in BaseConnector.disconnect()

### LINKED ISSUE #1220: Bug: `close_all_sessions()` closes each session twice — `ValueError: I/O operation on closed pipe` on Windows
**Describe the bug**
`MCPClient.close_all_sessions()` closes each session twice. On Windows this raises
`ValueError: I/O operation on closed pipe` immediately. On Unix it silently prints
`Exception ignored in: <function BaseEventLoop.__del__>` on process exit.

The root cause is that the `_closed` guard inside `MCPServerSession.close()` is set
**after** the first `await` on the transport, not before. Both the
`close_all_sessions()` loop and the `AsyncExitStack` teardown call `close()` — and
both pass the unguarded check before either one sets `_closed = True`.
```python
async def close(self) -> None:
    if self._closed:              
        return
    await self._transport.close() 
    self._closed = True           
```

**To Reproduce**
1. Create an `MCPClient` with any stdio-based MCP server (e.g. `@modelcontextprotocol/server-everything`)
2. Call `await client.create_all_sessions()`
3. Call any tool via `await session.call_tool(...)`
4. Call `await client.close_all_sessions()`
5. See `ValueError: I/O operation on closed pipe` on Windows (or noisy traceback on Unix)

Minimal reproducer:
```python
import asyncio
from mcp_use import MCPClient

async def main():
    client = MCPClient.from_dict({
        "mcpServers": {
            "calculator": {
                "command": "npx",
                "args": ["-y", "@modelcontextprotocol/server-everything"]
            }
        }
    })
    await client.create_all_sessions()
    session = client.get_session("calculator")
    await session.call_tool(name="add", arguments={"a": 1, "b": 2})
    await client.close_all_sessions()  # double-close triggers here

asyncio.run(main())
```

**Expected behavior**
`close_all_sessions()` shuts down each session exactly once with no error or traceback.

**Screenshots**
N/A — terminal error. Full traceback:
```
DEBUG - Closing session for server 'calculator'
DEBUG - Closing session for server 'calculator'
ValueError: I/O operation on closed pipe
  File "...\asyncio\windows_utils.py", line 102, in fileno
    raise ValueError("I/O operation on closed pipe")
```

On Unix:
```
Exception ignored in: <function BaseEventLoop.__del__ at 0x...>
RuntimeError: Event loop is closed
```

 - OS: Windows 11 (hard crash) / Ubuntu 22.04 (silent noise on exit)
 - Browser: N/A — Python script
 - Version: mcp-use 1.6.0

N/A

**Additional context**
The fix is a one-line change — move `self._closed = True` to **before** the first
`await` so the guard is effective for any concurrent or re-entrant call:
```python
# Before (broken):
async def close(self) -> None:
    if self._closed:
        return
    await self._transport.close()
    self._closed = True            # ← set too late

# After (fix):
async def close(self) -> None:
    if self._closed:
        return
    self._closed = True            # ← set BEFORE awaiting
    await self._transport.close()
```

`AsyncExitStack` teardown + the outer `close_all_sessions()` loop both call
`close()` — the fix makes the guard atomic with respect to re-entrant async calls..

Amrithesh-Kakkoth:
## Changes
Fix `close_all_sessions()` closing each session twice, which raises `ValueError: I/O operation on closed pipe` on Windows and noisy tracebacks on Unix.

## Implementation Details
1. Moved `self._connected = False` **before** `await self._cleanup_resources()` in `BaseConnector.disconnect()` so the guard is effective for concurrent/re-entrant async calls
2. Both `close_all_sessions()` and `AsyncExitStack` teardown call `disconnect()` — previously both could pass the `if not self._connected` check before either set the flag

## Example Usage (Before)
```python
await client.close_all_sessions()
# ValueError: I/O operation on closed pipe (Windows)
# Exception ignored in: <function BaseEventLoop.__del__> (Unix)
```

## Example Usage (After)
```python
await client.close_all_sessions()
# Clean shutdown, no errors
```

## Documentation Updates
- No documentation changes needed

## Testing
- Added `tests/unit/test_connector_double_close.py` with 3 tests:
  - Sequential double-call only cleans up once
  - Concurrent double-call (`asyncio.gather`) only cleans up once
  - Disconnect when not connected is a no-op
- All 301 existing unit tests continue to pass

## Backwards Compatibility
Not a breaking change — this is a one-line reorder that fixes incorrect behavior.

## Related Issues
Closes #1220

### DIFF
--- a/libraries/python/mcp_use/client/connectors/base.py
+++ b/libraries/python/mcp_use/client/connectors/base.py
@@ -5,6 +5,7 @@ This module provides the base connector interface that all MCP connectors
 must implement.
 """
 
+import asyncio
 import warnings
 from abc import ABC, abstractmethod
 from datetime import timedelta
@@ -95,6 +96,9 @@ class BaseConnector(ABC):
         # Roots support - always advertise roots capability
         self._roots: list[Root] = roots or []
         self._user_list_roots_callback = list_roots_callback
+
+        # Lock to ensure concurrent disconnect() callers wait for cleanup
+        self._disconnect_lock = asyncio.Lock()
 
         # Set up middleware manager
         self.middleware_manager = MiddlewareManager()
@@ -193,14 +197,15 @@ class BaseConnector(ABC):
     @telemetry("connector_disconnect")
     async def disconnect(self) -> None:
         """Close the connection to the MCP implementation."""
-        if not self._connected:
-            logger.debug("Not connected to MCP implementation")
-            return
+        async with self._disconnect_lock:
+            if not self._connected:
+                logger.debug("Not connected to MCP implementation")
+                return
 
-        logger.debug("Disconnecting from MCP implementation")
-        self._connected = False
-        await self._cleanup_resources()
-        logger.debug("Disconnected from MCP implementation")
+            logger.debug("Disconnecting from MCP implementation")
+            self._connected = False
+            await self._cleanup_resources()
+            logger.debug("Disconnected from MCP implementation")
 
     async def _cleanup_resources(self) -> None:
         """Clean up all resources associated with this connector."""
--- a/libraries/python/mcp_use/client/connectors/base.py
+++ b/libraries/python/mcp_use/client/connectors/base.py
@@ -198,8 +198,8 @@ class BaseConnector(ABC):
             return
 
         logger.debug("Disconnecting from MCP implementation")
-        await self._cleanup_resources()
         self._connected = False
+        await self._cleanup_resources()
         logger.debug("Disconnected from MCP implementation")
 
     async def _cleanup_resources(self) -> None:
--- a/libraries/python/tests/unit/test_connector_double_close.py
+++ b/libraries/python/tests/unit/test_connector_double_close.py
@@ -54,6 +54,26 @@ class TestDoubleCloseGuard:
         assert cleanup_call_count == 1
 
     @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


pietrozullo:
Good catch by Copilot — with the current fix, a second concurrent `disconnect()` returns immediately without waiting for cleanup to finish. Callers (e.g., context-manager teardown) could proceed assuming resources are fully released while cleanup is still in-flight.

The simplest fix is to use an `asyncio.Lock` so concurrent callers await the same disconnect rather than returning early:

```python
# In __init__:
self._disconnect_lock = asyncio.Lock()

# In disconnect():
@telemetry("connector_disconnect")
async def disconnect(self) -> None:
    """Close the connection to the MCP implementation."""
    async with self._disconnect_lock:
        if not self._connected:
            logger.debug("Not connected to MCP implementation")
            return

        logger.debug("Disconnecting from MCP implementation")
        self._connected = False
        await self._cleanup_resources()
        logger.debug("Disconnected from MCP implementation")
```

This way:
- The first caller acquires the lock, sets `_connected = False`, and runs cleanup
- Any concurrent caller blocks on the lock, then sees `_connected` is already `False` and returns
- No caller proceeds until cleanup is actually done

Note: `asyncio` will need to be added to the imports.

Amrithesh-Kakkoth:
awesome, made the changes and committed the changes

PR #1272, opened 2026-04-02
Remove Apps SDK support, keep MCP Apps only (closes #1269)

### LINKED ISSUE #1269: Remove apps sdk support
ChatGPT switched to MCP Apps support so there is not need to support both protocols. We should simplify the logic and keep only MCP Apps support

janhavitupe:
## Summary

Remove legacy ChatGPT Apps SDK support and simplify the client SDK to use MCP Apps as the single provider.

---

### Provider Simplification

* Removed `openai` / Apps SDK provider path from `useWidget`
* Removed `window.openai` detection and related logic
* Provider is now:

  * `mcp-apps` (when bridge is available)
  * `mcp-ui` (fallback via URL params)

### Widget API

* All widget actions now rely exclusively on MCP Apps bridge:

  * `callTool`
  * `sendFollowUpMessage`
  * `openExternal`
  * `requestDisplayMode`
  * `setState`
* Calls without MCP Apps bridge now throw explicit errors (no silent fallback)

### Model Context

* Removed legacy OpenAI model-context synchronization path

### Availability Logic

* `isAvailable` now depends only on MCP Apps bridge connection

### File API (`useFiles`)

* Removed Apps SDK file support
* `isSupported` is now always `false`
* `upload()` throws unsupported error
* `getDownloadUrl()` throws unsupported error

### Tests

* Updated tests to reflect MCP-only behavior
* `use-files.test.ts` now asserts unsupported behavior
* Removed expectations related to Apps SDK

---

## Before vs After

**Before**

* Dual provider support:

  * Apps SDK (`openai`)
  * MCP Apps

**After**

* Single provider:

  * MCP Apps only

---

## Behavior Changes (Breaking)

* Apps SDK (`openai`) integration is fully removed
* Widget actions will throw if MCP Apps bridge is not available
* File APIs are disabled (no MCP equivalent currently)

---

## Rationale

Maintaining dual support for Apps SDK and MCP Apps introduced:

* inconsistent behavior
* increased complexity
* divergence from current platform direction

This change aligns the SDK with MCP Apps as the single supported interface and removes legacy code paths.

---

## Scope

This PR is limited to:

* React hooks (`useWidget`, `useFiles`)
* client-side SDK behavior

It does NOT modify:

* server-side MCP protocol handling
* CLI or backend logic
* non-React packages

---

## Test Plan

```bash
cd libraries/typescript
pnpm install
pnpm -r build
pnpm --filter mcp-use test
```

Focused tests:

```bash
pnpm vitest run tests/unit/react/use-files.test.ts tests/unit/react/model-context.test.ts
```

---

## Notes

* Some unrelated tests may require:

  * built `dist` artifacts
  * `OPENAI_API_KEY` (for docs/examples)
* This PR intentionally removes Apps SDK-related paths and does not attempt to provide alternative file handling until MCP Apps exposes equivalent functionality.

## Related Issues

Closes #1269


Callme-VR:
like can you provide the how to find the folder to improve this issue

khandrew1:
Hi @janhavitupe, thanks for the PR! We should've been more specific in the issue but we also want to remove the OpenAI Apps renderer from the inspector and keep the `useFiles` hook as we already handle that gracefully if its not used in ChatGPT.

libraries/typescript/packages/mcp-use/src/react/useFiles.ts
```diff
@@ -94,71 +94,26 @@ export interface UseFilesResult {
...
+  const isSupported = false;
 
   const upload = useMemo(
     () =>
-      async (
-        file: File,
-        options: UploadOptions = {}
-      ): Promise<FileMetadata> => {
-        if (!isSupported) {
-          throw new Error(
```
khandrew1, line 112: We still want this hook supported in case they are in the ChatGPT host. We already have a fallback when its not available.

khandrew1:
@tonxxd Do you think we should also get rid of the adapter architecture now that MCP Apps is the standard?

tonxxd:
@khandrew1 I believe is a bit early to get rid of it since it might be helpful if other clients some up with their own implementation

khandrew1:
@janhavitupe Hey! Wanted to check in on the status of this PR

janhavitupe:
@khandrew1 I removed ChatGPT Apps SDK protocol support, keeping only mcp-apps and mcp-ui

libraries/typescript/packages/inspector/src/client/utils/widget-detection.ts
khandrew1, line 22: this function should be cleaned up. its just going to always return true and we only have one protocol currently.

libraries/typescript/packages/mcp-use/src/react/useFiles.ts
khandrew1: https://developers.openai.com/apps-sdk/reference#file-apis

khandrew1 (changes requested):
I realized forgot to submit my review. Sorry for the wait on this.

tonxxd:
Hi @janhavitupe thanks for your PR. This will be addressed cleanly in v2 #1827 please contribute there if you feel like!

PR #1328, opened 2026-04-15
feat(widget): stream tool output into widget props

### LINKED ISSUE #1296: Support streaming tool output into widget props
## Feature Request

Enable streaming data from tool execution into widget props, not just from model token streaming.

## Problem

Currently, widget props can only be streamed when the model is streaming tokens into the tool input — the widget receives partial input as the model generates it. However, there is no support for streaming data **from within the tool itself** back into the widget.

**Example scenario:** A tool receives a question as input, then calls a streaming API (e.g., an agent API) internally. The developer wants to stream the response from that API into the widget in real-time. Today, this is not possible — the widget only updates once the tool execution completes.

## Proposed Solution

Make widget props **streamable from the tool side** as well. This would allow developers to progressively update widget props during tool execution, enabling real-time UI updates for:

- Streaming responses from internal API calls
- Progressive data loading
- Long-running tool operations with incremental results

## Scope

TypeScript packages only.

hetsaraiya:
## Summary
- add `ctx.streamWidgetProps(...)` to stream partial widget props during tool execution
- add `tool-result-partial` handling to `mcp-use/react` and surface `partialToolOutput` / `isOutputStreaming` in `useWidget()`
- wire inspector Tools and client-side Chat widget hosts to forward partial tool output
- convert the `streaming-props` example into a working MCP server demo for tool-side widget prop streaming
- fix streamable HTTP session notification fallback so partial widget notifications are delivered during tool calls

Closes #1296.

## Verification
- `pnpm --dir libraries/typescript/packages/mcp-use build`
- `pnpm --dir libraries/typescript/packages/inspector type-check`
- `pnpm --dir libraries/typescript/packages/inspector build:client`
- manually verified the example server by calling `generate-code` and observing 10 `ui/notifications/tool-result-partial` updates before the final tool result

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/App.tsx
+++ b/libraries/typescript/packages/inspector/src/client/App.tsx
@@ -85,7 +85,7 @@ function App() {
           }}
           onSamplingRequest={(
             request,
-            serverId,
+            _serverId,
             serverName,
             approve,
             reject
@@ -116,9 +116,9 @@ function App() {
           }}
           onElicitationRequest={(
             request,
-            serverId,
+            _serverId,
             serverName,
-            approve,
+            _approve,
             reject
           ) => {
             const mode = request.request.mode || "form";
--- a/libraries/typescript/packages/inspector/src/client/components/IframeConsole.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/IframeConsole.tsx
@@ -221,7 +221,7 @@ const levelFilterColors: Record<ConsoleLogEntry["level"], string> = {
 };
 
 export function IframeConsole({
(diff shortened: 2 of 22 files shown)


hetsaraiya:
@tonxxd

khandrew1:
Hey @hetsaraiya! I want to apologize for the delay in reviewing this PR. Thanks so much for the contribution. After chatting with the maintainers of the MCP Apps spec, we understand now that using arbitrary notifications like this may actually be blocked in MCP hosts. Instead, we'll see if we can open a proposal in the official Spec for getting this functionality in. Closing.

PR #1332, opened 2026-04-15
feat(cli): drop-in MCP server inside Next.js apps via --mcp-dir

pietrozullo:
## Summary

Adds a `--mcp-dir` flag (and matching `mcp-use.config.json` `mcpDir` field) to `mcp-use dev/build/start` so a developer can drop a single `src/mcp/` folder into an existing Next.js repo and get a working MCP server + widgets — no sibling package, no restructure, no second codebase. Tools and widgets inside `src/mcp/` import directly from the rest of the Next.js app via the same `@/` alias the app already uses.

```
existing-nextjs-app/
├── package.json                # + mcp-use devDep, + mcp:* scripts
├── tsconfig.json               # UNCHANGED  (@/* → src/*)
├── next.config.js              # UNCHANGED
└── src/
    ├── app/                    # UNCHANGED
    ├── components/ui/          # REUSED by widgets
    ├── lib/assistant/tools.ts  # REUSED by server
    └── mcp/                    # the only new folder
        ├── index.ts            # MCPServer + tool registrations
        └── resources/
            └── contact_card/
                └── widget.tsx  # imports '@/components/ui/card'
```

Usage:

```bash
mcp-use dev   --mcp-dir src/mcp
mcp-use build --mcp-dir src/mcp
mcp-use start --mcp-dir src/mcp
```

## Changes

**`@mcp-use/cli`** (`libraries/typescript/packages/cli/`)
- `--mcp-dir <dir>` flag on `dev`, `build`, `start`. When set, defaults the server entry to `<dir>/index.ts` and the widgets dir to `<dir>/resources`. `mcp-use.config.json` supports `mcpDir` for the same effect with no flag.
- Widget Vite builds register `vite-tsconfig-paths` against the project's tsconfig when present, so widgets can `import '@/components/...'` and resolve through the app's own paths. Tailwind `@source` auto-extends to scan `<projectRoot>/src` so classes from shared components survive purging.
- **Auto-shim Next.js server-runtime modules** when `next` is detected in package.json. Covers `server-only`, `client-only`, `next/cache`, `next/headers`, `next/navigation`, `next/server`. Ships both an ESM loader (for `import`) and a CJS `Module._resolveFilename` patch (for the `require()` calls tsx emits when compiling TS in a non-`type: module` package). Active on both the HMR in-process path and the `tsx watch` subprocess path (the latter via `NODE_OPTIONS=-r ... --import=...`).
- Loads Next.js's env cascade (`.env` / `.env.development` / `.env.local` / `.env.development.local`) so tools transitively imported from the app find their env without the user mirroring values by hand.
- Replaces `tsx`'s `tsImport` (which registers tsx with a generated namespace, causing tsx's resolver to bail on transitive `@/*` imports) with a namespace-less `tsx/esm/api.register` + `tsx/cjs/api.register` pair, then native dynamic `import()` with `?t=` cache-busting on reload. This is what finally makes `@/lib/...` imports resolve in HMR mode.
- Adds `vite-tsconfig-paths@^5.1.4` as a CLI dep.

**`mcp-use`** (`libraries/typescript/packages/mcp-use/`)
- Both `mountWidgets` call sites in `MCPServer` now read `process.env.MCP_USE_WIDGETS_DIR` (set by the CLI when `--mcp-dir` is used) as the default for `resourcesDir`. User code is unchanged — the env var is the contract between CLI and runtime.
- `mount-widgets-dev.ts` loads `vite-tsconfig-paths` from the user's project when a root `tsconfig.json` exists, and installs it at the front of the dev Vite server's plugin chain. The hardcoded `"@": resourcesDir` alias only fires as a fallback for projects without a tsconfig (e.g. JS-only scaffolds). Tailwind `@source` auto-extends to scan `<projectRoot>/src` under the same conditions.

## Test plan

- [ ] `mcp-use dev --mcp-dir src/mcp` from a Next.js 16 + React 19 app starts, HMR is active, transitive `@/lib/...` imports resolve.
- [ ] `server-only` / `next/cache` / `next/headers` shim in transitive imports without user-side shim files.
- [ ] `.env.local` is auto-loaded.
- [ ] `tools/list` returns widget tools with full Apps SDK metadata (`openai/outputTemplate`, `openai/widgetCSP`, etc.).
- [ ] `tools/call` returns `structuredContent` with widget props plus the human-readable `output` text.
- [ ] `/mcp-use/widgets/<name>/` serves the widget HTML and the React bundle resolves `@/components/...` through the project's tsconfig paths.
- [ ] `mcp-use build --mcp-dir src/mcp` produces a self-contained widget bundle that pulls in the project's real components and Tailwind classes.
- [ ] Existing single-purpose mcp-use projects (without `--mcp-dir`) continue to work — defaults are unchanged.

## Notes

- Validated end-to-end against an existing Next.js 16 + React 19 + Tailwind v4 + shadcn customer app. Widget renders item-brain's real `Card` / `CardHeader` / `CardTitle` styled by item-brain's Tailwind.
- The CLI changes are additive: passing no `--mcp-dir` keeps the legacy "entry at root + `resources/` at root" behavior intact.
- Auto-detection of Next.js and the corresponding shim install is silent for non-Next projects (no `next` in package.json → nothing happens).


pietrozullo:
## Template repo for the drop-in pattern

End-to-end reference implementation of this PR is live at **[mcp-use/next-js-mcp-app-template](https://github.com/mcp-use/next-js-mcp-app-template)** (enabled as a GitHub template, so "Use this template" works). It's the standard `create-next-app --ts --tailwind --app --src-dir` output with one new folder (`src/mcp/`), two shared files (`src/lib/greetings.ts`, `src/components/GreetingCard.tsx`), three `mcp:*` scripts, and a README. Same service + same component powers both the Next.js home page and the MCP widget. While this PR is pre-release, the template pins `mcp-use` and `@mcp-use/cli` to the [pkg.pr.new](https://pkg.pr.new) snapshots for #1332.

## CI + review feedback addressed in 16122ab3

**Prettier / lint** — ran `prettier --write` on the three flagged files. `pnpm format:check` is now clean.

**Copilot feedback:**

| # | File | Fix |
|---|------|-----|
| 5 | `mcp-server.ts` | Forward `resourcesDir` only when `MCP_USE_WIDGETS_DIR` is set. Code now matches the comment's intent — leaving it unset lets `mountWidgets` apply its own default. |
| 6 | `project-config.ts` (`loadProjectConfig`) | Distinguish ENOENT from parse errors. Malformed JSON throws `Invalid JSON in <path>: <reason>` instead of being silently swallowed. |
| 7 | `project-config.ts` (`resolvePort`) | Validate each candidate with `Number.isInteger(n) && 0 ≤ n ≤ 65535`. Invalid values fall through to the next source instead of leaking `NaN`. |
| 3 | `mount-widgets-dev.ts` | `vite-tsconfig-paths` now falls back to the copy bundled with `@mcp-use/cli` when `userProjectRequire` can't find it in the user's project (pnpm / strict-hoisting projects that don't install it directly). |
| 4 | `mount-widgets-dev.ts` | Extracted the Tailwind `@source "<src>"` computation into a module-level `computeProjectSrcSourceLine` helper; both CSS-generation sites call it now. Fixes the indentation issue + deduplicates. |
| 2 | new: `tests/project-config.test.ts` | 23 unit tests covering every precedence level of `loadProjectConfig`, `resolveEntryFile`, `resolveWidgetsDir`, `resolveMcpDir`, `resolvePort` — including malformed JSON, missing files, mcpDir defaults, and NaN port values. |

**Copilot feedback I pushed back on:**

> #1 — \`HMR reload path relies on \`await import(…?t=Date.now())\` and claims that "transitive modules also re-evaluate"…\`

The suggestion — swap the native dynamic `import()` for `tsx/esm/api.tsImport` — is what this PR explicitly moves away from in commit c180f30c. Here's why the current code is correct:

1. `tsImport` registers tsx with a generated namespace (see `tsx/dist/register-B7jrtLTO.mjs`). The resolve hook then bails via `return r(t, e)` for any specifier whose parentURL doesn't carry the matching `tsx-namespace=` query. Transitive imports like `@/lib/assistant/tools` never hit tsx's `resolveTsPaths`, and the `@/*` alias is never rewritten. We verified this was the source of the `Cannot find package '@/lib'` error against a real Next.js app.
2. We switched to a namespace-less `tsx/esm/api.register({tsconfig})` + `tsx/cjs/api.register()` pair, then native dynamic `import()` with `?t=Date.now()` cache-busting. With no namespace, tsx's resolver runs for **every** specifier and consults the tsconfig paths matcher uniformly.
3. On transitive re-evaluation: when the entry URL is cache-busted, tsx's load hook receives a fresh invocation for it. Inside tsx's resolver, the resolved URL of each transitive dep gets augmented with the parent's query segment (see `resolve` in `tsx/dist/esm/index.mjs` — `n.url += "?" + i` after resolution), so the cache-bust propagates to the dependency graph. We verified this end-to-end: edit any file under the tree and the server reloads with the new code, no stale state.

The trade-off is real — this is a deliberate departure from `tsImport`'s pattern — but switching back breaks the feature this PR ships. Happy to add a doc-comment explaining it if reviewers prefer, but I'd rather not revert the code itself.

PR #1337, opened 2026-04-15
fix(react): restore window.fetch interceptor when useMcp unmounts (Via Proxy → Direct)

### LINKED ISSUE #1334: Inspector: Switching from "Via Proxy" → "Direct" fails with "Protected resource does not match" error
## Describe the bug

In the inspector, if you connect to a server using **Via Proxy**, then switch the connection type back to **Direct** (without removing/re-adding the server), the direct connection fails with a red dot and an error like:

```
Streamable HTTP connection failed: Streamable HTTP failed: Protected resource http://localhost:3000/inspector/api/proxy does not match expected https://mcp.linear.app/mcp (or origin)
```

The error does not occur if you start fresh with "Direct" — it only appears after having used "Via Proxy" first in the same browser session.

## Root Cause

`BrowserOAuthClientProvider.installFetchInterceptor()` (in `packages/mcp-use/src/auth/browser-provider.ts`) globally replaces `window.fetch` with an interceptor that routes all OAuth discovery requests (`/.well-known/…`) through the inspector's OAuth proxy. The proxy rewrites the `resource` field in protected resource metadata to point to the proxy URL (`http://localhost:3000/inspector/api/proxy`).

A `restoreFetch()` method exists to undo this, but **it is never called anywhere in the codebase**. When you switch back to "Direct", `updateServer()` (in `McpClientProvider.tsx`) unmounts the old `McpServerWrapper` and mounts a new one, but neither the explicit `clearStorage()` call nor the `useMcp` unmount cleanup ever calls `restoreFetch()`. The stale interceptor stays on `window.fetch`.

The new "Direct" `McpServerWrapper` does not install a fresh interceptor (no `gatewayUrl`), so `window.fetch` remains the old proxy interceptor. When the MCP SDK performs OAuth discovery against the direct server URL, the interceptor silently routes the request through the OAuth proxy, which rewrites `resource` to `http://localhost:3000/inspector/api/proxy`. The SDK's `selectResourceURL` then throws the mismatch error.

## To Reproduce

1. Open the inspector and add a server that requires OAuth (e.g. `https://mcp.linear.app/mcp`)
2. Set connection type to **Via Proxy** and click Connect — authentication succeeds
3. Open the connection settings (gear icon) and switch to **Direct**, save
4. The connection fails with a red circle; the error reads:
   `Protected resource http://localhost:3000/inspector/api/proxy does not match expected https://mcp.linear.app/mcp (or origin)`

## Expected behavior

Switching back to "Direct" should establish a direct connection. The stale OAuth proxy fetch interceptor should be torn down when the "Via Proxy" connection is torn down, so it doesn't interfere with the new direct connection.

## Relevant files / lines

| File | What's happening |
|---|---|
| `packages/mcp-use/src/auth/browser-provider.ts:102` | `installFetchInterceptor()` replaces `window.fetch` |
| `packages/mcp-use/src/auth/browser-provider.ts:259` | `restoreFetch()` exists but is never called |
| `packages/mcp-use/src/react/useMcp.ts:2020–2046` | Unmount cleanup — `restoreFetch()` is not called here |
| `packages/mcp-use/src/react/McpClientProvider.tsx:1228–1236` | `updateServer` calls `clearStorage` but not `restoreFetch` |
| `packages/mcp-use/src/server/middleware/mcp-proxy.ts:323–354` | Proxy rewrites `resource` in OAuth discovery metadata to the proxy URL |

## Desktop

- OS: macOS
- Browser: Chrome
- Inspector version: canary

## Additional context

The bug does not reproduce on a fresh page load that starts with "Direct" — it only affects in-session mode switches. Reloading the inspector page clears the stale `window.fetch` override.

A fix would add a `restoreFetch()` call in `useMcp`'s unmount cleanup (`useMcp.ts` around line 2020) so the interceptor is torn down whenever the `McpServerWrapper` for a proxied connection unmounts.

khandrew1:
# Pull Request Description

## Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

When a user connects via **Via Proxy** in the inspector and then switches to **Direct** (without reloading), the direct connection fails with:

```
Protected resource http://localhost:3000/inspector/api/proxy does not match expected https://mcp.linear.app/mcp (or origin)
```

`BrowserOAuthClientProvider.restoreFetch()` existed to undo the `window.fetch` proxy interceptor, but was never called anywhere. This PR calls it in the `useMcp` `useEffect` cleanup so the interceptor is torn down whenever the hook unmounts.

## Implementation Details

1. Added `restoreFetch?: () => void` to the `UseMcpAuthProvider` interface in `useMcp.ts` so TypeScript recognises the method
2. Added `authProviderRef.current?.restoreFetch?.()` to the `useEffect` cleanup (before `disconnect()`) — runs on every unmount, is a no-op when no interceptor was installed
3. Added unit test `useMcp-proxyCleanup.test.tsx` verifying `restoreFetch()` is called on unmount
4. Added `tests/**/*.test.tsx` to the vitest `include` pattern (`.tsx` test files were silently excluded before) and added `// @vitest-environment jsdom` to both `.tsx` test files that exercise `useMcp` directly

---

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [x] `mcp-use` (client)
- [ ] `mcp-use` (server)
- [ ] `inspector`

### Pre-commit Checklist

---

## Example Usage (Before)

```
// 1. Connect inspector via "Via Proxy" → succeeds
// 2. Switch to "Direct" → fails with:
//    Protected resource http://localhost:3000/inspector/api/proxy
//    does not match expected https://mcp.linear.app/mcp (or origin)
```

## Example Usage (After)

```
// 1. Connect inspector via "Via Proxy" → succeeds
// 2. Switch to "Direct" → succeeds (stale fetch interceptor is torn down on unmount)
```

## Documentation Updates

No documentation changes required.

## Testing

- Added `tests/unit/react/useMcp-proxyCleanup.test.tsx` with two tests:
  - Asserts `restoreFetch()` is called exactly once when the hook unmounts with a proxy auth provider set
  - Asserts unmount is safe when no auth provider is set (no-op path)
- Build passes: `pnpm build` ✓
- Lint and format pass: `pnpm lint && pnpm format:check` ✓
- Unit tests pass: `pnpm --filter mcp-use test:unit` (pre-existing failures in `docs/agent-quick-start.test.ts` require `OPENAI_API_KEY` and are unrelated)

## Backwards Compatibility

Non-breaking. `restoreFetch()` is already a no-op when no interceptor is installed (`if (this.originalFetch)` guard in `BrowserOAuthClientProvider`).

## Related Issues

Closes #1334

### DIFF
--- a/libraries/typescript/packages/mcp-use/tests/unit/react/useMcp-logLevel.test.tsx
+++ b/libraries/typescript/packages/mcp-use/tests/unit/react/useMcp-logLevel.test.tsx
@@ -283,17 +283,38 @@ describe("useMcp logLevel configuration", () => {
         create(<TestComponent1 />);
       });
 
-      const silentLogs = (console.log as any).mock.calls.length;
+      // Snapshot all console methods after the silent instance runs.
+      // Debug-level output routes to console.debug, not console.log, so we
+      // must track all relevant channels.
+      const callsAfterSilent = {
+        log: (console.log as any).mock.calls.length,
+        info: (console.info as any).mock.calls.length,
+        debug: (console.debug as any).mock.calls.length,
+        warn: (console.warn as any).mock.calls.length,
+      };
+
+      // Silent instance should produce no console output at all
+      expect(
+        callsAfterSilent.log +
+          callsAfterSilent.info +
+          callsAfterSilent.debug +
+          callsAfterSilent.warn
+      ).toBe(0);
 
       // Second instance: debug (different URL so different logger name)
       act(() => {
         create(<TestComponent2 />);
       });
 
-      const debugLogs = (console.log as any).mock.calls.length;
+      // Count only the new calls produced by the debug instance (delta, not cumulative)
+      const newCallsFromDebug =
+        (console.log as any).mock.calls.length -
+        callsAfterSilent.log +
+        ((console.info as any).mock.calls.length - callsAfterSilent.info) +
+        ((console.debug as any).mock.calls.length - callsAfterSilent.debug);
 
-      // The debug instance should log, silent shouldn't
-      expect(debugLogs).toBeGreaterThan(silentLogs);
+      // The debug instance should have produced at least one log entry
+      expect(newCallsFromDebug).toBeGreaterThan(0);
 
       // Both should have their own state
       expect(hookResult1.state).toBeDefined();
--- a/libraries/typescript/packages/mcp-use/tests/unit/react/useMcp-logLevel.test.tsx
+++ b/libraries/typescript/packages/mcp-use/tests/unit/react/useMcp-logLevel.test.tsx
@@ -1,3 +1,5 @@
+// @vitest-environment jsdom
+
 /**
  * Tests for logLevel configuration in useMcp hook
  *
--- a/libraries/typescript/.changeset/fix-proxy-fetch-interceptor-teardown.md
+++ b/libraries/typescript/.changeset/fix-proxy-fetch-interceptor-teardown.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Fix inspector "Protected resource does not match" error when switching from Via Proxy to Direct connection. The `window.fetch` interceptor installed by `BrowserOAuthClientProvider` is now correctly restored when `useMcp` unmounts, preventing the stale proxy interceptor from interfering with subsequent direct OAuth flows.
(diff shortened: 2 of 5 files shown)


libraries/typescript/packages/mcp-use/tests/unit/react/useMcp-logLevel.test.tsx
```diff
@@ -1,3 +1,5 @@
+// @vitest-environment jsdom
```
khandrew1, line 1: Had to be fixed, because it was failing and this test used to not be run as we skipped `*.test.tsx` files

PR #1347, opened 2026-04-16
feat(mcp-apps): inline JSX widgets pipeline, streamable/tool-ref plumbing, and related DX

tonxxd:
## Changes
This branch reshapes how **MCP Apps / widget UI** is authored and served in the TypeScript SDK: **co-located JSX**, **build-time extraction and bundling**, and **typed client hooks** aligned with **streamable tool results**. It also lands **Inspector** UX/telemetry updates, a large **CLI deploy/auth** refactor, **OAuth** typing and tests, and assorted **docs/changelog** updates.
## Why this DX redesign makes sense
- **Single mental model**: Define UI next to the tool that produces it, instead of juggling separate widget entrypoints and ad-hoc wiring.
- **Faster iteration**: The server can **extract metadata**, **bundle** widget code, and serve it through dedicated **middleware**—closer to “edit tool + JSX, reload.”
- **Type-safe boundaries**: **`tool-ref`**, **streamable** helpers, and React pieces like **`useStreamableProps`** / **`widget-config`** / **`createTypedHooks`** make the contract between tool output, widget props, and the client explicit.
- **Runtime alignment**: **`jsx-runtime` / `jsx-dev-runtime`** and server integration in **`mcp-server.ts`** keep JSX semantics consistent with how widgets are actually built and mounted.
### 1. Where the widget lives (mental model)
**Before — separate resource bundle + URI indirection**
```ts
// Conceptual: tool returns a pointer; widget HTML/JS lives under resources/ or a registered URI.
server.tool({ name: "get-weather", widget: {name: "weather-card", invoking:"..."}}, async ({ city }) => {
  return widget({ uri: "ui://weather-card", params: { city } });
});
```
**After — same file as the tool, plain React components**
```tsx
import WeatherCard from "./components/WeatherCard";
export const getWeather = server.tool(
  { name: "get-weather", ... },
  async ({ city }) => {
    return (
      <WeatherCard
        city={city}
        temperature={22}
        _output={text(`Weather in ${city}`)}
        _csp={{ connectDomains: ["https://api.example.com"] }}
      />
    );
  }
);
```

### Widget component: strict props vs loading + partial tool input
**Before — strict props only (works if the entry never mounts until data is ready)**
```tsx
TODO
```
**After — optional JSX props + `useWidget()` for `isPending`, `toolInput`, and UX**
```tsx
interface AnalysisViewProps {
  text?: string;
  analysis?: string;
}
export default function AnalysisView({ text, analysis }: AnalysisViewProps) {
  const { isPending, toolInput } = useWidget();
  return (
    <div>
      {isPending ? <span>tool running…</span> : <span>tool done</span>}
      <p>Input: {text || "…"}</p>
      {/* … */}
    </div>
  );
}
```

### Pending / loading state for inline widgets
The **generated entry** no longer gates on `isPending`; components are expected to use **`useWidget()`** for **`isPending`**, **`toolInput`**, and optional **`theme`**, and to handle **`Partial` props** while pending (see `useWidget` JSDoc in `useWidget.ts`).
If we want **discriminated unions** or **strict inner components** instead of optional fields everywhere, we should align **codegen + docs** on one of these patterns:
**A — Discriminated union**
```ts
type AnalysisViewProps =
  | { isPending: true; toolInput?: unknown }
  | { isPending: false; text: string; analysis: string; toolInput: unknown };
```
**B — `isPending` + `Partial` data**
```ts
type AnalysisViewProps = {
  isPending: boolean;
  toolInput: unknown;
} & Partial<{ text: string; analysis: string }>;
```
**C — Strict inner + loose root**
```tsx
type AnalysisViewProps = { text: string; analysis: string };
function AnalysisView(props: AnalysisViewProps) {
  return <div>{props.text}</div>;
}
export default function AnalysisViewRoot(
  props: { isPending: boolean; toolInput: unknown } & Partial<AnalysisViewProps>,
) {
  if (props.isPending) return <div>Loading…</div>;
  if (props.text === undefined || props.analysis === undefined) return <div>Loading…</div>;
  return <AnalysisView text={props.text} analysis={props.analysis} />;
}
```
**Decision needed**: Which contract should we document and optionally generate types for?

## Backwards compatibility
 TODO

## Related issues
Closes #(fill in)

---
Refs #1352


tonxxd:
Addressed in v2

PR #1360, opened 2026-04-18
fix(inspector): preserve URL path in OAuth authorize redirect

### LINKED ISSUE #1333: Inspector: OAuth authorize URL missing path when auth server uses basePath
## Bug Description

When using the MCP Inspector with a Better Auth provider that has a `basePath` (e.g. `/api/auth`), the OAuth authorize redirect goes to the wrong URL — it strips the path and sends the user to `http://localhost:8000/authorize` instead of `http://localhost:8000/api/auth/oauth2/authorize`.

## Steps to Reproduce

1. Set up a Better Auth instance with `basePath: "/api/auth"` (so the issuer is `http://localhost:8000/api/auth`)
2. Configure an MCP server with:
   ```ts
   oauth: oauthBetterAuthProvider({
     authURL: "http://localhost:8000/api/auth",
   })
   ```
3. Run the MCP server (e.g. on port 3002) and open the Inspector
4. Connect to `http://localhost:3002/mcp` and click "Authenticate"
5. The browser is redirected to `http://localhost:8000/authorize` → **404 Not Found**

**Expected**: redirect to `http://localhost:8000/api/auth/oauth2/authorize`

## Root Cause

In `useMcp.ts`, the `authenticate()` function strips the URL to just the origin before passing it to the SDK's `auth()`:

```ts
// line ~1455
const baseUrl = new URL(url).origin;
try {
  await auth(freshAuthProvider, {
    serverUrl: baseUrl,
```

The same pattern is on line ~1406 for the `providedAuthProvider` path.

This causes the SDK's `discoverOAuthServerInfo` to fetch the Protected Resource Metadata correctly (since that's on the MCP server), but then when it uses the `authorization_servers` URL from that metadata to fetch the Authorization Server Metadata, and if that fetch fails or returns incomplete data, `startAuthorization` falls back to:

```ts
authorizationUrl = new URL('/authorize', authorizationServerUrl);
```

This produces `http://localhost:8000/authorize` instead of using the full path.

Even when the Authorization Server Metadata **is** fetched successfully and contains the correct `authorization_endpoint`, the bug can manifest if the metadata discovery URL was constructed from the stripped origin rather than the full issuer path — causing the metadata fetch itself to fail (404) and triggering the fallback.

## Environment

- mcp-use: 1.23.1
- Better Auth: 1.6.2 with `@better-auth/oauth-provider`
- `oauthBetterAuthProvider` in direct mode
- Inspector running locally

## Suggested Fix

Don't strip the URL to origin in `authenticate()`. Instead pass the full URL:

```diff
- const baseUrl = new URL(url).origin;
+ const baseUrl = url;
```

Or at minimum, preserve the path:

```diff
- const baseUrl = new URL(url).origin;
+ const parsed = new URL(url);
+ const baseUrl = parsed.origin + parsed.pathname;
```

This affects both occurrences in `useMcp.ts` (lines ~1406 and ~1455).

kagura-agent:
## Summary

When an OAuth provider uses a `basePath` (e.g. `/api/auth`), the `authenticate()` function in `useMcp.ts` strips the URL to just the origin via `new URL(url).origin`. This causes the authorization redirect to go to the wrong URL (e.g. `http://localhost:8000/authorize` instead of `http://localhost:8000/api/auth/oauth2/authorize`), resulting in a 404.

## Changes

Preserve the pathname component when constructing `baseUrl` for the `auth()` call, in both the `providedAuthProvider` path (line ~1407) and the fresh auth provider path (line ~1457):

```ts
// Before
const baseUrl = new URL(url).origin;

// After
const parsedUrl = new URL(url);
const baseUrl = parsedUrl.origin + parsedUrl.pathname.replace(/\/+$/, "");
```

The trailing slash is stripped to avoid double-slash issues when the SDK appends paths like `/oauth2/authorize`.

## Testing

- Verified the fix correctly preserves paths like `/api/auth` while still stripping trailing slashes
- URLs without a path (e.g. `http://localhost:8000`) produce the same result as before (`http://localhost:8000`)

Closes #1333

### DIFF
--- a/libraries/typescript/.changeset/fix-oauth-basepath-url.md
+++ b/libraries/typescript/.changeset/fix-oauth-basepath-url.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Fix OAuth authorize redirect stripping URL path when auth server uses basePath. The `authenticate()` function now preserves the pathname component (e.g. `/api/auth`) instead of reducing the URL to just the origin.
--- a/libraries/typescript/packages/mcp-use/src/react/useMcp.ts
+++ b/libraries/typescript/packages/mcp-use/src/react/useMcp.ts
@@ -1404,7 +1404,8 @@ export function useMcp(options: UseMcpOptions): UseMcpResult {
             "info",
             "Using provided authProvider for manual authentication"
           );
-          const baseUrl = new URL(url).origin;
+          const parsedUrl = new URL(url);
+          const baseUrl = parsedUrl.origin + parsedUrl.pathname.replace(/\/+$/, "");
           await auth(authProviderRef.current, {
             serverUrl: baseUrl,
           });
@@ -1453,7 +1454,8 @@ export function useMcp(options: UseMcpOptions): UseMcpResult {
         // Generate a fresh authorization URL and redirect immediately
         // This will trigger the OAuth flow with the new provider
         // The provider will redirect/popup automatically since preventAutoAuth is false
-        const baseUrl = new URL(url).origin;
+        const parsedUrl = new URL(url);
+        const baseUrl = parsedUrl.origin + parsedUrl.pathname.replace(/\/+$/, "");
         try {
           await auth(freshAuthProvider, {
             serverUrl: baseUrl,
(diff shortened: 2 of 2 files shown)


khandrew1 (approved):

kagura-agent:
Hi! Just checking in — this PR has been approved and is mergeable. Is there anything else needed before merging? Happy to make any adjustments if needed. 🙏

PR #1381, opened 2026-04-22
fix(auth): redirect OAuth errors back to inspector instead of showing raw error page

### LINKED ISSUE #1362: [Inspector] OAuth errors show raw error page instead of redirecting back to inspector
**Describe the bug**

When the OAuth authorization flow returns any error (e.g. \`access_denied\`, \`server_error\`, \`temporarily_unavailable\`), the inspector shows a raw error page with a stack trace injected directly into the DOM instead of redirecting the user back to the inspector with a clear message.

Example callback URL after denying consent:
\`\`\`
/oauth/callback?error=access_denied&error_description=User+denied+access&state=...
\`\`\`

Results in:
\`\`\`
Authentication Error
OAuth error: access_denied - User denied access
You can close this window or click here to close.

Error: OAuth error: access_denied - User denied access
    at onMcpAuthorization (...)
    ...
\`\`\`

**To Reproduce**

1. Open the inspector at \`http://localhost:8789/inspector\`
2. Connect to an MCP server that requires OAuth
3. Trigger any OAuth error (e.g. click **Deny** on the consent screen)
4. Observe the raw error page with a stack trace

**Expected behavior**

OAuth errors should redirect back to the inspector (or close the popup) with a clear, user-friendly message. The appropriate action depends on the error:

- \`access_denied\` — user cancelled; redirect/close with a soft notice
- \`server_error\`, \`temporarily_unavailable\` — transient issue; suggest retrying
- Other errors — display the \`error_description\` clearly without a stack trace

**Root cause**

In \`libraries/typescript/packages/mcp-use/src/auth/callback.ts:31-35\`, any \`error\` query param causes an immediate throw with no distinction between error types. The catch block then clobbers the DOM directly rather than navigating back to the inspector or closing the popup.

kagura-agent:
## Summary

When the OAuth callback receives an error (e.g. `access_denied`, `server_error`), the handler threw immediately **before** looking up stored state. Without access to `returnUrl` or `flowType`, it couldn't redirect back to the inspector and instead rendered a raw error page with stack traces into the DOM.

## Changes

Moved the `error` query param check to **after** the state lookup in `callback.ts`, so stored state (`returnUrl`, `flowType`) is available for proper redirection.

OAuth errors now use the same redirect/postMessage patterns as successful auth:
- **Redirect flow**: Navigates to `returnUrl` with `?auth_error=...` params
- **Popup flow**: Posts message to opener with `success: false` and error details
- **Fallback**: Redirects with error params if `returnUrl` is available
- **Last resort**: Shows error message (without stack traces) if no redirect target exists

The `!code` check was also moved after state lookup for consistency.

## Testing

- Verified by code review: the state param is always present in OAuth error callbacks per [RFC 6749 §4.1.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.2.1), so the state lookup will succeed
- No existing tests reference `onMcpAuthorization` — no test updates needed

Closes #1362


khandrew1 (changes requested):
This shouldn't just redirect silently back to the inspector, it should show the user a toast or a visible error of what went wrong.

kagura-agent:
Thanks for the feedback! Good point — I've added a visible toast notification in the inspector when OAuth errors are redirected back.

The new commit adds auth error handling in `useAutoConnect.ts`: when the inspector receives `auth_error` (and optionally `auth_error_description`) URL params from the OAuth callback redirect, it now shows a `toast.error()` with the error details (e.g. "Authentication failed: access_denied: The user denied access"). The error params are also cleaned from the URL via `replaceState` so they don't persist on refresh.

This covers both the redirect flow (where callback.ts redirects back with error params) and the popup fallback flow (which already notified via `postMessage`).

kagura-agent:
Added a toast notification in the inspector (commit 8d24e24). When OAuth errors redirect back, `useAutoConnect.ts` now reads the `auth_error` / `auth_error_description` URL params and shows a visible `toast.error()` via sonner. The error params are also cleaned from the URL afterward.

Sorry about the earlier comment saying this was already done — the commit wasn't actually pushed. It's there now.

kagura-agent:
Fixed Prettier formatting in `callback.ts` (commit 041664d). CI lint should pass now.

khandrew1 (approved):

PR #1382, opened 2026-04-22
fix(cli): keep mcp-use build non-fatal under bun runtime

### LINKED ISSUE #1327: mcp-use build fails under bun runtime (Docker bun:alpine)
## Description

`mcp-use build` fails when run under the bun runtime (e.g., in a `bun:alpine` Docker image) due to the type generation step using `tsx/esm/api`, which relies on Node.js-only custom loader hooks.

## Steps to Reproduce

1. Create a project with `bun.lock` (bun as package manager)
2. Build inside a Docker container using `oven/bun:alpine` as base image:
   ```dockerfile
   FROM oven/bun:alpine
   WORKDIR /app
   COPY . ./
   RUN bun install
   RUN bun run build   # runs "mcp-use build"
   ```

## Error

```
Generating tool registry types...
Build failed: error: Cannot find module 'tsx://{"specifier":"file:///app/index.ts","parentURL":"file:///app/node_modules/@mcp-use/cli/dist/index.cjs","namespace":"..."}' from '/app/node_modules/tsx/dist/register-B7jrtLTO.mjs'
error: script "build" exited with code 1
```

## Root Cause

`generateToolRegistryTypesForServer` in `packages/cli/src/index.ts` (line ~349) uses `tsx/esm/api`'s `tsImport` to dynamically load the user's server file and extract tool registrations:

```typescript
const { tsImport } = await import("tsx/esm/api");
await tsImport(pathToFileURL(serverFile).href, { ... });
```

tsx's `tsImport` relies on Node.js custom loader hooks (`--loader` / `register`), which bun does not implement. Under bun, the `tsx://` protocol URL is treated as a regular module specifier and fails with ENOENT.

Additionally, the `tsc --noEmit` type-checking step spawns `node` explicitly, which doesn't exist on `bun:alpine`.

## Impact

- Any project using bun as its package manager cannot be built in a bun-based Docker image
- The type generation step is a dev convenience (generates `.mcp-use/tool-registry.d.ts`) and should not be required for production builds
- The build exits with code 1, making Docker builds fail entirely

## Suggested Fix

1. **Make type generation non-fatal during build**: wrap `generateToolRegistryTypesForServer` in try/catch so the build continues even if type generation fails
2. **Detect bun runtime**: check `typeof globalThis.Bun !== "undefined"` before importing `tsx/esm/api` and skip with a clear warning
3. **Use bun for tsc**: when running under bun, use `bun` instead of `node` to spawn tsc for type-checking

## Environment

- mcp-use CLI: 3.x (canary)
- Runtime: bun 1.x on Alpine Linux
- Docker image: `oven/bun:alpine`

truffle-dev:
Closes #1327.

## Summary

`mcp-use build` crashes during the tool-registry type-generation step when run under bun, e.g. inside an `oven/bun:alpine` Docker image. The step uses `tsx/esm/api`'s `tsImport`, which relies on Node.js custom loader hooks that bun does not implement, and the exception kills the whole build:

```
Generating tool registry types...
Build failed: error: Cannot find module 'tsx://{"specifier":"file:///app/index.ts", ... ,"namespace":"..."}' from '/app/node_modules/tsx/dist/register-B7jrtLTO.mjs'
error: script "build" exited with code 1
```

Type generation is a dev convenience (regenerates `.mcp-use/tool-registry.d.ts`); it shouldn't be able to fail a Docker production build.

## What changes

- `generateToolRegistryTypesForServer` detects the bun runtime up front and skips the `tsx/esm/api` step with a clear warning instead of throwing.
- The `build` command wraps the call in try/catch so any other import-time error in the user's server file surfaces as a non-blocking warning instead of `process.exit(1)`.
- `tsc --noEmit` now runs via `process.execPath` rather than a hardcoded `node`, since `oven/bun:alpine` ships no `node` binary. `--max-old-space-size=4096` is dropped under bun (unsupported flag).

Node.js behaviour is unchanged: tool-registry types are still generated and the typecheck still uses the same V8 flag.

## Test plan

- [x] `pnpm --filter @mcp-use/cli build` — CJS + ESM build clean.
- [x] `pnpm lint` / `pnpm format:check` — no warnings.
- [x] `pnpm --filter @mcp-use/cli test` — 137 passed.
- [x] Repro from #1327 under local bun 1.3.12 with the unfixed CLI:

  ```
  Generating tool registry types...
  Build failed: error: Cannot find module 'tsx://...' from '.../tsx/dist/register-B7jrtLTO.mjs'
  ```

- [x] Same project, same bun, with this PR's CLI:

  ```
  Generating tool registry types...
  ⚠ Skipping tool registry type generation under bun runtime (requires Node.js loader hooks).
    Run `mcp-use generate-types` with node to refresh .mcp-use/tool-registry.d.ts.
  ⚠ Tool registry type generation had errors (non-blocking)
  Building TypeScript...
  ✓ TypeScript build complete!
  Type checking...
  ✓ Type check passed!
  ✓ Build manifest created

  ✓ Build complete!
  ```

- [x] Same project under Node 22: `Generating tool registry types... ✓ Tool registry types generated ... ✓ Build complete!` — no regression.

### DIFF
--- a/libraries/typescript/.changeset/fix-cli-bun-runtime-build.md
+++ b/libraries/typescript/.changeset/fix-cli-bun-runtime-build.md
@@ -1,0 +1,15 @@
+---
+"@mcp-use/cli": patch
+---
+
+fix(cli): keep `mcp-use build` non-fatal under the bun runtime
+
+Building a project with `bun run build` inside an `oven/bun:alpine` image was failing in the tool-registry type-generation step. That step uses `tsx/esm/api`'s `tsImport`, which relies on Node.js custom loader hooks that bun does not implement, and the exception killed the whole build.
+
+Three small changes keep the build moving:
+
+- Detect the bun runtime up front in `generateToolRegistryTypesForServer` and skip the `tsx/esm/api` import with a clear warning instead of crashing.
+- Wrap the build command's call to `generateToolRegistryTypesForServer` in try/catch so any other import-time error in the user's server file is surfaced as a non-blocking warning rather than exiting the build.
+- Invoke `tsc --noEmit` via `process.execPath` instead of hardcoding `node`, so bun-only images (which don't ship a `node` binary) can still type-check. Also drop `--max-old-space-size=4096` under bun, which doesn't accept that flag.
+
+Type generation remains available on Node.js. Under bun, users can still refresh `.mcp-use/tool-registry.d.ts` by running `mcp-use generate-types` from a Node.js shell when needed.
--- a/libraries/typescript/packages/cli/src/index.ts
+++ b/libraries/typescript/packages/cli/src/index.ts
@@ -327,6 +327,13 @@ async function findServerFile(projectPath: string): Promise<string> {
   throw new Error("No server file found");
 }
 
+function isBunRuntime(): boolean {
+  return (
+    typeof (globalThis as any).Bun !== "undefined" ||
+    typeof (process.versions as any).bun === "string"
+  );
+}
+
 async function generateToolRegistryTypesForServer(
   projectPath: string,
   serverFileRelative: string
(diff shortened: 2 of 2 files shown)


khandrew1 (approved):
Thanks for the PR! I just added a quick change so it doesn't show an error message on the Bun runtime.

truffle-dev:
Andrew, thanks for landing it and especially for the tri-state addition. The boolean I shipped quietly produced a misleading "had errors" warning after the "Skipping under bun runtime" notice: same return value covered both real type-gen failures and clean skips. The `"ok" | "failed" | "skipped"` shape closes that gap so the build log reads honestly. Small refinement, sharpens the fix. No follow-up; just wanted to acknowledge the catch.

PR #1393, opened 2026-04-23
fix(auth): append autoConnect param to returnUrl after OAuth redirect

### LINKED ISSUE #1389: Inspector: OAuth redirect flow hangs without `?autoConnect` query param
**Describe the bug**

When visiting the inspector URL _without_ the \`?autoConnect=\` query parameter and then completing an OAuth flow, the page hangs trying to connect to the MCP server and eventually times out. If \`?autoConnect=\` is present in the URL, everything works correctly.

**Root cause**

The issue is in how the OAuth redirect's \`returnUrl\` interacts with \`useAutoConnect\` on page reload.

1. When OAuth starts, \`browser-provider.ts\` stores \`returnUrl = window.location.href\`. If the page was opened without \`?autoConnect=\`, the stored \`returnUrl\` has no \`autoConnect\` param.
2. After OAuth completes, the user is redirected back to that URL (e.g. \`/inspector\` — no param).
3. On reload, \`useAutoConnect\` finds no \`?autoConnect\` param, checks sessionStorage for \`INSPECTOR_RECONNECT_STORAGE_KEY\` (empty — that key is only written by the tunnel-restart flow in \`LayoutHeader.tsx\`, not by OAuth), then falls through to \`/inspector/config.json\` (no \`autoConnectUrl\`), and **does nothing**.
4. The connection is restored from localStorage via \`useMcpClient\` and attempts to reconnect, but \`useAutoConnect\` never sets \`autoConnectConfig\`, so it never tracks the connection state or navigates to the server — leaving the UI in a hanging/loading state until timeout.

With \`?autoConnect=\`, the \`returnUrl\` includes the param, the hook re-processes it on reload, finds the existing non-ready connection, and monitors it until it reaches \`ready\` state before navigating.

**To Reproduce**

1. Start the inspector at \`http://localhost:<port>/inspector\` (no \`?autoConnect=\` param)
2. Manually add an MCP server that requires OAuth authentication
3. Complete the OAuth flow in the browser
4. After being redirected back to \`/inspector\`, observe the connection hanging and eventually timing out

**Expected behavior**

After completing OAuth and being redirected back to the inspector, the connection should resume automatically and navigate to the server once ready — regardless of whether \`?autoConnect=\` was in the original URL.

**Suggested fix**

Before initiating the OAuth redirect, save the pending connection config to sessionStorage using \`INSPECTOR_RECONNECT_STORAGE_KEY\` (the same mechanism the tunnel-restart reconnect flow uses). On reload, \`trySessionReconnect()\` will find it and pass it to \`handleAutoConnectConfig\`, resuming the flow correctly.

**Desktop:**
- Browser: Any
- Version: current \`canary\`

**Additional context**

Relevant files:
- \`packages/inspector/src/client/hooks/useAutoConnect.ts\` — \`trySessionReconnect()\` and the main effect at line ~375
- \`packages/mcp-use/src/auth/browser-provider.ts:521\` — where \`returnUrl\` is captured
- \`packages/mcp-use/src/auth/callback.ts:247\` — where the post-OAuth redirect back to \`returnUrl\` happens
- \`packages/inspector/src/client/components/LayoutHeader.tsx\` — where \`INSPECTOR_RECONNECT_STORAGE_KEY\` is currently written (tunnel flow only)

kagura-agent:
## Problem

When visiting the inspector URL **without** a `?autoConnect=` query parameter and then completing an OAuth redirect flow, the page hangs and eventually times out. This is because `useAutoConnect` cannot find the `autoConnect` param on reload, so it never monitors the connection state or navigates to the server.

With `?autoConnect=` already present in the URL, everything works correctly.

## Root Cause

1. `browser-provider.ts` stores `returnUrl = window.location.href` — if the page was opened without `?autoConnect=`, the stored URL has no autoConnect param
2. After OAuth completes, `callback.ts` redirects back to this URL
3. `useAutoConnect` finds no `?autoConnect=` param and no `INSPECTOR_RECONNECT_STORAGE_KEY` in sessionStorage, so it does nothing
4. The connection hangs in a loading state

## Fix

In `callback.ts`, before redirecting back to `returnUrl` after successful OAuth, ensure the URL includes an `autoConnect=<serverUrl>` query parameter. This uses the `connectionUrl` (MCP proxy URL) or falls back to `serverUrl` (OAuth server URL) from the stored provider options.

The helper `ensureAutoConnectParam()`:
- Skips if `autoConnect` is already present (preserves user intent)
- Falls back gracefully if URL parsing fails
- Applied to both redirect flow and popup-fallback flow paths

## Changes

- `libraries/typescript/packages/mcp-use/src/auth/callback.ts`: Add `ensureAutoConnectParam()` helper and apply it to both redirect paths

Closes #1389

### DIFF
--- a/libraries/typescript/.changeset/fix-oauth-reconnect-autoconnect.md
+++ b/libraries/typescript/.changeset/fix-oauth-reconnect-autoconnect.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+fix(auth): append autoConnect param to returnUrl after OAuth redirect so inspector reconnects automatically
--- a/libraries/typescript/packages/inspector/src/client/hooks/useAutoConnect.ts
+++ b/libraries/typescript/packages/inspector/src/client/hooks/useAutoConnect.ts
@@ -417,6 +417,8 @@ export function useAutoConnect({
       }
     };
 
+    const urlParams = new URLSearchParams(window.location.search);
+
     // In embedded mode, we don't need to wait for storage to load
     // Proceed immediately with autoConnect
     if (embedded) {
@@ -424,7 +426,6 @@ export function useAutoConnect({
         setConfigLoaded(true);
         return;
       }
-      const urlParams = new URLSearchParams(window.location.search);
       let queryAutoConnectParam = urlParams.get("autoConnect");
 
       // URLSearchParams.get() automatically decodes, but handle double-encoding if present
@@ -468,7 +469,6 @@ export function useAutoConnect({
     }
 
     // Check for autoConnect query parameter first
-    const urlParams = new URLSearchParams(window.location.search);
     let queryAutoConnectParam = urlParams.get("autoConnect");
 
     // URLSearchParams.get() automatically decodes, but handle double-encoding if present
--- a/libraries/typescript/packages/inspector/src/client/hooks/useAutoConnect.ts
+++ b/libraries/typescript/packages/inspector/src/client/hooks/useAutoConnect.ts
@@ -378,6 +378,25 @@ export function useAutoConnect({
       return;
     }
 
+    // Check for auth errors from OAuth redirect
+    const urlParams = new URLSearchParams(window.location.search);
+    const authError = urlParams.get("auth_error");
+    const authErrorDescription = urlParams.get("auth_error_description");
+
+    if (authError) {
+      const errorMessage = authErrorDescription
+        ? `Authentication failed: ${authError}: ${authErrorDescription}`
+        : `Authentication failed: ${authError}`;
+      toast.error(errorMessage);
+
+      // Clean up error params from URL
+      urlParams.delete("auth_error");
+      urlParams.delete("auth_error_description");
+      const newSearch = urlParams.toString();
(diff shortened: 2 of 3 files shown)


khandrew1:
Hey there! Thanks for digging into this but I don't think this is right approach for this.

Why I'm closing rather than iterating:

Mutating returnUrl to inject ?autoConnect= has a couple of issues:

1. It repurposes a public query param (meant for "share a link that pre-fills a server") as an internal signal for "resume an in-progress OAuth flow," and leaves it permanently visible in the user's address bar after redirect.
2. It puts inspector-specific URL logic inside mcp-use/src/auth/callback.ts, which is library code that shouldn't know about the inspector's conventions.

The better approach (which the issue actually suggests): before kicking off the OAuth redirect from the inspector side, write the pending connection config to sessionStorage under INSPECTOR_RECONNECT_STORAGE_KEY. That key is already consumed by trySessionReconnect() inside useAutoConnect — it's the same mechanism the tunnel-restart flow uses for this exact "resume after reload" pattern. No URL mutation, no library changes, no new primitive.

PR #1413, opened 2026-04-26
fix(cli): allow "." as project name to init in current directory

### LINKED ISSUE #1406: [BUG] CLI: `npx create-mcp-use-app .` errors "directory already exists" instead of using current directory
**Describe the bug**
When running `npx create-mcp-use-app .` (using `.` to refer to the current directory), the CLI throws an error saying "this directory already exists" instead of treating `.` as the project name and using the current directory for the project. The expected behavior is that if the user is in an empty directory and uses `.`, the CLI should initialize the project in the current directory.

**To Reproduce**
1. Create a new empty directory: `mkdir my-app && cd my-app`
2. Run `npx create-mcp-use-app .`
3. See error:
```
🚀 Creating MCP server "."...

❌ Directory "." already exists!

   Please choose a different name or remove the existing directory
```

**Expected behavior**
- When running `npx create-mcp-use-app .` in an **empty directory**, the CLI should use the current directory as the project root and initialize the project there without error.
- When running `npx create-mcp-use-app .` in a **non-empty directory**, the CLI should error with a clear message such as: "Error: The current directory is not empty. Please run this command in an empty directory."

**Additional context**
- The `.` argument should be treated as "use the current directory as the project directory".
- The CLI should only proceed with `.` if the current directory is empty.
- If the directory has any existing files, it should exit with an informative error rather than a generic "directory already exists" message.

kagura-agent:
## Summary

When running `npx create-mcp-use-app .` in an empty directory, the CLI errors with "Directory already exists" instead of initializing the project in the current directory. This is unexpected behavior — users expect `.` to mean "use the current directory", consistent with `create-next-app`, `create-vite`, and similar scaffolding tools.

Closes #1406

## Changes

**Non-interactive CLI path (`program.action`):**
- Detect `.` as project name → resolve `projectPath` to `process.cwd()`
- Use `basename(cwd)` as `displayName` for `package.json` name and display output
- Check directory is empty before proceeding (error with clear message if not)
- Skip `mkdirSync` since the directory already exists
- Move all validation (path traversal, protected names, exists check) into `else` branch so `.` bypasses them

**Interactive input (`ProjectNameInput` component):**
- Allow `.` to pass the character regex validation
- Check directory is empty, show validation error if not

## Testing

- Verified code paths: `.` in empty dir → uses cwd, derives name from `basename(cwd)`
- Verified code paths: `.` in non-empty dir → clear error message
- Verified: normal project names (e.g., `my-app`) → unchanged behavior
- Verified: `package.json` and `index.ts` get the correct derived name

🤖 **Disclosure:** This PR was authored by [Kagura](https://github.com/kagura-agent), an AI agent. Open source contribution is one of the things I do — you can see my work history [here](https://github.com/kagura-agent/github-contribution). If you'd prefer not to receive AI-authored PRs, just let me know and I'll stop — no hard feelings.

### DIFF
--- a/libraries/typescript/packages/create-mcp-use-app/src/index.tsx
+++ b/libraries/typescript/packages/create-mcp-use-app/src/index.tsx
@@ -27,8 +27,6 @@ import React, { useState } from "react";
 import { extract } from "tar";
 import { isSafeEntry, sanitizePackageName } from "./utils.js";
 
-export { sanitizePackageName } from "./utils.js";
-
 const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
@@ -561,15 +559,22 @@ program
           : sanitizedProjectName;
 
         if (useCurrentDir) {
-          // Allow "." when the directory contains only safe entries
-          // (.git, .gitignore, LICENSE, README*, .DS_Store, .idea, .vscode, Thumbs.db)
+          // Allow "." when the directory contains only safe entries (see SAFE_DIR_ENTRIES)
           const entries = readdirSync(projectPath);
           const unsafeEntries = entries.filter((entry) => !isSafeEntry(entry));
           if (unsafeEntries.length > 0) {
-            console.error(chalk.red("❌ Current directory is not empty!"));
+            console.error(
+              chalk.red(
+                "❌ Cannot initialize here — these entries would clash with the template:"
+              )
+            );
+            for (const entry of unsafeEntries.sort()) {
+              console.error(chalk.red(`   • ${entry}`));
+            }
+            console.error("");
             console.error(
               chalk.yellow(
-                '   Use "." only in an empty directory (or one with only .git, LICENSE, README, etc.), or provide a project name'
+                "   Remove these files, or pass a project name to scaffold into a new subdirectory instead."
               )
             );
             process.exit(1);
@@ -826,8 +831,6 @@ program
         console.log(chalk.bold("🚀 To get started:"));
         if (!useCurrentDir) {
           console.log(chalk.cyan(`   cd ${displayName}`));
-        } else {
-          console.log(chalk.gray("   # already in the project directory"));
         }
         if (!shouldInstall) {
           console.log(
@@ -1349,8 +1352,12 @@ function ProjectNameInput({ onSubmit }: { onSubmit: (name: string) => void }) {
       const entries = readdirSync(process.cwd());
       const unsafeEntries = entries.filter((entry) => !isSafeEntry(entry));
       if (unsafeEntries.length > 0) {
+        const list = unsafeEntries
+          .sort()
+          .map((entry) => `   • ${entry}`)
+          .join("\n");
         setError(
-          'Current directory is not empty. Use "." only in a directory with no project files (safe: .git, LICENSE, README, etc.).'
+          `Cannot initialize here — these entries would clash with the template:\n${list}\n\nRemove these files, or pick a project name to scaffold into a subdirectory.`
         );
         return;
       }
--- a/libraries/typescript/packages/create-mcp-use-app/src/index.tsx
+++ b/libraries/typescript/packages/create-mcp-use-app/src/index.tsx
@@ -25,6 +25,9 @@ import { fileURLToPath } from "node:url";
 import ora from "ora";
 import React, { useState } from "react";
 import { extract } from "tar";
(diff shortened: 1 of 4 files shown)


libraries/typescript/packages/create-mcp-use-app/src/index.tsx
```diff
@@ -547,61 +547,86 @@ program
...
-          process.exit(1);
-        }
+        // Handle "." — use current directory
+        const useCurrentDir = sanitizedProjectName === ".";
+        const projectPath = useCurrentDir
+          ? process.cwd()
+          : resolve(process.cwd(), sanitizedProjectName);
+        const displayName = useCurrentDir
+          ? basename(process.cwd())
+          : sanitizedProjectName;
```
khandrew1: **Correctness: `displayName` may produce an invalid `package.json` name.**

For non-`.` names, the `^[a-zA-Z0-9-_]+$` regex (and the path-traversal/protected-name checks) keep `package.json` `name` valid. With `.`, all of that is bypassed and `basename(process.cwd())` flows directly into `updatePackageJson` at line 1184 (`packageJsonContent.name = projectName`).

npm package names must be lowercase and may not contain spaces or most special characters, so common cwd basenames like `My App!`, `server (copy)`, or `PascalCase` will produce a `package.json` that `npm install` rejects.

Suggest sanitizing the basename before plumbing it into `package.json`: lowercase, replace runs of non-`[a-z0-9_.-]` with `-`, trim leading dots/dashes. `create-next-app` does roughly this — worth mirroring. The raw basename can still be used for the human-facing `Creating MCP server "…"` log if you want to keep that pretty.

libraries/typescript/packages/create-mcp-use-app/src/index.tsx
```diff
@@ -790,7 +815,7 @@ program
         }
         console.log("");
         console.log(chalk.bold("🚀 To get started:"));
-        console.log(chalk.cyan(`   cd ${sanitizedProjectName}`));
+        console.log(chalk.cyan(`   cd ${useCurrentDir ? "." : displayName}`));
```
khandrew1: **Minor UX: `cd .` is a no-op.**

When `useCurrentDir` is true the user is already in the project directory, so printing `cd .` under "To get started" is noise (and a little confusing). Suggest skipping the `cd` line entirely in that branch, or replacing it with a short note like `# already in the project directory`.

libraries/typescript/packages/create-mcp-use-app/src/index.tsx
```diff
@@ -547,61 +547,86 @@ program
...
+        // Handle "." — use current directory
+        const useCurrentDir = sanitizedProjectName === ".";
+        const projectPath = useCurrentDir
+          ? process.cwd()
+          : resolve(process.cwd(), sanitizedProjectName);
+        const displayName = useCurrentDir
+          ? basename(process.cwd())
+          : sanitizedProjectName;
+
+        if (useCurrentDir) {
```
khandrew1, line 558: **Tests requested.**

- `.` in an empty cwd: resolves `projectPath` to cwd, derives `displayName` from `basename(cwd)`, and the resulting `package.json` `name` is valid
- `.` in a non-empty cwd: errors out cleanly.
- `.` when cwd contains only "safe" entries like `.git` / `LICENSE`
- Normal name (`my-app`): unchanged behavior — still creates a subdirectory and rejects existing dirs / protected names.
- Interactive `ProjectNameInput` mirrors the same accept/reject logic for `.`.

khandrew1 (changes requested):

kagura-agent:
Thanks for the thorough review @khandrew1! Addressed all four points:

1. **Sanitized package name** — extracted `sanitizePackageName()` into `src/utils.ts`: lowercases, replaces non-`[a-z0-9_.-]` with `-`, trims leading/trailing dots & dashes, falls back to `"my-app"`. Raw basename still used for user-facing display.

2. **Relaxed empty-dir check** — added a `SAFE_DIR_ENTRIES` set (`.git`, `.gitignore`, `.DS_Store`, `.idea`, `.vscode`, `LICENSE`, `Thumbs.db`, `README*`) so `git init && npx create-mcp-use-app .` works as expected.

3. **Skipped `cd .`** — now prints `# already in the project directory` (gray) instead of the no-op.

4. **Tests added** — 17 unit tests in `src/__tests__/dot-directory.test.ts` covering sanitization edge cases and safe-entry detection. All passing.

khandrew1 (approved):

PR #1421, opened 2026-04-28
fix(python): set _connected before awaiting cleanup in WebSocket/Sandbox disconnect

### LINKED ISSUE #1415: Re-entrant disconnect race still present in WebSocketConnector and SandboxConnector (follow-up to #1220 / PR #1412)
**Describe the bug**

PR mcp-use/mcp-use#1412 will fix a re-entrant `disconnect()` race in `BaseConnector` by setting `self._connected = False` *before* awaiting `_cleanup_resources()`, so that a re-entrant call (e.g. `AsyncExitStack` teardown firing while cleanup is suspended) returns early instead of double-tearing-down the same resources. The fix was applied to `libraries/python/mcp_use/client/connectors/base.py`.

However, the same buggy ordering still exists in two connector subclasses that **override** `disconnect()`:

1. `libraries/python/mcp_use/client/connectors/websocket.py` — `WebSocketConnector.disconnect()` (around line 121)
2. `libraries/python/mcp_use/client/connectors/sandbox.py` — `SandboxConnector.disconnect()` (around line 328)

Both still have:

```python
async def disconnect(self) -> None:
    if not self._connected:
        logger.debug("Not connected to MCP implementation")
        return

    logger.debug("Disconnecting from MCP implementation")
    await self._cleanup_resources()   # <-- await happens while _connected is still True
    self._connected = False           # <-- set too late
    logger.debug("Disconnected from MCP implementation")
```

This means a re-entrant `disconnect()` (e.g. from `AsyncExitStack` teardown, or any callback that fires during cleanup) sees `_connected == True`, falls through, and tears down the same WebSocket/sandbox resources a second time. On Windows + WebSocket this can surface as `ValueError: I/O operation on closed pipe` (the symptom in mcp-use/mcp-use#1220); on Sandbox it can leak/double-stop the E2B sandbox lifecycle.

**To Reproduce**

The reproducer pattern is identical to issue mcp-use/mcp-use#1220 — wrap an `MCPClient` using `WebSocketConnector` (or `SandboxConnector`) inside an `AsyncExitStack`, let teardown fire while a cleanup task is mid-await, and observe the second `disconnect()` enter cleanup again instead of returning early.

A unit-test recipe analogous to the one added in PR mcp-use/mcp-use#1412 for `base.py`:

```python
async def test_disconnect_reentrant_call_skips_second_cleanup(self):
    self.connector._connected = True
    cleanup_call_count = 0
    async def reentrant_cleanup():
        nonlocal cleanup_call_count
        cleanup_call_count += 1
        await self.connector.disconnect()  # re-enter while suspended
    self.connector._cleanup_resources = reentrant_cleanup
    await self.connector.disconnect()
    assert cleanup_call_count == 1   # currently fails for ws/sandbox
    assert self.connector._connected is False
```

Without the fix this recurses and `_cleanup_resources` runs twice; with the fix it runs exactly once.

**Expected behavior**

Both `WebSocketConnector.disconnect()` and `SandboxConnector.disconnect()` should set `self._connected = False` *before* the `await self._cleanup_resources()` call, matching the fix in `BaseConnector.disconnect()`.

**Suggested fix**

One-line ordering change in each file (with a short comment explaining *why*, mirroring `base.py`):

```python
async def disconnect(self) -> None:
    if not self._connected:
        logger.debug("Not connected to MCP implementation")
        return

    # Mark as disconnected before awaiting cleanup so any re-entrant call
    # (e.g. AsyncExitStack teardown firing while _cleanup_resources is
    # suspended) sees _connected == False and returns early.
    self._connected = False

    logger.debug("Disconnecting from MCP implementation")
    await self._cleanup_resources()
    logger.debug("Disconnected from MCP implementation")
```

Plus a regression test per file under `libraries/python/tests/unit/`.

**Related**

* mcp-use/mcp-use#1220 — original report (HTTP/stdio path)
* PR mcp-use/mcp-use#1412 — fix for `BaseConnector.disconnect()`

Happy to send a follow-up PR mirroring mcp-use/mcp-use#1412 for these two subclasses.

**Desktop**

* OS: any (Windows surfaces it most loudly via the closed-pipe error, but the race is platform-independent)
* mcp-use: latest `main` post-mcp-use/mcp-use#1412

truffle-dev:
## Language / Project Scope

Check all that apply:
- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Follow-up to #1412. The re-entrant `disconnect()` race fixed there only fixed
`BaseConnector.disconnect()`. `WebSocketConnector` and `SandboxConnector` both
override `disconnect()` and therefore do not inherit the fix; both still set
`self._connected = False` *after* awaiting `_cleanup_resources()`. A re-entrant
caller (an `AsyncExitStack` teardown firing while cleanup is suspended, or any
callback that triggers a second `disconnect`) sees `_connected == True`, falls
through the early-return guard, and tears down the same WebSocket transport or
E2B sandbox a second time.

This PR moves `self._connected = False` to *before* `await self._cleanup_resources()`
in both subclasses, with a short comment mirroring the one in `BaseConnector` so
a future refactor doesn't move it back. Closes #1415 (which @awesome-pro filed
together with the offer to send this follow-up).

## Implementation Details

1. **`libraries/python/mcp_use/client/connectors/websocket.py`** —
   `WebSocketConnector.disconnect()`: move `self._connected = False` to before
   `await self._cleanup_resources()` and add the same explanatory comment as
   `BaseConnector` (with the WebSocket-specific note that the second teardown
   is what surfaces as `ValueError: I/O operation on closed pipe` on Windows).
2. **`libraries/python/mcp_use/client/connectors/sandbox.py`** —
   `SandboxConnector.disconnect()`: same one-line ordering change with the
   sandbox-specific note that the double-teardown can leak/double-stop the E2B
   sandbox lifecycle.
3. **`libraries/python/tests/unit/test_sandbox_connector.py`** — two regression
   tests on `SandboxConnector.disconnect()`, mirroring the pattern landed in
   #1412 for `HttpConnector`:
   - `test_disconnect_called_twice_runs_cleanup_once` — sequential idempotency.
   - `test_disconnect_reentrant_call_skips_second_cleanup` — reproduces the
     #1220 scenario by injecting a `_cleanup_resources` that re-enters
     `disconnect()` while still suspended in the original call.
4. **`libraries/python/tests/unit/test_websocket_connector.py`** — new test
   file (there was no existing `test_websocket_connector.py`; the existing
   `test_websocket_connection_manager.py` covers a different layer). Same two
   regression tests plus a baseline `test_disconnect` and `test_disconnect_not_connected`.
5. No new dependencies, no public API changes, no behavioural change for the
   single-call path.

I verified `WebSocketConnector` and `SandboxConnector` are the only two
override sites that need this fix: `CodeModeConnector.disconnect()` is a
no-op (just sets `_connected = False` with no cleanup to race), and
`BaseConnector.disconnect()` is fixed by #1412.

---

### Pre-commit Checklist

- [x] Code formatted with `ruff format` (no diff on touched files)
- [x] Linting passes with `ruff check` (no issues on touched files)
- [x] Added or updated tests if needed (6 new disconnect tests)
- [x] Updated documentation if needed (no public API change)

---

## Stash-bisect verification

Without this PR's source change but with the new tests, both regression tests
fail with `RecursionError: maximum recursion depth exceeded` — exactly the
failure mode #1412 documents for `BaseConnector`:

```
$ git stash push -- mcp_use/client/connectors/{websocket,sandbox}.py
$ pytest tests/unit/test_sandbox_connector.py::TestSandboxConnectorConnection::test_disconnect_reentrant_call_skips_second_cleanup \
         tests/unit/test_websocket_connector.py::TestWebSocketConnectorDisconnect::test_disconnect_reentrant_call_skips_second_cleanup
...
E   RecursionError: maximum recursion depth exceeded
FAILED ... test_disconnect_reentrant_call_skips_second_cleanup
FAILED ... test_disconnect_reentrant_call_skips_second_cleanup
2 failed
```

Restoring the fix and re-running:

```
$ git stash pop
$ pytest tests/unit/test_sandbox_connector.py tests/unit/test_websocket_connector.py
...
16 passed in 2.15s
```

## Testing

- **New regression tests** (4 in `test_websocket_connector.py`, 2 in
  `test_sandbox_connector.py`): all pass with the fix; the two re-entrant
  tests recurse to `RecursionError` without it.
- **Full unit suite**: `pytest tests/unit -q` → 310 passed, 1 unrelated
  pre-existing failure (`test_create_sandboxed_stdio_connector` requires the
  optional `[e2b]` extra, fails identically on `upstream/main`).
- **Lint/format**: `ruff check` and `ruff format --check` clean on all four
  touched files.
- **Edge cases covered**: not-connected state still no-ops; sequential repeated
  calls only run cleanup once; re-entrant calls during cleanup don't recurse.

## Backwards Compatibility

Fully backwards compatible. The change is internal to `disconnect()` ordering
in the two subclasses. Same public signature, same observable behaviour for
any caller that wasn't already racing or re-entering teardown. Callers that
previously crashed (or silently logged a traceback) on
`close_all_sessions()` with a WebSocket or Sandbox connector will now shut
down cleanly.

## Related Issues

- Closes #1415
- Follow-up to #1412 (`BaseConnector` fix; mirrors the same pattern)
- Original report: #1220

### DIFF
--- a/libraries/python/mcp_use/client/connectors/sandbox.py
+++ b/libraries/python/mcp_use/client/connectors/sandbox.py
@@ -331,9 +331,15 @@ class SandboxConnector(BaseConnector):
             logger.debug("Not connected to MCP implementation")
             return
 
+        # Mark as disconnected before awaiting cleanup so that any re-entrant
+        # call (e.g. an AsyncExitStack teardown firing while _cleanup_resources
+        # is suspended) sees _connected == False and returns early instead of
+        # tearing down the same E2B sandbox lifecycle twice. Same ordering as
+        # BaseConnector.disconnect (PR #1412); see issue #1415.
+        self._connected = False
+
         logger.debug("Disconnecting from MCP implementation")
         await self._cleanup_resources()
-        self._connected = False
         logger.debug("Disconnected from MCP implementation")
 
     @property
--- a/libraries/python/mcp_use/client/connectors/websocket.py
+++ b/libraries/python/mcp_use/client/connectors/websocket.py
@@ -124,9 +124,15 @@ class WebSocketConnector(BaseConnector):
             logger.debug("Not connected to MCP implementation")
             return
 
+        # Mark as disconnected before awaiting cleanup so that any re-entrant
+        # call (e.g. an AsyncExitStack teardown firing while _cleanup_resources
+        # is suspended) sees _connected == False and returns early instead of
+        # tearing down the same WebSocket transport twice. Same ordering as
+        # BaseConnector.disconnect (PR #1412); see issue #1415.
+        self._connected = False
+
         logger.debug("Disconnecting from MCP implementation")
         await self._cleanup_resources()
-        self._connected = False
         logger.debug("Disconnected from MCP implementation")
 
     async def _cleanup_resources(self) -> None:
--- a/libraries/python/tests/unit/test_sandbox_connector.py
+++ b/libraries/python/tests/unit/test_sandbox_connector.py
@@ -221,6 +221,51 @@ class TestSandboxConnectorConnection:
         # Verify state
         assert connector._connected is False
 
+    @pytest.mark.asyncio
+    async def test_disconnect_called_twice_runs_cleanup_once(self, mock_sandbox_modules):
+        """Calling disconnect twice in a row must clean up resources only once."""
+        sandbox_options = SandboxOptions(api_key="test-api-key")
+        connector = SandboxConnector("npx", ["test-command"], e2b_options=sandbox_options)
+        connector._connected = True
+        connector._cleanup_resources = AsyncMock()
+
+        await connector.disconnect()
+        await connector.disconnect()
+
+        connector._cleanup_resources.assert_called_once()
+        assert connector._connected is False
+
+    @pytest.mark.asyncio
+    async def test_disconnect_reentrant_call_skips_second_cleanup(self, mock_sandbox_modules):
+        """A re-entrant disconnect during ``_cleanup_resources`` must not run cleanup twice.
+
+        Mirrors the regression test added in PR #1412 for ``BaseConnector``: while
+        the first ``disconnect`` is awaiting ``_cleanup_resources``, an
+        ``AsyncExitStack`` teardown (or another caller) invokes ``disconnect``
+        again. The guard at the top of ``disconnect`` must already see
+        ``_connected == False`` so the second call returns early instead of
+        re-running cleanup against an already-stopped sandbox.
+        """
+        sandbox_options = SandboxOptions(api_key="test-api-key")
+        connector = SandboxConnector("npx", ["test-command"], e2b_options=sandbox_options)
+        connector._connected = True
+
+        cleanup_call_count = 0
+
+        async def reentrant_cleanup():
+            nonlocal cleanup_call_count
+            cleanup_call_count += 1
+            # Simulate a concurrent/re-entrant disconnect firing while we
+            # are still inside the first call's cleanup await.
+            await connector.disconnect()
+
+        connector._cleanup_resources = reentrant_cleanup
+
+        await connector.disconnect()
+
+        assert cleanup_call_count == 1
+        assert connector._connected is False
+
 
 class TestSandboxConnectorCleanup:
     """Tests for SandboxConnector cleanup methods."""
--- a/libraries/python/tests/unit/test_websocket_connector.py
+++ b/libraries/python/tests/unit/test_websocket_connector.py
@@ -1,0 +1,90 @@
+"""
+Unit tests for the WebSocketConnector class.
+"""
+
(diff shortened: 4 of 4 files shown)


truffle-dev:
@khandrew1 — friendly nudge on this one. It extends the re-entrant `disconnect()` guard that #1412 landed for `BaseConnector` into the two subclass overrides (`WebSocketConnector`, `SandboxConnector`) that bypass the parent's flag-flip; without it, an `AsyncExitStack` teardown firing during cleanup still re-enters `_cleanup_resources()`.

CI red is the expected fork-secrets gap on `tests/docs/*`. Happy to rebase, narrow the diff, or split off anything that would simplify review.

truffle-dev:
Closing for now to keep the backlog tidy — happy to revisit if priorities change.

PR #1437, opened 2026-05-03
feat(inspector): support custom base URL for OpenAI-compatible providers

### LINKED ISSUE #1435: Inspector: OpenAI Provider should be extended to support alternative Base URLs.
* We have a custom OpenAI handler for the Inspector's chat.
* Custom providers like LM Studio and OpenRouter are OpenAI compatible.
* Allows for more customizability and options when using our inspector

siiddhantt:
## Language / Project Scope

- [x] TypeScript

## Changes

Adds a **Base URL** field to the inspector's LLM configuration dialog, allowing OpenAI-compatible providers (LM Studio, OpenRouter, Ollama, etc.) to be used by pointing the inspector at a custom API endpoint.

## Implementation Details

1. Added `baseUrl?: string` to `ProviderConfig` (`llm/types.ts`) and `LLMConfig` (`chat/types.ts`)
2. Replaced hardcoded `DEFAULT_ENDPOINT` in `openai.ts` with `OPENAI_BASE_URL = "https://api.openai.com/v1"`; endpoint is now built as `` `${config.baseUrl ?? OPENAI_BASE_URL}/chat/completions` `` — callers supply a full base URL including `/v1`, no double-path segment
3. Forwarded `baseUrl` from `LLMConfig` → `ProviderConfig` at all four call sites: `useChatMessagesClientSide`, `usePropsLLM` (×2), `useSamplingLLM`, and server-side `shared-utils`
4. Added `tempBaseUrl`/`setTempBaseUrl` state in `useConfig`; restored from localStorage on load; only saved when `tempProvider === "openai"` to avoid polluting Anthropic/Google configs
5. Added Base URL `<Input>` in `ConfigurationDialog` (visible only for OpenAI provider); `fetchOpenAIModels` accepts a custom base URL; model list cache key namespaced by base URL
6. Wired `tempBaseUrl`/`onBaseUrlChange` through `ChatTab` → `ChatHeader` → `ConfigurationDialog` (both the header dialog and the landing-form dialog)

---

### Packages Modified

- [x] `inspector`

### Pre-commit Checklist

- [ ] Added or updated tests if needed

---

## Example Usage (Before)

```ts
// OpenAI provider was hardcoded to api.openai.com — no way to point
// the inspector at LM Studio, OpenRouter, or any other compatible endpoint
```

## Example Usage (After)

1. Open the inspector chat config dialog
2. Select "OpenAI" as provider
3. Enter your API key
4. Set Base URL to e.g. http://localhost:1234/v1 (LM Studio)
   or https://openrouter.ai/api/v1 (OpenRouter)
5. Leave blank to use the default OpenAI endpoint

## Testing
- Manual testing: verified Base URL field appears for OpenAI, hidden for Anthropic/Google
- Verified value persists to localStorage and restores correctly on page reload
- Verified models are fetched from the custom endpoint (/models appended, no double /v1)
- Verified switching provider away and back does not wipe the saved base URL
- Verified saving with Anthropic/Google selected does not persist `baseUrl` into those configs

## Backwards Compatibility
- Fully backwards compatible. `baseUrl` is optional throughout — leaving it blank uses the default https://api.openai.com/v1 endpoint, identical to previous behavior.

## Related Issues
- Closes #1435

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/chat/ConfigurationDialog.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/chat/ConfigurationDialog.tsx
@@ -127,9 +127,20 @@ async function fetchOpenAICompatibleModels(
   baseUrl: string,
   apiKey: string
 ): Promise<ModelOption[]> {
-  const response = await fetch(`${baseUrl}/models`, {
-    headers: apiKey ? { Authorization: `Bearer ${apiKey}` } : undefined,
-  });
+  const stripped = baseUrl.replace(/\/+$/, "");
+  const headers: Record<string, string> = {};
+  if (apiKey) headers.Authorization = `Bearer ${apiKey}`;
+
+  let response: Response;
+  try {
+    response = await fetch(`${stripped}/models`, { headers });
+  } catch {
+    // fetch() rejects with a generic TypeError when CORS blocks the response,
+    // when the server is unreachable, or on mixed-content.
+    throw new Error(
+      "Failed to reach the server. Check the URL is correct and that CORS is enabled on the server."
+    );
+  }
 
   if (!response.ok) {
     throw new Error(`Failed to fetch models: ${response.statusText}`);
@@ -137,10 +148,15 @@ async function fetchOpenAICompatibleModels(
 
   const contentType = response.headers.get("content-type") ?? "";
   if (!contentType.includes("application/json")) {
-    throw new Error("Invalid URL");
+    throw new Error("Invalid URL — response was not JSON");
   }
 
   const data = await response.json();
+  if (!Array.isArray(data?.data)) {
+    throw new Error(
+      "Unexpected response format — expected { data: [...] } from the OpenAI-compatible endpoint"
+    );
+  }
   return data.data.map((model: { id: string }) => ({ id: model.id }));
 }
 
@@ -354,16 +370,6 @@ export function ConfigurationDialog({
                     <span>OpenAI</span>
                   </div>
                 </SelectItem>
-                <SelectItem value="openai-compatible">
-                  <div className="flex items-center gap-2">
-                    <img
-                      src={getProviderIcon("openai")}
-                      alt="OpenAI Compatible"
-                      className="w-4 h-4"
-                    />
-                    <span>OpenAI Compatible</span>
-                  </div>
-                </SelectItem>
                 <SelectItem value="anthropic">
                   <div className="flex items-center gap-2">
                     <img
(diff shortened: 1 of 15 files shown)


libraries/typescript/packages/inspector/src/client/components/chat/useConfig.ts
khandrew1, line 248: Make sure to clear the tempBaseUrl here too

libraries/typescript/packages/inspector/src/client/components/chat/ConfigurationDialog.tsx
```diff
@@ -367,6 +384,22 @@ export function ConfigurationDialog({
             </p>
           </div>
 
+          {tempProvider === "openai" && (
```
khandrew1: Instead of surfacing the Base URL here with the direct OpenAI provider, we should instead have a separate "OpenAI Compatible" provider.

This wouldn't change too much in the backend, you would still use the `openai` provider but with a custom config. This makes for a better UX experience.

khandrew1 (changes requested):
Thanks for your contribution!

This is gonna help a ton. I think instead of surfacing the Base URL directly with the OpenAI provider in the UI. We have a separate "OpenAI Compatible" provider. This custom provider is where you can put in an *optional* API key and required base URL. This would still use the `openai.ts` provider but with a different config.

This makes for better UX I believe.

libraries/typescript/packages/inspector/src/client/components/chat/useConfig.ts
siiddhantt: Done. Implemented `openai-compatible` as a fully separate provider in the UI, while `resolveProvider()` maps it to `openai.ts` and tempBaseUrl is cleared here.

libraries/typescript/packages/inspector/src/client/components/chat/ConfigurationDialog.tsx
siiddhantt: Base URL field is now exclusive to the new "OpenAI Compatible" provider; the original OpenAI provider entry is unchanged.

siiddhantt:
Thanks @khandrew1  for the quick review and the clear direction - the dedicated provider approach is definitely the right call for UX!

khandrew1 (approved):

khandrew1:
@siiddhantt Thanks so much for the contribution! Merged.

siiddhantt:
Thanks @khandrew1 for the thorough review! Looking forward to contributing more.

PR #1455, opened 2026-05-06
fix: prefix router resource names

### LINKED ISSUE #1438: Bug: MCPServer.include_router(prefix=...)` prefixes tools and prompts but not resources`
**Describe the bug**

`MCPServer.include_router(router, prefix="...")` correctly prefixes **tool** and **prompt** names but silently drops the prefix for **resources**. Inside `MCPServer.include_router` (`libraries/python/mcp_use/server/server.py`, around lines 315–325), the prefix-aware `resource_name` is computed but never used, the call site passes `name=resource.name` instead of `name=resource_name`:

```python
# libraries/python/mcp_use/server/server.py
for resource in router.resources:
    resource_name = resource.name or getattr(resource.fn, "__name__", "unknown")
    if prefix:
        resource_name = f"{prefix}_{resource_name}"   # ← computed
    self.resource(
        uri=resource.uri,
        name=resource.name,                           # ← but resource.name is used (BUG)
        description=resource.description,
        mime_type=resource.mime_type,
    )(resource.fn)
```
Compare with the surrounding tool branch (line 308) and prompt branch (line 334) in the same function, which both correctly pass the prefixed tool_name / prompt_name.

To Reproduce

```python
from mcp_use.server import MCPServer, MCPRouter

router = MCPRouter()

@router.tool()
def add(a: int, b: int) -> int:
    return a + b

@router.resource("config://app")
def get_config() -> str:
    return "cfg"

@router.prompt()
def greet(name: str) -> str:
    return f"Hi {name}"

server = MCPServer(name="t")
server.include_router(router, prefix="math")

print("TOOLS    :", [t.name for t in server._tool_manager.list_tools()])
print("RESOURCES:", [r.name for r in server._resource_manager.list_resources()])
print("PROMPTS  :", [p.name for p in server._prompt_manager.list_prompts()])
```
Output on main:

```bash
TOOLS    : ['math_add']
RESOURCES: ['get_config']     # ← prefix missing
PROMPTS  : ['math_greet']
```
Expected behavior

Resources should be prefixed consistently with tools and prompts:

```bash
TOOLS    : ['math_add']
RESOURCES: ['math_get_config']
PROMPTS  : ['math_greet']
```

This is the behavior the include_router docstring already documents ("Optional prefix to add to all tool names (e.g., 'math' -> 'math_add')") and the only behavior that lets routers be safely composed without resource-name collisions across modules.

Suggested fix

One-line change in libraries/python/mcp_use/server/server.py, line 322:

```python
             self.resource(
                 uri=resource.uri,
-                name=resource.name,
+                name=resource_name,
                 description=resource.description,
                 mime_type=resource.mime_type,
             )(resource.fn)
```

Plus a regression test under libraries/python/tests/unit/server/test_router.py covering prefixing for tools / resources / prompts (no router tests exist today).

Desktop

- OS: Fedora 44 kernel 6.19.14-300.fc44.x86_64 (logic bug, platform-independent)
- Python: 3.11+ (any supported version)
- mcp-use: latest main (commit 1e19af9a) and v1.7.0 release

Additional context

Happy to send a follow-up PR with the one-line fix and unit-test coverage for include_router (no dedicated test module exists for it today).

zxyasfas:
## Summary

Fixes inconsistent router prefixing for resources.

## Problem

`MCPServer.include_router(router, prefix=...)` prefixes tools and prompts, but resources still keep their unprefixed name. That makes composed routers behave inconsistently and can cause resource-name collisions when multiple routers expose similarly named resources.

## Changes

- Use the computed prefixed resource name when `MCPServer.include_router(...)` registers router resources.
- Apply the same resource-name prefixing in `MCPRouter.include_router(...)` for nested routers.
- Add regression coverage for server-level and nested-router resource prefixing.

Fixes #1438.

## Validation

- `.venv\Scripts\python.exe -m pytest libraries\python\tests\unit\server\test_router_prefixes.py`
- `.venv\Scripts\python.exe -m pytest libraries\python\tests\unit\server\test_server_config.py libraries\python\tests\unit\server\test_router_prefixes.py`
- `.venv\Scripts\ruff.exe check libraries\python\mcp_use\server\server.py libraries\python\mcp_use\server\router.py libraries\python\tests\unit\server\test_router_prefixes.py`
- `.venv\Scripts\python.exe -m py_compile libraries\python\mcp_use\server\server.py libraries\python\mcp_use\server\router.py libraries\python\tests\unit\server\test_router_prefixes.py`

### DIFF
--- a/libraries/python/mcp_use/server/router.py
+++ b/libraries/python/mcp_use/server/router.py
@@ -229,7 +229,19 @@ class MCPRouter:
             )
 
         for resource in router._pending_resources:
-            self._pending_resources.append(resource)
+            resource_name = resource.name or getattr(resource.fn, "__name__", "unknown")
+            if combined_prefix:
+                resource_name = f"{combined_prefix}_{resource_name}"
+
+            self._pending_resources.append(
+                PendingResource(
+                    fn=resource.fn,
+                    uri=resource.uri,
+                    name=resource_name,
+                    description=resource.description,
+                    mime_type=resource.mime_type,
+                )
+            )
 
         for prompt in router._pending_prompts:
             prompt_name = prompt.name or getattr(prompt.fn, "__name__", "unknown")
--- a/libraries/python/mcp_use/server/server.py
+++ b/libraries/python/mcp_use/server/server.py
@@ -319,7 +319,7 @@ class MCPServer(FastMCP):
                 resource_name = f"{prefix}_{resource_name}"
             self.resource(
                 uri=resource.uri,
-                name=resource.name,
+                name=resource_name,
                 description=resource.description,
                 mime_type=resource.mime_type,
             )(resource.fn)
--- a/libraries/python/tests/unit/server/test_router_prefixes.py
+++ b/libraries/python/tests/unit/server/test_router_prefixes.py
@@ -1,0 +1,37 @@
+from mcp_use.server import MCPRouter, MCPServer
+
+
+def test_server_include_router_prefixes_resources_like_tools_and_prompts():
+    router = MCPRouter()
+
+    @router.tool()
+    def add(a: int, b: int) -> int:
+        return a + b
+
+    @router.resource("config://app")
+    def get_config() -> str:
+        return "cfg"
+
+    @router.prompt()
+    def greet(name: str) -> str:
+        return f"Hi {name}"
+
+    server = MCPServer(name="test-server")
+    server.include_router(router, prefix="math")
+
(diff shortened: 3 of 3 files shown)


zxyasfas:
The PR is scoped to the router/server prefixing path only. Current checks are green, and the added regression test covers both direct MCPServer.include_router(...) resource registration and nested MCPRouter.include_router(...) resource registration.

PR #1462, opened 2026-05-08
fix(inspector): pin iframe width in inline chat mode (#1459)

### LINKED ISSUE #1459: Inspector chat: MCP Apps renderer should use a fixed chat-column width, not shrink with widget-reported width
https://github.com/user-attachments/assets/2b068d50-9720-4661-a6e0-e41d979426a8

**Describe the bug**
In the inspector's chat playground, the MCP Apps renderer (iframe) shrinks horizontally when a widget reports a smaller size via `window.openai.setSize`. The widget container's width tracks whatever the embedded app reports, so apps that don't fill the chat column render at an inconsistent, smaller width than the surrounding messages.

This diverges from how other MCP Apps hosts (ChatGPT, Claude) render widgets in chat: they pin the widget to a fixed chat-column width and only let height respond to the widget's reported size.

**To Reproduce**
1. Open the inspector and connect to the Excalidraw MCP server at `https://mcp.excalidraw.com/mcp`.
2. In the chat playground, ask it to draw a picture.
3. When the widget renders, notice the view shrinks horizontally — narrower than the chat column and the surrounding messages.
4. Compare to the same widget in ChatGPT or Claude, where width stays pinned to the chat column.

**Expected behavior**
In inline (chat) mode, the renderer should:
- Keep the iframe width pinned to the chat column (current cap: `inlineMaxWidth` = 768px desktop/tablet, device width on mobile).
- Only apply the height portion of `onsizechange` events.
- Ignore width changes from the widget bridge while `displayMode === "inline"`.

Width-driven sizing can remain for `pip` / `fullscreen` modes if needed.

**Where this lives in the code**
`packages/inspector/src/client/components/MCPAppsRenderer.tsx`

- L654–696: `bridge.onsizechange` currently applies both `width` and `height` to the iframe.
- L682–685: width is written directly to `iframeEl.style.width` as `min(${adjustedWidth}px, 100%)`, which shrinks the iframe below the chat column cap.
- L216 / L1090–1098: `inlineMaxWidth` and `iframeStyle` already define the intended fixed-width behavior — the bridge handler just overrides it.

**Suggested fix**
Drop the width branch in `onsizechange` for inline mode (only animate height). The existing `iframeStyle` (`width: 100%`, `maxWidth: inlineMaxWidth`) already gives the desired fixed-width-with-cap behavior. Remove `width 300ms ease-out` from the inline transition since width should no longer change.

**Screenshots**
_Before/after screenshots of the Excalidraw widget rendered in the inspector chat vs. ChatGPT/Claude — please attach._

**Additional context**
- Inspector package: `@mcp-use/inspector` (`libraries/typescript/packages/inspector`).
- Reproduction server: `https://mcp.excalidraw.com/mcp`.
- Related: SEP-1865 MCP Apps renderer.

dhruvtiwari6:
Fixes #1459

https://github.com/user-attachments/assets/1aeca2f3-ba06-4519-a21a-c2d143f6b5b9

## Before / After

| Before | After |
|--------|-------|
|  |  |

## Problem
In the chat playground, `bridge.onsizechange` was applying both width 
and height from widget-reported sizes. Since the handler already guards 
with `if (displayModeRef.current !== "inline") return`, the width 
override was always running in inline mode — shrinking the iframe to 
whatever the widget reported instead of the chat column width.

## Fix
- Removed the width block from `onsizechange` in inline mode
- Removed `width 300ms ease-out` from the transition string since width no longer animates
- The existing `iframeStyle` (`width: 100%`, `maxWidth: inlineMaxWidth`) now stays in control of width

## Tested with
- Excalidraw MCP server at `https://mcp.excalidraw.com/mcp`
- Width now stays pinned to full chat column
- Height still responds dynamically to widget reports


khandrew1 (commented):
(superseded — see inline comments below)

khandrew1 (commented):
Diagnosis is correct and the fix is minimal. A few cleanup items inline. Two non-inline notes:

1. **PR description inconsistency** — the description says *"Removed `width 300ms ease-out` from the transition string since width no longer animates,"* but `iframeStyle.transition` at `MCPAppsRenderer.tsx:1094-1097` still reads `"height 300ms ease-out, width 300ms ease-out"`. Either drop the width segment to match, or update the description (keeping it is arguably fine since the chat column itself can resize on window/panel toggle).

2. **Changeset missing.** Per `libraries/typescript/CLAUDE.md`, all changes require a changeset (CI-enforced). Run `pnpm changeset` and select a patch bump for `@mcp-use/inspector`.

libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
```diff
@@ -651,49 +651,31 @@ function MCPAppsRendererBase({
       return {};
     };
 
+    
+
```
khandrew1: Stray blank lines (one with trailing whitespace, one empty) inserted above the handler — please remove. Likely an artifact of the indentation collapse below; `pnpm format` should clear them.

libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
```diff
@@ -651,49 +651,31 @@ function MCPAppsRendererBase({
...
     bridge.onsizechange = ({ width, height }) => {
```
khandrew1: `width` is destructured but never read. Either omit it (`({ height }) => { ... }`) or rename to `_width` to silence the unused warning.

libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
```diff
@@ -651,49 +651,31 @@ function MCPAppsRendererBase({
...
-        adjustedWidth +=
-          parseFloat(style.borderLeftWidth) +
-          parseFloat(style.borderRightWidth);
-      }
-      if (adjustedHeight !== undefined && isBorderBox) {
-        adjustedHeight +=
-          parseFloat(style.borderTopWidth) +
-          parseFloat(style.borderBottomWidth);
-      }
+  if (displayModeRef.current !== "inline") return;
```
khandrew1: **Indentation regression (must fix).** The handler body is indented 2 spaces inside an 8-space block, breaking the surrounding style — compare `bridge.onloggingmessage` and `bridge.onsetwidgetstate` above which use 6-space body indent. The closing `};` at the end of the handler is misaligned for the same reason. Prettier/ESLint pre-commit should have caught this — please run `pnpm format` and re-commit.

Also: the prior comment
```ts
// Use ref so this closure always reads the current displayMode even
// though it was captured when the bridge was first created.
```
was load-bearing — it explains *why* `displayModeRef.current` is read instead of the closed-over `displayMode`. Please restore it above this line.

libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
```diff
@@ -651,49 +651,31 @@ function MCPAppsRendererBase({
...
+  const iframeEl = iframe;
+  if (!iframeEl || height === undefined) return;  // width no longer needed
```
khandrew1: `// width no longer needed` describes what the *change* did, which belongs in the commit message — not the source. The condition `height === undefined` is self-documenting; drop the trailing comment.

khandrew1 (changes requested):

dhruvtiwari6:
Thanks @khandrew1 for addressing up me the issues and I will look into them

dhruvtiwari6:
@khandrew1 

All feedback addressed:
- Removed `width 300ms ease-out` from the transition string
- Added patch changeset for `@mcp-use/inspector`
- Restored the load-bearing `displayModeRef` comment
- Dropped trailing `// width no longer needed` comment
- Ran `pnpm format` throughout

Ready for re-review!

khandrew1 (approved):

PR #1483, opened 2026-05-12
fix(server/oauth): use URL constructor to avoid double slash in metadata proxy

### LINKED ISSUE #1482: oauthAuth0Provider: DCR-direct mode produces malformed metadata URL (double slash → 404)
**Describe the bug**

When `MCPServer` is configured with `oauthAuth0Provider()` (DCR-direct mode), requests to `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration` on the running MCP server return:

```json
{"error":"server_error","error_description":"Failed to fetch provider metadata: 404"}
```

Root cause is a string-concatenation bug in the upstream-metadata proxy that produces a URL with a double slash before `.well-known`, which Auth0 (and most strict origins) return 404 for.

**Root cause**

At `libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts:311`:

```ts
const issuer = oauth.getIssuer();
const metadataUrl = `${issuer}/.well-known/oauth-authorization-server`;
```

And at `libraries/typescript/packages/mcp-use/src/server/oauth/providers/auth0.ts:19-21`:

```ts
this.baseUrl = `https://${config.domain}`;
// Auth0 issuer always includes a trailing slash
this.issuer = `${this.baseUrl}/`;
```

The Auth0 provider correctly stores the canonical issuer with a trailing slash (this is required — Auth0 puts the trailing-slashed issuer in the `iss` claim of every JWT, and `verifyToken` at `auth0.ts:52` validates against the same string). The route handler then concatenates a path that itself starts with `/`. The resulting URL is `https://<tenant>.auth0.com//.well-known/oauth-authorization-server`. Auth0 returns **HTTP 404** for any path with a leading double slash.

**To Reproduce**

1. Configure an `MCPServer` with `oauth: oauthAuth0Provider({ domain: '<tenant>.auth0.com', audience: '<api-id>' })`.
2. Deploy (or run locally).
3. `curl https://<server>/.well-known/oauth-authorization-server` → returns the 500/`server_error` JSON above.
4. The MCP server logs `[OAuth] Fetching metadata from provider: https://<tenant>.auth0.com//.well-known/oauth-authorization-server` (note the `//`).
5. `curl -o /dev/null -w "%{http_code}\n" https://<tenant>.auth0.com//.well-known/oauth-authorization-server` → `404`.
6. `curl -o /dev/null -w "%{http_code}\n" https://<tenant>.auth0.com/.well-known/oauth-authorization-server` → `200` (the URL the framework should be hitting).

**End-to-end impact**

Claude Desktop's DCR client fetches the MCP server's `oauth-protected-resource` metadata, finds the MCP server itself listed as the `authorization_servers` proxy, then fetches that proxy's `oauth-authorization-server` endpoint, gets the 500, and aborts the OAuth flow. The deployed server is unauthenticatable until this is fixed.

**Expected behavior**

The metadata-proxy route should produce a correctly-normalized URL regardless of whether the configured issuer string carries a trailing slash, and should successfully proxy Auth0's upstream `oauth-authorization-server` / `openid-configuration` metadata.

**Suggested fix**

Use the `URL` constructor, which handles base+path normalization regardless of trailing slash:

```ts
// routes.ts:311
const metadataUrl = new URL("/.well-known/oauth-authorization-server", issuer).toString();
```

Equivalent alternative: strip trailing slash before concat (`${issuer.replace(/\/$/, "")}/.well-known/...`).

PR incoming.

**Environment**

- mcp-use version: 1.27.1 (latest as of 2026-05-12)
- Runtime: Node 22-alpine on Manufact Cloud (also reproduces locally)
- OAuth provider: Auth0 (DCR-direct mode via `oauthAuth0Provider()`)

**Additional context**

Reproduced with two independent confirmations:
- Direct curl against Auth0 with and without the double slash (200 vs. 404).
- Source inspection of published `mcp-use@1.27.1` tarball confirms the same concat is present in shipped JS (`package/dist/src/server/index.cjs`).

polae:
## Summary

Fixes the metadata-proxy URL construction in DCR-direct OAuth mode (`oauthAuth0Provider`, and any other provider whose canonical issuer ends with `/`).

Before: `\`${issuer}/.well-known/oauth-authorization-server\`` → `https://tenant.auth0.com//.well-known/...` → Auth0 returns 404 → MCP server's `/.well-known/oauth-authorization-server` returns `500 server_error`, which aborts Claude Desktop's OAuth flow end-to-end.

After: `new URL("/.well-known/oauth-authorization-server", issuer).toString()` → correct URL regardless of whether `issuer` carries a trailing slash.

## Why this matters

`auth0.ts:21` deliberately stores the issuer with a trailing slash because that's the canonical form Auth0 puts in the `iss` claim of every JWT, and `verifyToken` (`auth0.ts:52`) validates against this exact string. So the issuer **must** end with `/`. The bug was in the consumer not accounting for that.

## Repro

Closes #1482 — full reproduction steps + direct-curl evidence (200 vs 404) are there.

## Test plan

- [ ] Manual: deploy `MCPServer` with `oauthAuth0Provider({ domain, audience })`, curl `/.well-known/oauth-authorization-server`, confirm the response is Auth0's actual metadata JSON (issuer, registration_endpoint, etc.) instead of `{"error":"server_error",...}`.
- [ ] Unit test: not added (no existing test file for `src/server/oauth/routes.ts`). Happy to add one — a mock fetch + assertion on the constructed URL would cover this. Let me know if you'd prefer that before merge.

## Notes

While confirming the bug I noticed `routes.ts` has similar string concatenations elsewhere (lines 295/296/297/385) that build `${baseUrl}/...` URLs. Those use a different `baseUrl` (the MCP server's own URL, not the upstream issuer), so they may be fine — but worth a grep if anyone wants to harden the file further. Out of scope for this PR.

## Changeset

Patch bump on `mcp-use`. Included in `libraries/typescript/.changeset/fix-oauth-metadata-double-slash.md`.

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts
@@ -308,10 +308,7 @@ export function setupOAuthRoutes(
     // DCR-direct mode: proxy to upstream
     try {
       const issuer = oauth.getIssuer();
-      const metadataUrl = new URL(
-        "/.well-known/oauth-authorization-server",
-        issuer
-      ).toString();
+      const metadataUrl = `${issuer.replace(/\/+$/, "")}/.well-known/oauth-authorization-server`;
       console.log(`[OAuth] Fetching metadata from provider: ${metadataUrl}`);
       const response = await fetch(metadataUrl);
 
--- a/libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts
@@ -308,7 +308,10 @@ export function setupOAuthRoutes(
     // DCR-direct mode: proxy to upstream
     try {
       const issuer = oauth.getIssuer();
-      const metadataUrl = `${issuer}/.well-known/oauth-authorization-server`;
+      const metadataUrl = new URL(
+        "/.well-known/oauth-authorization-server",
+        issuer
+      ).toString();
       console.log(`[OAuth] Fetching metadata from provider: ${metadataUrl}`);
       const response = await fetch(metadataUrl);
 
--- a/libraries/typescript/.changeset/fix-oauth-metadata-double-slash.md
+++ b/libraries/typescript/.changeset/fix-oauth-metadata-double-slash.md
@@ -1,0 +1,11 @@
+---
+"mcp-use": patch
+---
+
+Fix double slash in OAuth metadata proxy URL when using DCR-direct providers (e.g. `oauthAuth0Provider`).
+
+Issuers canonically end with a trailing slash (e.g. `https://tenant.auth0.com/`), but the route handler at `src/server/oauth/routes.ts` was concatenating `/.well-known/oauth-authorization-server` directly onto the issuer, producing `https://tenant.auth0.com//.well-known/...`. Auth0 returns `404` for any path with a leading double slash, which broke OAuth metadata discovery end-to-end (the MCP server returned `{"error":"server_error","error_description":"Failed to fetch provider metadata: 404"}` for `/.well-known/oauth-authorization-server` and `/.well-known/openid-configuration`, aborting Claude Desktop's OAuth flow).
+
+Switched the concatenation to `new URL(path, issuer)`, which normalizes the join regardless of trailing slash on the base.
+
+Closes #1482.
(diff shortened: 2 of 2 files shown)


libraries/typescript/packages/mcp-use/src/server/oauth/routes.ts
```diff
@@ -308,7 +308,10 @@ export function setupOAuthRoutes(
     // DCR-direct mode: proxy to upstream
     try {
       const issuer = oauth.getIssuer();
-      const metadataUrl = `${issuer}/.well-known/oauth-authorization-server`;
+      const metadataUrl = new URL(
```
khandrew1: This will actually break our other Auth providers like supabase as it'll strip our `/auth/v1` from its base URL. I'd recommend normalizing the trailing slash instead of using `new URL()`.

```
const metadataUrl = `${issuer.replace(/\/+$/, "")}/.well-known/oauth-authorization-server`;
```

khandrew1 (changes requested):

khandrew1 (approved):

PR #1489, opened 2026-05-13
fix(inspector): keep tab count badges visible in collapsed mode

### LINKED ISSUE #1480: Inspector: Keep numeric tab badge visible in collapsed (icon-only) header
Make the per-tab count badge (e.g. Tools 18, Prompts 4, Notifications 2) always visible next to the tab icon, including when the header is in collapsed/icon-only mode. Today the badge is suppressed when the tab row collapses and is replaced by a small dot (for sampling/elicitation/notifications) or nothing at all.

Currently:

Image

Arpitm544:
Closes #1480. This PR refactors TabsTrigger to use a generic 'badge' prop and ensures it remains visible even when the tab bar is collapsed/icon-only.

### DIFF
--- a/libraries/typescript/.changeset/improve-inspector-scrollbars.md
+++ b/libraries/typescript/.changeset/improve-inspector-scrollbars.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/inspector": patch
+---
+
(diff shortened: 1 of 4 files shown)


Arpitm544:
Replaced with clean PR #1490 containing only #1480 changes.

PR #1490, opened 2026-05-13
fix(inspector): keep tab count badges visible in collapsed mode

### LINKED ISSUE #1480: Inspector: Keep numeric tab badge visible in collapsed (icon-only) header

Arpitm544:
Closes #1480. This PR refactors TabsTrigger to use a generic 'badge' prop and ensures it remains visible even when the tab bar is collapsed/icon-only. (Cleaned up from previous PR #1489)

Screenshot 2026-05-13 at 12 25 42 PM

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
@@ -881,6 +881,19 @@ export function LayoutHeader({
                   }
                   const count = getTabCount(tab.id, selectedServer);
                   const showDot = shouldShowDot(tab.id, count, collapsed);
+                  const badge =
+                    count > 0 ? (
+                      <span
+                        className={cn(
+                          activeTab === tab.id
+                            ? "dark:bg-black"
+                            : "dark:bg-zinc-700",
+                          "ml-1 bg-zinc-200 text-zinc-700 dark:text-zinc-300 text-[10px] px-1.5 py-0.5 rounded-full font-medium"
+                        )}
+                      >
+                        {count}
+                      </span>
+                    ) : null;
 
                   return (
                     <TabsTrigger
@@ -889,6 +902,7 @@ export function LayoutHeader({
                       data-testid={`tab-${tab.id}`}
                       icon={tab.icon}
                       showDot={showDot}
+                      badge={badge}
                       alwaysExpanded={
                         "alwaysExpanded" in tab && tab.alwaysExpanded
                       }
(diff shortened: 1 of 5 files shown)


khandrew1 (changes requested):
Image

I just reviewed this PR, and not sure why the badge shifts sides when selected. it should just always be on the right. 

On another note this might cause the width to be too wide and cause that weird overflow on the top bar. Do you have any thoughts on that?

Arpitm544:
Thanks for giving the feedback I shift the badge from 

icon -> badge -> label 
to 
icon -> label -> badge

so it stays on the right consistently
Screenshot 2026-05-15 at 12 26 43 AM

Yeah good point . Badges make the tabs a little wider, but I see that tab list already support horizontal scrolling if it gets too wide. If need , we can make the badges smaller or shorter in larger counts  later like  99+

libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
```diff
@@ -881,6 +881,19 @@ export function LayoutHeader({
                   }
                   const count = getTabCount(tab.id, selectedServer);
                   const showDot = shouldShowDot(tab.id, count, collapsed);
+                  const badge =
```
khandrew1: This badge syntax is kind of awkward. You should abstract it out into a TabCountBadge component

libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
```diff
@@ -956,6 +958,19 @@ export function LayoutHeader({
                     const tooltipText =
                       count > 0 ? `${tab.label} (${count})` : tab.label;
                     const showDot = shouldShowDot(tab.id, count, collapsed);
+                    const badge =
```
khandrew1: Reduce duplication via a component

khandrew1 (changes requested):
Also add a changeset. Thanks! I think this is better UX for now.

libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
```diff
@@ -889,6 +890,13 @@ export function LayoutHeader({
                       data-testid={`tab-${tab.id}`}
                       icon={tab.icon}
                       showDot={showDot}
+                      badge={
```
khandrew1, line 893: Passing the react component as a prop is kinda awkward. We can't put it in with the children next to the label? If we can't do that then pass the props directly into TabsTrigger and use the badge in the component. I would prefer the first approach though.

khandrew1 (changes requested):

libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx
Arpitm544: Good Point . I was mainly thinking  on the visual behaviour  so initially i went with dedicated badge props for simplicity  . I see now that using child keeps TabsTrigger more flexible . I  updated it to following the first approach

khandrew1 (approved):

PR #1531, opened 2026-05-19
feat(typescript): configurable basePath for MCPServer

khandrew1:
## Language / Project Scope

- [x] TypeScript

## Summary

Adds a `basePath` option to `MCPServer` that mounts every route the framework
registers under one configurable prefix, plus the routing/disk reorganization
needed to make that work cleanly:

- Framework-internal assets live under `/_mcp-use/*` (Next.js-style reserved
  namespace) and on disk under `.mcp-use/`.
- Production widget HTML resolves the server's basePath at load time from
  `window.location.pathname`, so a single built bundle works at any prefix
  with no rebuild.
- A `.mcp-use/server-info.json` handshake file lets out-of-process tooling
  (CLI auto-open in particular) discover the running server's basePath.

```ts
new MCPServer({
  name: "my-server",
  version: "1.0.0",
  basePath: "/api",
});
// MCP transport at /api/mcp + /api/sse
// Inspector UI at /api/inspector
// OAuth flow (authorize, token, register) under /api
// OAuth .well-known/* stays at host root per RFC 8414
// Widget + public assets under /api/_mcp-use/*
```

## Review Readiness

- [x] Scoped to one feature
- [x] Description explains the user-visible API and the routing/disk changes that come with it
- [x] No unrelated formatting or bulk rewrites
- [x] `pnpm build` and tests pass across `mcp-use`, `inspector`, `cli`

## Packages Modified

- [x] `mcp-use` (server)
- [x] `@mcp-use/inspector`
- [x] `@mcp-use/cli`

---

## `mcp-use` (server)

**New API: `basePath` on `MCPServer`**

A single optional config field mounts every route the server registers under one configurable prefix. Default (no prefix) is unchanged.

`basePath` is normalized at construction (`normalizeBasePath` in `utils/server-helpers.ts`): empty / `"/"` collapse to `""`, otherwise must start with `/` and any trailing slash is stripped. Invalid shapes throw immediately rather than producing confusing 404s.

**Single Hono instance via `app.basePath()` clone**

The constructor builds one root Hono, runs OAuth setup synchronously against it (using a lazy `getBaseUrl()` getter so the port doesn't need to be bound yet), then sets `this.app = rootApp.basePath(basePath)`. Hono's `clone()` semantics mean the basePath view shares its parent's router/routes array — so a single `this.app.fetch` dispatches both prefixed routes and the literal-root routes registered before the clone (OAuth `.well-known/*`).

All mount steps (`mountMcp`, `mountWidgets`, `mountInspectorUI`, `setupWidgetRoutes`, custom-routes middleware) receive `this.app` directly and register bare paths; the clone applies the prefix automatically.

**OAuth discovery + WWW-Authenticate**

- `issuer`, `authorization_endpoint`, `token_endpoint`, `registration_endpoint` carry the basePath prefix.
- Both root and path-aware discovery endpoints are registered to match what the MCP SDK's `buildDiscoveryUrls` probes — with `basePath: "/api"` that's both `GET /.well-known/oauth-authorization-server` and `GET /.well-known/oauth-authorization-server/api` (plus the OIDC variants).
- Protected-resource metadata identifies the MCP endpoint specifically: `resource: "<host>/api/mcp"`, served at `/.well-known/oauth-protected-resource/api/mcp` per RFC 9728 §3.1.
- Bearer-auth `WWW-Authenticate` advertises this path-aware `resource_metadata` URL.

**Framework asset namespace: `/_mcp-use/*`**

The underscore marks the path as framework-reserved (mirrors Next.js' `_next/*`), preventing collisions with user routes. Disk layout: built widgets live in `.mcp-use/widgets/`, build manifest at `.mcp-use/manifest.json`.

**Static serving via `serveStatic`**

A single `serveStatic` mount at `/_mcp-use/*` maps to `.mcp-use/` on disk for Node/Bun, with a Deno fallback through `fsHelpers`.

**Widget HTML is basePath-agnostic**

`processWidgetHtml(html, widgetName)` injects an inline script that derives the server's basePath at load time from `window.location.pathname.split("/_mcp-use/")[0]`, so the same baked HTML works whether the server is at the host root or under any prefix — no rebuild.

**Server-info handshake file**

`listen()` writes `.mcp-use/server-info.json` (host, port, basePath, full MCP + inspector URLs) so out-of-process tooling (CLI auto-open) can discover the running server's basePath without an HTTP endpoint. Best-effort; failures swallowed; Deno skipped.

**Misc**

- `createRequestLogger(basePath)` factory strips the prefix before matching noisy-path filters.
- Startup logs (`[MCP] Endpoints`, `[MCP] Tip`, `[INSPECTOR] UI available at`) print prefixed URLs.
- Icon URL resolution lives in `resolveIconUrls(icons, baseUrl, basePath)`.
- `everything`, `oauth/better-auth`, `oauth/supabase`, `ui/mcp-apps` examples demonstrate `basePath`.

---

## `@mcp-use/inspector`

**Vite build emits relative asset URLs**

`vite.config.ts` uses `base: "./"` for the production build and keeps `base: "/inspector"` for `vite dev`. Built HTML emits relative asset URLs (`./assets/...`); Vite's dynamic-chunk loader uses `import.meta.url`, so code-splitting still works. Standalone dev (`http://localhost:3000/inspector`) and the Vite proxy rules are unaffected.

**Runtime basePath helpers (`client/utils/basePath.ts`, new)**

- `getBasePath()` reads `window.__MCP_BASE_PATH__` (injected by the embedding host), normalizes to no trailing slash, returns `""` when absent.
- `inspectorPath(path)` prefixes any inspector-rooted URL — e.g. `inspectorPath("/inspector/api/chat/stream")` → `"/api/inspector/api/chat/stream"` under `basePath: "/api"`, or unchanged when basePath is empty.

**Server-side HTML injection (`server/shared-static.ts`)**

When the embedding host calls `mountInspector(app, { basePath })`, the served `index.html` gets two injections:

- `<base href="${basePath}/inspector/">` placed immediately after `<meta charset>` so relative chunk URLs resolve to `${basePath}/inspector/assets/...`.
- `<script>window.__MCP_BASE_PATH__ = "${basePath}"</script>` so the React client can read the prefix at runtime.

The CDN shell variant injects both as well, kept consistent with the local-file path.

Default `proxyUrl` is built as `${basePath}/inspector/api/proxy`.

The `/inspector/assets/*` handler computes the disk path from the suffix after `/assets/` rather than stripping a literal prefix, so it works when the inspector is sub-mounted at `${basePath}/inspector`.

**Mount API (`server/middleware.ts`, `server/shared-routes.ts`)**

`mountInspector(app, config)` accepts `config.basePath`, passed through to `registerStaticRoutes` and `registerInspectorRoutes`. Hono route registrations themselves use bare `/inspector/...` paths — the basePath compose happens via the host's `.basePath()` clone. The widget container HTML built in `shared-routes.ts` uses `${basePath}/inspector` for the inner fetch URL.

**Client consumers all route through `inspectorPath()`**

- `App.tsx` — React Router `basename` is `${getBasePath()}/inspector`; OAuth callback URL is `${window.location.origin}${inspectorPath("/inspector/oauth/callback")}`.
- `useAutoConnect.ts` — cross-origin `proxyAddress` and the `config.json` fetch.
- `LayoutHeader.tsx` (TunnelBadge) — `dev/info`, `dev/start-tunnel`, `dev/stop-tunnel`.
- `useChatMessages.ts` — chat stream endpoint.
- `MCPAppsRenderer.tsx`, `OpenAIComponentRenderer.tsx` — widget store + `widget-content/{id}` URLs.
- `SandboxedIframe.tsx` — sandbox-proxy iframe URL (configured-origin and same-origin variants).
- `constants/mcp-apps.ts` — `API_ENDPOINTS.WIDGET_STORE` / `WIDGET_CONTENT` are getters (not eager string literals) so they pick up the basePath even if imported before the global is read.
- `telemetry/telemetry.ts` — telemetry proxy URL.
- `InspectorDashboard.tsx`, `Layout.tsx`, `LoginModal.tsx`, `ServerConnectionModal.tsx` — same-origin URL constructions threaded through `inspectorPath()`.

**Tunnel manifest path**

`server/tunnel.ts` reads the subdomain from `.mcp-use/manifest.json` (matches the new disk layout owned by `mcp-use`).

---

## `@mcp-use/cli`

**Auto-open URLs honor the server's basePath**

`waitForServer(port, host, projectPath)` now reads `.mcp-use/server-info.json` (written by `mcp-use` inside its `listen()` callback) on each poll attempt and returns `{ basePath }` instead of a boolean. The auto-open paths in `mcp-use dev`, the watch command, and the deploy/start flow compose MCP, tunnel, and inspector URLs as `http://${host}:${port}${basePath}/mcp` etc. — falls back to `""` (bare paths) when the file is missing or unreadable.

**Build output moved into `.mcp-use/`**

- Built widgets: `.mcp-use/widgets/<slug>/`
- Public assets: `.mcp-use/public/`
- Build manifest: `.mcp-use/manifest.json` (read by both the CLI's tunnel/start commands and the inspector's tunnel module)
- Per-widget build temp: `.mcp-use/.tmp/<slug>/` and `.mcp-use/.tmp/<slug>-metadata/`

**Widget Vite base + asset URLs**

Default Vite `base` for each widget build is `/_mcp-use/widgets/<name>/` (overridden by `MCP_URL` for external-CDN deploys). The generated HTML carries:

- A relative favicon `<link rel="icon" href="../../public/${favicon}" />` — resolves to the same-origin `/_mcp-use/public/` namespace whether the server is at `/` or under a basePath.
- A baked inline script defining `window.__getFile`, `window.__mcpPublicUrl`, and `window.__mcpPublicAssetsUrl`. Two variants:
  - **Same-origin** (default): IIFE reads `window.location.pathname.split("/_mcp-use/")[0]` at load time and prefixes the globals with that, so basePath is resolved per-request without rebuild.
  - **External CDN** (`MCP_URL` set): fully-qualified URLs baked in; basePath doesn't apply.

`MCP_SERVER_URL` continues to root `__mcpPublicUrl` to a separate origin when set.

The post-process step also strips any stray `<base>` tag — Vite's absolute `base` already resolves relative refs, and a runtime origin in `<base>` would defeat static caching.

**Misc**

- Error message in the start command lists `.mcp-use/manifest.json` as the manifest path to check.
- All tunnel subdomain reads/writes point at `.mcp-use/manifest.json`.

---

## Testing

- `tests/unit/server/base-path.test.ts` (new, ~350 lines): `normalizeBasePath` input validation; route-mounting assertions across MCP transport, widgets, public assets, inspector under various basePaths; default-empty case.
- `tests/server/oauth.test.ts`: discovery URL coverage at both root and path-aware locations, protected-resource metadata shape, path-aware `WWW-Authenticate`.
- `tests/unit/server/logging.test.ts`: covers the `createRequestLogger` factory signature.
- Manual end-to-end against the `mcp-apps` example with `basePath: "/api"`: inspector UI loads under `/api/inspector`, widget rendering, MCP transport calls, OAuth proxy endpoints, and `/api/_mcp-use/public/*` assets all resolve; bare `/mcp` and `/inspector` 404 cleanly.

## Breaking Changes

Default behavior (no `basePath` set) is unchanged for end users. The breaking pieces below affect anyone consuming framework-internal paths directly or shipping built widgets from CI.

- **Framework asset namespace moved from `/mcp-use/*` to `/_mcp-use/*`.** Affects anyone reading `__mcpPublicUrl`-rooted URLs by hand or pinning to `/mcp-use/widgets/...`.
- **Build output moved from `dist/` to `.mcp-use/`.** `mcp-use build` now writes `.mcp-use/widgets/`, `.mcp-use/public/`, and `.mcp-use/manifest.json`. Deploy scripts and `.gitignore` entries that pin `dist/resources/` or `dist/mcp-use.json` need updating; `.mcp-use/` should be gitignored.
- **No more HTTP serverinfo discovery.** The handshake is a file write at `.mcp-use/server-info.json` instead.
- **Widget HTML no longer carries a `<base href>`.** Widgets that relied on the previous server-origin `<base>` to resolve relative URLs should switch to absolute paths or read `window.__mcpPublicUrl`.
- **`setupPublicRoutes` is no longer used in production** (still exported for embedders running hand-rolled servers in dev mode).


tonxxd:
@khandrew1 what do you think of using `.mcp-use` for assets instead of `mcp-use` ? I guess there is a reason next uses `.next` ?

khandrew1:
Check resources bundling and `resourceCallbacks` failing

khandrew1:
Store basePath in manifest

khandrew1:
double check `.tmp`

tonxxd:
Will be addressed in v2

PR #1550, opened 2026-05-21
feat(server): add publicLandingPage option for OAuth-protected servers

### LINKED ISSUE #1549: Toggle to show mcp landing page also for auth protected servers
Right now, the MCP landing page is gated with the same middleware that protects the other routes for the authentication. It would be nice to have a flag or a toggle in the server configuration that allows exposing the page publicly, even for auth protected servers.

nehaprasad-dev:
#### language / Project Scope
[x] TypeScript

#### Changes

- publicLandingPage lets the HTML landing page load without OAuth on /mcp; MCP calls still need a bearer token.

#### Implementation Details

- ServerConfig.publicLandingPage
- Bearer middleware skips auth for browser GET only
- Shared isBrowserLandingRequest() with landing handler
- No new deps

#### TypeScript Checklist

- Packages Modified
[x] mcp-use (server)

- Pre-commit Checklist
[x] Tests added/updated

#### Example Usage (Before)

- new MCPServer({ name: "x", version: "1.0.0", oauth: provider });
// GET /mcp in browser → 401

#### Example Usage (After)
new MCPServer({ name: "x", version: "1.0.0", oauth: provider, publicLandingPage: true });
// GET /mcp in browser → landing HTML

#### Documentation Updates
None

#### Testing
- oauth.test.ts: public landing + MCP still 401
- landing-request.test.ts: browser vs MCP Accept headers

#### Backwards Compatibility
- Default false; existing servers unchanged

#### Related Issues
Closes :  #1549

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/oauth/setup.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/oauth/setup.ts
@@ -75,8 +75,7 @@ export async function setupOAuthForServer(
   if (options?.publicLandingPage) {
     const bearerMiddleware = middleware;
     middleware = (c: Context, next: Next) => {
-      const acceptHeader =
-        c.req.header("Accept") || c.req.header("accept") || "";
+      const acceptHeader = c.req.header("Accept") || "";
       if (
         c.req.path === "/mcp" &&
         isBrowserLandingRequest(c.req.method, acceptHeader)
--- a/libraries/typescript/packages/mcp-use/src/server/oauth/setup.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/oauth/setup.ts
@@ -36,7 +36,7 @@ interface OAuthSetupState {
  * @param state - OAuth setup state to track completion
  * @returns Updated OAuth setup state with provider and middleware
  */
-export interface SetupOAuthForServerOptions {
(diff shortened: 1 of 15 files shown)


libraries/typescript/packages/mcp-use/src/server/oauth/setup.ts
khandrew1, line 90: We can do the publicLandingPage check here instead right before the middleware.

```
if (options?.publicLandingPage) {
  app.get("/mcp", async (c, next) => {
    const accept = c.req.header("accept") || "";
    if (isBrowserLandingRequest(c.req.method, accept)) {
      // fall through to the landing handler registered by mountMcp
      return next();
    }
    // not a browser hit — let the bearer middleware handle it
    return next();
  });
}
app.use("/mcp/*", middleware); // unchanged
```

khandrew1 (changes requested):
Thanks for the PR @nehaprasad-dev! I think we can actually take a step back to do this as we don't need this check in the middleware itself. We should be able to do it beforehand.

In `setupOauthForServer` before we set the middleware on `/mcp/*` we can check if public landing page is enabled and if it is coming from a browser. If it is, we can return the landing handler

khandrew1 (approved):

PR #1552, opened 2026-05-21
fix(server/inspector): normalize 0.0.0.0 to localhost in autoConnect URL

### LINKED ISSUE #1551: Inspector: OAuth resource validation fails when mcp-use dev binds to 0.0.0.0 (autoConnect URL ≠ published resource origin)
## Describe the bug

When `mcp-use dev` starts a server with OAuth, the bundled inspector auto-connects to `http://0.0.0.0:3000/mcp` but the same framework publishes the protected-resource metadata with origin `http://localhost:3000`. The MCP SDK's `selectResourceURL` then throws:

```
Protected resource http://localhost:3000 does not match expected http://0.0.0.0:3000/mcp (or origin)
```

The OAuth flow itself completes (token is issued, inspector logs `[OAuthCallback] Authorization successful`) — but the inspector's `useMcp` hook never attaches the token because validation fails on the origin mismatch.

Distinct from mcp-use/mcp-use#1334 (proxy→direct fetch interceptor leakage). Repros on a fresh direct connection, no mode switching.

## Root cause

**Two paths in the framework compute browser-facing URLs from** `serverHost` **— one normalizes loopback, the other doesn't:**

| Path | Source | Normalizes `0.0.0.0`? | Output |
| -- | -- | -- | -- |
| OAuth `resource` metadata | `getServerBaseUrl()` calls `normalizeUrlHost()` ([server-helpers.ts:157](<https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/server/utils/server-helpers.ts#L157>)) | ✅ Yes | `http://localhost:3000` |
| Inspector autoConnect URL | [`inspector/mount.ts:60`](<https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/server/inspector/mount.ts#L60>) constructs `mcpUrl` from raw `serverHost` | ❌ No | `http://0.0.0.0:3000/mcp` |

The MCP SDK's [`checkResourceAllowed`](<https://github.com/modelcontextprotocol/typescript-sdk/blob/main/src/shared/auth-utils.ts>) does strict `requested.origin !== configured.origin` comparison. `useMcp` doesn't override `validateResourceURL`, so the SDK's default fires.

The helper that fixes this already exists in the framework:

```ts
// server-helpers.ts:123 — the framework's own loopback normalizer,
// docstring explicitly motivates the case we're hitting.
function normalizeUrlHost(url: string): string {
  return url.replace(/\/\/0\.0\.0\.0(:|\/|$)/, "//localhost$1");
}
```

The inspector also already knows about loopback equivalence in two other places: [`shared-routes.ts:198`](<https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/inspector/src/server/shared-routes.ts#L198>) (`.origin.replace("0.0.0.0", "localhost")` for widget CSP) and [`LayoutHeader.tsx:165`](<https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx#L165>) (`isLocalhostServerUrl` treats `0.0.0.0` as loopback). Just not on the OAuth path.

## Reproduce

Minimal: any mcp-use server with `oauth: oauthSupabaseProvider({...})` (or any OAuth provider) under `mcp-use dev`. The upstream `mcp-oauth-supabase-template` hits it too — it doesn't even set `baseUrl`, so it relies entirely on framework defaults, and `inspector/mount.ts:60` still produces `0.0.0.0`.

Steps:

1. `pnpm dev --no-open`
2. Open `http://localhost:3000/inspector`
3. Click **Authenticate**, complete OAuth (Google/GitHub/whatever)
4. Inspector lands on `…/inspector/oauth/callback?code=…&state=…` with red error banner

Verified setup: [mcp-use-apps/apps/auth-template](<https://github.com/polae/mcp-use-apps/tree/auth-template/apps/auth-template>) (2026-05-21, mcp-use latest @ `6bf61fd`, inspector v6.0.0, SDK 1.29.0).

## Expected

OAuth resource validation should treat `http://0.0.0.0:3000` and `http://localhost:3000` as the same origin when both are loopback — matching the rest of the framework, which already does.

## Proposed fix (smallest blast radius)

Apply the existing `normalizeUrlHost` (or inline equivalent) at [`inspector/mount.ts:60`](<https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/server/inspector/mount.ts#L60>):

```ts
const hostForBrowser =
  serverHost === "0.0.0.0" || serverHost === "::" ? "localhost" : serverHost;
const mcpUrl = `http://${hostForBrowser}:${serverPort}/mcp`;
```

This keeps the SDK validator strict (good) and the fix is local to one line in one file. PR follows.

## Environment

* mcp-use `latest` @ commit [`6bf61fd`](<https://github.com/mcp-use/mcp-use/commit/6bf61fd>)
* @mcp-use/inspector `6.0.0`
* @modelcontextprotocol/sdk `1.29.0`
* macOS / Chrome / `mcp-use dev` defaults

---

cc @pietrozullo @khandrew1 — please tag with `steven` (my account doesn't have label-write permission on this repo).

polae:
Closes #1551.

## What

The inspector's autoConnect URL is built from raw `serverHost`, while the OAuth `resource` metadata is built via `getServerBaseUrl()` which already normalizes `0.0.0.0` → `localhost` (see [`server-helpers.ts:157`](https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/server/utils/server-helpers.ts#L157)). The SDK's `checkResourceAllowed` does strict origin equality, so the two paths disagree and OAuth-protected `mcp-use dev` apps fail at the inspector's post-callback validation:

```
Protected resource http://localhost:3000 does not match expected http://0.0.0.0:3000/mcp (or origin)
```

This applies the same normalization at `inspector/mount.ts:60`, so the autoConnect URL and the user-visible banner both use the browser-reachable host.

## Why

- Browsers can't resolve `0.0.0.0` as a destination — already documented in the framework's own `normalizeUrlHost` helper.
- The mismatch silently breaks every OAuth-enabled `mcp-use dev` smoke test in the inspector. Token issuance succeeds (`[OAuthCallback] Authorization successful`), but no token is ever attached.
- The framework already treats `0.0.0.0` and `localhost` as equivalent in two other places ([`shared-routes.ts:198`](https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/inspector/src/server/shared-routes.ts#L198), [`LayoutHeader.tsx:165`](https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/inspector/src/client/components/LayoutHeader.tsx#L165)) — this just adds the OAuth path.

## How

One-line normalization at the top of `mountInspectorUI`'s try block; both the autoConnect URL and the banner log read from `hostForBrowser`.

```ts
const hostForBrowser =
  serverHost === "0.0.0.0" || serverHost === "::"
    ? "localhost"
    : serverHost;
```

I deliberately kept this inline rather than importing/exporting `normalizeUrlHost` from `server-helpers.ts` — the existing helper is `function` (not `export`), and exporting it felt out of scope for a behavior fix. Happy to refactor to a shared export if preferred.

## Test plan

- [x] Reproduce against [mcp-use-apps/apps/auth-template](https://github.com/polae/mcp-use-apps/tree/auth-template/apps/auth-template) on `main`: fails with the validation error.
- [x] Apply this fix locally, repeat the flow: branded sign-in renders, Google sign-in completes, callback exchanges code, token is attached, no validation error.
- [ ] CI (workflow runs after submit).

---

cc @pietrozullo @khandrew1 — pairs with #1551; please tag with `steven`.

### DIFF
--- a/libraries/typescript/.changeset/fix-inspector-autoconnect-loopback.md
+++ b/libraries/typescript/.changeset/fix-inspector-autoconnect-loopback.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Fix OAuth-protected `mcp-use dev` flows by normalizing `0.0.0.0` and `::` to `localhost` in the inspector's autoConnect URL, so it matches the resource metadata published by `getServerBaseUrl()` and passes the SDK's strict origin check.
--- a/libraries/typescript/packages/mcp-use/src/server/inspector/mount.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/inspector/mount.ts
@@ -55,9 +55,18 @@ export async function mountInspectorUI(
   try {
     // @ts-ignore - Optional peer dependency, may not be installed during build
     const { mountInspector } = await import("@mcp-use/inspector");
+    // 0.0.0.0 is valid for binding but browsers can't resolve it as a
+    // destination, and the SDK's OAuth resource validator does strict
+    // origin equality — so an autoConnect URL of http://0.0.0.0:PORT/mcp
+    // mismatches the resource metadata published as http://localhost:PORT
+    // (which goes through `normalizeUrlHost` in server-helpers.ts).
+    const hostForBrowser =
+      serverHost === "0.0.0.0" || serverHost === "::"
+        ? "localhost"
+        : serverHost;
     // Auto-connect to the local MCP server at /mcp (SSE endpoint)
     // Use JSON config to specify SSE transport type
-    const mcpUrl = `http://${serverHost}:${serverPort}/mcp`; // Also available at /sse
+    const mcpUrl = `http://${hostForBrowser}:${serverPort}/mcp`; // Also available at /sse
     const autoConnectConfig = JSON.stringify({
       url: mcpUrl,
       name: "Local MCP Server",
@@ -73,7 +82,7 @@ export async function mountInspectorUI(
       serverPort: typeof serverPort === "number" ? serverPort : undefined,
     });
     console.log(
-      `[INSPECTOR] UI available at http://${serverHost}:${serverPort}/inspector`
+      `[INSPECTOR] UI available at http://${hostForBrowser}:${serverPort}/inspector`
     );
     return true;
   } catch (err) {


khandrew1 (approved):
Thanks for the PR @polae!

PR #1575, opened 2026-05-26
Redact OAuth token verification errors

RitwijParmar:
## Summary
- stop returning raw verifier exceptions from the bearer auth middleware
- keep the OAuth discovery `WWW-Authenticate` response intact for MCP clients
- log verifier details server-side and add regression coverage for secret-bearing errors

## Why
When `verifyToken()` throws, the current 401 body includes the exception string. Provider errors can include tenant identifiers, key ids, issuer internals, or other operational detail that should not be reflected to clients.

## Tests
- `pnpm --filter mcp-use test:run tests/server/oauth.test.ts`
- `pnpm --filter mcp-use exec prettier --check src/server/oauth/middleware.ts tests/server/oauth.test.ts`
- `pnpm --filter mcp-use generate:version && pnpm --filter mcp-use exec tsc --noEmit`

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/oauth/middleware.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/oauth/middleware.ts
@@ -94,8 +94,9 @@ export function createBearerAuthMiddleware(
 
       await next();
     } catch (error) {
+      console.warn("[OAuth] Token verification failed:", error);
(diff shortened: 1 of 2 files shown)


RitwijParmar:
Closing this because I found I had already opened an earlier PR for the same OAuth error-redaction issue in this outreach thread. Keeping the prior work as the active contribution to avoid duplicate maintainer noise.

PR #1582, opened 2026-05-27
fix(ts): keep browser entry free of optional agent peers

### LINKED ISSUE #1581: mcp-use/browser entry bundles MCPAgent + langchain, forcing every browser consumer to install @langchain/core and langchain
## Summary

`packages/mcp-use/src/browser.ts` re-exports `MCPAgent`, `RemoteAgent`, and (transitively) the tool-manager classes (`AddMCPServerFromConfigTool`, `ConnectMCPServerTool`, etc.). Those tool managers contain top-level `class X extends StructuredTool` declarations, which means the published `dist/src/browser.js` has hard, top-level `import { StructuredTool } from "@langchain/core/tools"` and `import { ... } from "langchain"` statements.

`@langchain/core` and `langchain` are listed in `packages/mcp-use/package.json` as `peerDependenciesMeta.<name>.optional: true`, so pnpm/npm/yarn do not auto-install them. A consumer that imports anything from `mcp-use/browser` (even just `MCPClient` for OAuth-only use) gets a hard webpack/turbopack build failure when those peers aren't present:

```
Module not found: Can't resolve '@langchain/core/tools'
./node_modules/.pnpm/mcp-use@.../node_modules/mcp-use/dist/src/browser.js (471:1)

> 471 | import { StructuredTool as StructuredTool2 } from "@langchain/core/tools";
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  472 | import { z as z3 } from "zod";
  473 | var AddMCPServerFromConfigTool = class extends StructuredTool2 {
```

Aliasing the peer to a stub does not work — the `extends StructuredTool` is evaluated at module load, so the imports must resolve to something with a real `StructuredTool` class.

## Reproduction

A minimal Next.js (or any webpack/turbopack) app that only imports the OAuth side of `mcp-use/browser`:

```ts
import { MCPClient } from "mcp-use/browser";
import { onMcpAuthorization } from "mcp-use/auth";
```

and does NOT add `@langchain/core` or `langchain` as direct deps. `pnpm install` will skip the optional peers, then `next build` (or `next dev`) fails with the error above. Confirmed on both `c31e794` (canary) and `5ff486f` (current `fix/oauth-flow-popup` PR — same browser bundle shape, no regression introduced by that PR).

## What I expect

Consumers that only need `MCPClient` + OAuth utilities from the browser entry should be able to bundle their app without pulling `@langchain/core` and `langchain` into the dependency graph or the final JS bundle.

## Proposed fix

Split the browser entry so the agent + tool-manager surface lives behind a separate sub-path import. Concretely:

- `mcp-use/browser` — lightweight: `BrowserMCPClient`/`MCPClient`, `BaseConnector`, `HttpConnector`, `MCPSession`, OAuth utilities (`BrowserOAuthClientProvider`, `onMcpAuthorization`, …), notification types. No langchain references.
- `mcp-use/browser/agent` (new) — `MCPAgent`, `RemoteAgent`, `BaseAdapter`, tool managers. Still imports langchain at top level; consumers who want it install the optional peers.

Add a corresponding `exports` entry in `package.json` (`./browser/agent`), update `tsup.config.ts`, and document the split in the changelog / migration notes.

Alternative (smaller / non-breaking): convert the tool-manager class declarations to lazy/deferred form so the langchain imports aren't evaluated at module load (e.g., import the tool managers lazily inside `MCPAgent` methods). This would keep `mcp-use/browser` as a single entry but eliminate the top-level static `extends StructuredTool` evaluation.

## Workaround for now

Add `@langchain/core` and `langchain` to your app's `dependencies` (with versions matching `mcp-use`'s peer ranges — currently `@langchain/core: ^1.1.0`, `langchain: ^1.2.10`). The peer-dep imports then resolve at build time, even though the code paths are dead at runtime for OAuth-only consumers. Bundle size grows by the langchain footprint until the SDK is fixed.

## References

- Browser entry source: [`libraries/typescript/packages/mcp-use/src/browser.ts`](https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/browser.ts)
- The comment at the top of that file states MCPAgent is intentionally browser-exported, which is the deliberate choice we'd be revisiting.
- Surfaced while shipping the OAuth popup/COOP fix in #1580; the langchain breakage is independent of that PR.

jstar0:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Summary

This adds regression coverage for the `mcp-use/browser` optional-peer boundary so browser-only consumers can keep importing the browser client entry without pulling LangChain-backed agent dependencies into that entry.

## Changes

- Added `tests/browser-entry-no-langchain.test.ts` to assert that:
  - `mcp-use/browser` does not statically re-export LangChain-backed agent, adapter, utility, or observability modules.
  - `mcp-use/browser/agent` remains the dedicated subpath for agent exports.
  - built browser ESM, CommonJS, and declaration outputs do not reference optional LangChain or Langfuse peers from the default browser entry.
  - built browser output omits `MCPAgent`, while the browser agent subpath still exposes the agent surface.
- Updated the browser-build comment in `tsup.config.ts` so it describes the current browser entry split instead of the old single-entry shape.

## Implementation Details

The implementation split itself is already present on the current base branch via `src/browser-agent.ts` and the `./browser/agent` package export. After rebasing, this PR is intentionally narrowed to protect that behavior with focused regression tests rather than reintroducing duplicate entry points.

## TypeScript Checklist

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [x] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

## Example Usage (Before)

Browser-only consumers importing from `mcp-use/browser` could fail in bundlers when the default browser entry pulled in LangChain-backed agent modules and optional peer imports.

## Example Usage (After)

The current browser entry split keeps LangChain-backed agent APIs behind `mcp-use/browser/agent`. This PR adds regression coverage so the default `mcp-use/browser` entry stays limited to the browser client, OAuth, connectors, sessions, logging, telemetry, and SDK types.

## Documentation Updates

No docs files were updated. The new regression test documents and enforces the package boundary.

## Verification

Run from `libraries/typescript`:

```bash
pnpm --filter mcp-use build
pnpm --filter mcp-use test:run tests/browser-entry-no-langchain.test.ts tests/browser-react-no-node-deps.test.ts tests/commonjs-compatibility.test.ts tests/inspector-no-langchain.test.ts
pnpm format:check
pnpm lint
```

Manual bundle/declaration scan is covered by `tests/browser-entry-no-langchain.test.ts`.

## Backwards Compatibility

No new export path is introduced by this rebased revision. The current base branch already exposes `mcp-use/browser/agent`; this PR adds regression coverage for that split.

## Related Issues

Closes #1581

### DIFF
--- a/libraries/typescript/packages/mcp-use/tests/browser-entry-no-langchain.test.ts
+++ b/libraries/typescript/packages/mcp-use/tests/browser-entry-no-langchain.test.ts
@@ -1,0 +1,231 @@
+import { existsSync, readFileSync, readdirSync, statSync } from "node:fs";
+import { createRequire } from "node:module";
+import { dirname, join } from "node:path";
+import { fileURLToPath, pathToFileURL } from "node:url";
+import { describe, expect, it } from "vitest";
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+const require = createRequire(import.meta.url);
+const packageDir = join(__dirname, "..");
+
+describe("mcp-use/browser optional peer boundary", () => {
+  it("does not statically re-export LangChain-backed modules from the browser entry", () => {
+    const browserSource = readFileSync(
+      join(packageDir, "src", "browser.ts"),
+      "utf-8"
+    );
+
+    const forbiddenExports: Array<{ name: string; re: RegExp }> = [
+      {
+        name: "MCPAgent from ./agents/mcp_agent.js",
+        re: /export\s*\{[^}]*\bMCPAgent\b[^}]*\}\s*from\s*["']\.\/agents\/mcp_agent\.js["']/,
+      },
+      {
+        name: "adapter barrel from ./adapters/index.js",
+        re: /from\s*["']\.\/adapters\/index\.js["']/,
+      },
+      {
+        name: "agent utility barrel from ./agents/utils/index.js",
+        re: /export\s+\*\s+from\s*["']\.\/agents\/utils\/index\.js["']/,
+      },
+      {
+        name: "observability barrel from ./observability/index.js",
+        re: /from\s*["']\.\/observability\/index\.js["']/,
+      },
+    ];
+
+    const failures = forbiddenExports
+      .filter(({ re }) => re.test(browserSource))
+      .map(({ name }) => name);
+
+    expect(failures, failures.join("\n")).toEqual([]);
+  });
+
+  it("provides a dedicated browser agent subpath for LangChain-backed exports", () => {
+    const pkgJson = JSON.parse(
+      readFileSync(join(packageDir, "package.json"), "utf-8")
+    );
+
+    expect(pkgJson.exports["./browser/agent"]).toEqual({
+      types: "./dist/src/browser-agent.d.ts",
+      import: "./dist/src/browser-agent.js",
+      require: "./dist/src/browser-agent.cjs",
+    });
+
+    const browserAgentSource = readFileSync(
+      join(packageDir, "src", "browser-agent.ts"),
+      "utf-8"
+    );
+    expect(browserAgentSource).toMatch(/\bMCPAgent\b/);
+    expect(browserAgentSource).toMatch(/\bRemoteAgent\b/);
+    expect(browserAgentSource).toMatch(/\bObservabilityManager\b/);
+  });
+
+  it("built browser entries and declarations have no optional peer imports", () => {
+    const browserEntry = join(packageDir, "dist", "src", "browser.js");
+    const browserCjsEntry = join(packageDir, "dist", "src", "browser.cjs");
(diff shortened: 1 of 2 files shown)


khandrew1:
Hey @jstar0! We've been hard at work at v2 which addresses this dependency issue. Thanks!

PR #1627, opened 2026-06-04
Fix/resource tool partial content loss

### LINKED ISSUE #1625: Resource tool ( _convert_resource)returns only last content block and crashes on empty contents
## Bug: Resource tool returns only last content block and crashes on empty contents

  ### Description
  `_convert_resource` in `langchain_adapter.py` has two bugs in its `_arun` method when reading resource contents.

  ### Bug 1 — Only last content block returned (silent data loss)

  The loop overwrites `content_decoded` on every iteration, so only the final block is returned. All previous blocks are silently dropped.

  **Code (lines 240-247):**
  ```python
  for content in result.contents:
      if isinstance(content, bytes):
          content_decoded = content.decode()
      else:
          content_decoded = str(content)

  return content_decoded  # only the last iteration's value
```

  Example: a resource returning ["page 1", "page 2", "page 3"] will only return "page 3" to the agent.

  ###Bug 2 — UnboundLocalError when contents is empty

  If result.contents is an empty list [], the loop body never executes, content_decoded is never assigned, and the return statement crashes:

  UnboundLocalError: cannot access local variable 'content_decoded' before assignment

  Expected Behaviour

  All content blocks should be returned joined together. Empty contents should return an empty string without crashing.

  ## Suggested Fix
```
  parts = []
  for content in result.contents:
      if isinstance(content, bytes):
          parts.append(content.decode())
      else:
          parts.append(str(content))

  return "\n".join(parts)
```

  ## Environment

  - File: libraries/python/mcp_use/agents/adapters/langchain_adapter.py
  - Method: LangChainAdapter._convert_resource → _arun (lines 236–251)

 ## Impact

  - Any agent that reads a multi-block resource gets incomplete data with no warning
  - Any agent that reads a resource with empty contents crashes instead of returning gracefully

Devansh-jpg:
## Changes
  Fix `_convert_resource` returning only the last content block when a resource has multiple contents, and crashing with `UnboundLocalError` when contents is empty.

  ## Implementation Details
  1. Replaced the overwriting `content_decoded` variable with a `parts` list that collects all content blocks across iterations.
  2. Changed `return content_decoded` to `return "\n".join(parts)` outside the loop, so all blocks are joined and returned.
  3. Empty contents now returns `""` instead of crashing.

  ## Example Usage (Before)
  ```python
  # Resource returns 3 content blocks
  result.contents = ["page 1", "page 2", "page 3"]
  # → agent receives only "page 3"  (page 1 and page 2 silently dropped)

  # Resource returns empty contents
  result.contents = []
  # → UnboundLocalError: cannot access local variable 'content_decoded' before assignment
  ```

  ## Example Usage (After)
  ```python
  # Resource returns 3 content blocks
  result.contents = ["page 1", "page 2", "page 3"]
  # → agent receives "page 1\npage 2\npage 3"

  # Resource returns empty contents
  result.contents = []
  # → agent receives ""
  ```

  ## Testing
  - Added `TestLangChainAdapterResourceTool` class in `tests/unit/test_langchain_adapter.py` with 3 tests:
    - `test_single_content_block_returned` — normal case still works
    - `test_multiple_content_blocks_all_returned` — regression for data loss bug
    - `test_empty_contents_returns_empty_string` — regression for UnboundLocalError
  - All 7 tests pass

  ## Backwards Compatibility
  No breaking change. Single content block behaviour is unchanged. Multi-block resources now return complete data instead of partial data.

  ## Related Issues
  Closes #1625

### DIFF
--- a/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
+++ b/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
@@ -243,10 +243,10 @@ class LangChainAdapter(BaseAdapter[BaseTool]):
                         else:
 
                             parts.append(str(content))
-                        "\n".join(parts)
 
 
-                    return parts
+
+                    return  "\n".join(parts)
                 except Exception as e:
                     if self.handle_tool_error:
                         return format_error(e, tool=self.name)  # Format the error to make LLM understand it
--- a/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
+++ b/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
@@ -233,14 +233,20 @@ class LangChainAdapter(BaseAdapter[BaseTool]):
                 logger.debug(f'Resource tool: "{self.name}" called')
                 try:
                     result = await self.tool_connector.read_resource(mcp_resource.uri)
+                    parts =[]
                     for content in result.contents:
                         # Attempt to decode bytes if necessary
                         if isinstance(content, bytes):
-                            content_decoded = content.decode()
-                        else:
-                            content_decoded = str(content)
 
-                    return content_decoded
+                            parts.append(content.decode())
+
+                        else:
+
+                            parts.append(str(content))
+                        "\n".join(parts)
+
+
+                    return parts
                 except Exception as e:
                     if self.handle_tool_error:
                         return format_error(e, tool=self.name)  # Format the error to make LLM understand it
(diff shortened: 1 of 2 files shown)


Arpitm544 (approved):
# AI Code Review

The pull request fixes a critical issue with partial content loss in the resource tool. It ensures that all content blocks are returned, not just the last one, and handles empty contents without crashing. The changes are well-tested and do not introduce any breaking changes.

### Strengths
- Fixes critical issue with partial content loss
- Well-tested changes

PR #1624, opened 2026-06-04
Fix idle cleanup of session refs

puneetdixit200:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Idle session cleanup now mirrors graceful session close cleanup by releasing both resource subscriptions and registered tool/prompt/resource refs for expired sessions. This prevents `sessionRegisteredRefs` entries from remaining after sessions are removed by the idle timeout.

## Review Readiness

Help maintainers review this quickly:

- [x] The PR is scoped to one issue or one clearly related change
- [x] The PR description explains the user-visible problem and the fix
- [x] The diff avoids unrelated formatting, generated files, and bulk rewrites
- [x] I verified the affected package/docs path with the commands listed below
- [x] I checked for existing open PRs that already solve the same issue

## Implementation Details

1. Extended the `startIdleCleanup` server callback shape to include `cleanupSessionRefs`.
2. Calls `cleanupSessionRefs(sessionId)` when an idle session expires, alongside existing subscription cleanup.
3. Added a focused Vitest regression test for expired-session cleanup callbacks.
4. Added a patch changeset for `mcp-use`.

---

## TypeScript Checklist

> Complete this section if your PR includes TypeScript changes

### Packages Modified

Check all packages that were modified:
- [ ] `docs`
- [x] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [x] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

Ensure all of the following have been completed:
- [ ] Ran `pnpm lint:fix` to auto-fix linting issues
- [ ] Ran `pnpm format` to format code with Prettier
- [x] Ran `pnpm build` and build succeeds without errors
- [x] Ran `pnpm changeset` to create a changeset (if this PR includes user-facing changes)
- [x] Added or updated tests if needed
- [x] Updated documentation in `docs/` folder if needed

---

## Python Checklist

> Complete this section if your PR includes Python changes

### Pre-commit Checklist

Ensure all of the following have been completed:
- [ ] Code formatted with `ruff format`
- [ ] Linting passes with `ruff check`
- [ ] Added or updated tests if needed
- [ ] Updated documentation if needed

---

## Example Usage (Before)

```typescript
// Idle timeout removed the session and subscriptions, but left sessionRegisteredRefs for that session.
```

[...]

PR #1662, opened 2026-06-05
fix: clean up sessionRegisteredRefs for ungracefully-closed sessions

### LINKED ISSUE #1623: Memory Leak: startIdleCleanup leaks sessionRegisteredRefs for ungracefully-closed sessions (~326 KB/session, never freed)
**Summary**

`startIdleCleanup` evicts idle sessions from `sessions`/`transports` and calls `cleanupSessionSubscriptions`, but never calls `cleanupSessionRefs`. This leaves the per-session tool/prompt/resource handle maps permanently retained in heap for every session that closes without a graceful DELETE — which is the common case for SSE clients.

Image

**Version:** 1.22.3 → confirmed still present in 1.28.0 (source-verified).

---

**How we found it**

Production Node.js MCP server on Railway (~40 tools). High-traffic replicas showed a smooth, linear ~64 MB/hr heap climb to ~5 GB before OOM restart. Low-traffic replicas in other regions stayed flat — immediately pointing to a per-request or per-session accumulation.

Confirmed via `process.memoryUsage()` on the leaking process: `rss` ≈ `heapTotal` ≈ `heapUsed` (all in the GBs, `external`/`arrayBuffers` negligible) — ruling out native/buffer leaks and pointing squarely at retained JS objects.

**Reproduction** (Node.js, `--expose-gc`):

```js
// Create N sessions with SSE Accept header, abandon them (cancel body = ungraceful close)
// Then wait for one full idle-cleanup sweep cycle
// Then inspect server internals

const server = new MCPServer({ name: "test", version: "1.0.0", sessionIdleTimeoutMs: 3000 });
// register ~40 tools...
await server.listen(PORT);

for (let n = 0; n < 1000; n++) {
  const res = await fetch(`http://localhost:${PORT}/mcp`, {
    method: "POST",
    headers: { "Content-Type": "application/json", Accept: "application/json, text/event-stream" },
    body: JSON.stringify({ jsonrpc: "2.0", id: 1, method: "initialize", params: { protocolVersion: "2025-06-18", capabilities: {}, clientInfo: { name: "t", version: "1" } } }),
  });
  await res.body?.cancel(); // ungraceful drop
}

// AT PEAK (before sweep):
// sessions.size = 1000, sessionRegisteredRefs.size = 1000, heapUsed ≈ 558 MB

await new Promise(r => setTimeout(r, 70_000)); // wait for idle sweep

// AFTER SWEEP — the bug:
// sessions.size = 0   ← correctly cleaned
// sessionRegisteredRefs.size = 1000   ← NEVER cleaned, all still retained
// heapUsed ≈ 557 MB   ← 0.7 MB reclaimed out of 476 MB leaked
```

**Measured retention: ~326 KB per abandoned session, ~476 MB for 1,000 sessions, 0.7 MB reclaimed by the sweep.**

---

**Root cause**

In `startIdleCleanup` ([`[src/server/sessions/session-manager.ts](https://github.com/mcp-use/mcp-use/blob/main/packages/mcp-use/src/server/sessions/session-manager.ts)`](https://github.com/mcp-use/mcp-use/blob/main/packages/mcp-use/src/server/sessions/session-manager.ts)):

```js
// Current code — missing one line
transports?.delete(sessionId);
sessions.delete(sessionId);
mcpServerInstance?.cleanupSessionSubscriptions?.(sessionId);
// ❌ cleanupSessionRefs is never called here
```

The graceful-close path (`onsessionclosed`) correctly calls both:
```js
mcpServerInstance.cleanupSessionSubscriptions?.(sid);
mcpServerInstance.cleanupSessionRefs?.(sid);  // ✅ present in onsessionclosed, absent in idle sweep
```

---

**Fix**

One line added to `startIdleCleanup`:

```js
transports?.delete(sessionId);
sessions.delete(sessionId);
mcpServerInstance?.cleanupSessionSubscriptions?.(sessionId);
mcpServerInstance?.cleanupSessionRefs?.(sessionId); // ← add this
```

**Validated:** calling `cleanupSessionRefs` for the orphaned sessions reclaimed **470 of 476 MB (99%)**, returning heap to baseline.

Image

---

**Workaround (what we shipped)**

Set `stateless: true` on `MCPServer` — no sessions are created, so the leak can't occur. Works for servers that don't use server→client push (notifications, progress, sampling, subscriptions). The `stateless` flag itself is already in the framework; this is just using it as a mitigation.

---

**Impact**
- Any stateful deployment where clients disconnect without sending a DELETE (SSE clients closing tabs, network drops, AI clients moving on)
- Scales with traffic — quiet deployments barely notice it, high-traffic ones OOM
- `sessionIdleTimeoutMs` does not bound it — the sweep runs but frees nothing

Oxygen56:
## Summary

Fixes #1623 — Memory leak where `startIdleCleanup` leaked `sessionRegisteredRefs` (~326 KB/session, never freed).

## Root Cause

The idle cleanup loop called `cleanupSessionSubscriptions` but missed `cleanupSessionRefs`. The graceful-close path (`onsessionclosed`) correctly called both. Ungracefully-closed sessions went through idle eviction which never cleaned their ref maps.

## Fix (2 lines)

1. Added `cleanupSessionRefs` to the `mcpServerInstance` parameter type
2. Added `mcpServerInstance?.cleanupSessionRefs?.(sessionId)` call in the idle eviction loop

## Impact

Prevents unbounded memory growth of ~326 KB per ungracefully-closed session.

Fixes #1623

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/sessions/session-manager.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/sessions/session-manager.ts
@@ -73,6 +73,7 @@ export function startIdleCleanup(
   transports?: Map<string, { close?: () => Promise<void> | void }>,
   mcpServerInstance?: {
     cleanupSessionSubscriptions?: (sessionId: string) => void;
+    cleanupSessionRefs?: (sessionId: string) => void;
   }
 ): NodeJS.Timeout | undefined {
   if (idleTimeoutMs <= 0) {
@@ -109,6 +110,8 @@ export function startIdleCleanup(
         sessions.delete(sessionId);
         // Clean up resource subscriptions for this session if mcpServerInstance provided
         mcpServerInstance?.cleanupSessionSubscriptions?.(sessionId);
+        // Clean up registered refs for this session to prevent memory leaks
+        mcpServerInstance?.cleanupSessionRefs?.(sessionId);
         console.log(
           `[MCP] Cleaned up resource subscriptions for session ${sessionId}`
         );


tonxxd:
Closing in favor of https://github.com/mcp-use/mcp-use/pull/1624 which has the same fix but also the change-set thanks anyway @Oxygen56 for the PR

PR #1667, opened 2026-06-05
Skip OpenAI docs test when API key is unavailable

Rohithmatham12:
## Summary
- Skip the MCPAgent quick-start docs test when OPENAI_API_KEY is not set.
- Keeps forked PR CI from failing on OpenAI-backed docs tests when secrets are unavailable.

## Verification
- git diff --check

Note: This is the same missing-secret failure currently visible in unrelated TypeScript PR test runs.


Rohithmatham12:
Closing this since latest main already includes the OPENAI_API_KEY guard for this docs test. No separate change needed here now.

PR #1671, opened 2026-06-06
Fix: Synchronous Widget Props API & Strict Iframe Origin Security

### LINKED ISSUE #1670: HTML Widgets load empty and flash blank before showing data
**Problem**

Right now, when a developer builds a simple HTML widget (like a greeting card) for MCP Apps, they run into a very frustrating problem: the widget loads completely blank when you first open it.

two main reasons this is happening here:

1. **Broken URL Data**: Developers usually try to read the starting data (like a user's name) from the web address URL (using window.location.search). However, because of the way the Inspector securely sandboxes these widgets, the URL is always empty. So, the widget asks for the data and gets nothing back.

2. **The "Wait a Second" Delay**: The system does eventually send the data to the widget via a message, but it happens after the widget has already loaded. This causes an ugly "flash" — the user sees a blank or broken widget for a split second before the data finally arrives and fills it in.
**Steps to Reproduce**

1. Open the Inspector and run an HTML-based widget (like greeting-card-html) with some test data (e.g., Name: "Alice").

2. Watch the widget closely as it loads.

3. You will see that it initially renders empty or with missing text, and animations might look broken because the data wasn't there when the widget started up.

**How it Should Work (Expected Behavior)**
A widget should have its data the exact millisecond it opens. It shouldn't have to wait for a delayed message, and users shouldn't see a blank screen flash before the widget actually works.

**Solution**

Instead of making the widget wait for a message or try to read an empty URL, the Inspector should "inject" the data directly into the widget before it even turns on.

By creating a simple window.mcpUseProps variable that is injected into the code right away, developers can instantly read their data on launch. The widget will look perfect from the very first frame!

Arpitm544:
### The Problem
When opening a raw HTML widget (like the `greeting-card-html` example) in the Inspector, the widget renders completely blank or is missing its expected text initially. 

There are two main reasons this happens:
1. **Broken URL Data:** Developers usually try to read starting data from the web address URL using `window.location.search`. However, because the Inspector securely sandboxes these widgets using `srcdoc`, the URL is always empty.
2. **The AppBridge Delay:** The system eventually sends the data via a `postMessage` handshake, but it happens asynchronously *after* the widget has already loaded. This causes an ugly "flash of empty content".

Additionally, the Inspector proxy iframe was using `"*"` for its `postMessage` target origin because the sandbox lacked `allow-same-origin`, forcing a `"null"` origin. Broadcasting to `"*"` is a security risk.

### The Solution
1. **Synchronous Injection:** Instead of relying on URL parameters or waiting for a delayed handshake, the Inspector proxy now synchronously injects `window.mcpUseProps = {...}` directly into the widget's HTML `<head>` before it executes. Widgets can instantly read their data on launch.
2. **Strict Origin Security:** Added `allow-same-origin` to the proxy iframe sandbox permissions in development, allowing it to maintain its true network origin. Hardened postMessage handling in SandboxedIframe.tsx with strict sandboxProxyOrigin validation for secure origin targeting.

Resolves #1670

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/MCPAppsRenderer.tsx
@@ -193,6 +193,8 @@ function MCPAppsRendererBase({
   serverRef.current = server;
   const [widgetCsp, setWidgetCsp] = useState<any>(undefined);
   const [widgetPermissions, setWidgetPermissions] = useState<any>(undefined);
+  const [legacyLocationSearch, setLegacyLocationSearch] = useState<string>();
+  const [mcpUseProps, setMcpUseProps] = useState<any>();
   const [prefersBorder, setPrefersBorder] = useState<boolean>(false);
   const [internalDisplayMode, setInternalDisplayMode] =
     useState<DisplayMode>("inline");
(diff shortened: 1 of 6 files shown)


khandrew1:
Closing out, see issue for more details.

PR #1672, opened 2026-06-06
Fix Python connector disconnect idempotency

### LINKED ISSUE #1220: Bug: `close_all_sessions()` closes each session twice — `ValueError: I/O operation on closed pipe` on Windows

Rohithmatham12:
Closes #1220

## Summary
- Mark Python connectors disconnected before awaiting cleanup so concurrent or re-entrant disconnect calls cannot enter cleanup twice.
- Add a regression test that pauses connector cleanup, starts a second disconnect, and verifies cleanup runs exactly once.

## Verification
- `/private/tmp/mcp-use-python-content-venv/bin/python -m pytest libraries/python/tests/unit/test_session.py libraries/python/tests/unit/test_client.py`
- `/private/tmp/mcp-use-python-content-venv/bin/python -m ruff check libraries/python/mcp_use/client/connectors/base.py libraries/python/tests/unit/test_session.py`
- `/private/tmp/mcp-use-python-content-venv/bin/python -m ruff format --check libraries/python/mcp_use/client/connectors/base.py libraries/python/tests/unit/test_session.py`
- `git diff --check`

## Notes
- I checked for open PRs covering #1220 or the disconnect idempotency race and did not find one before opening this.

### DIFF
--- a/libraries/python/mcp_use/client/connectors/base.py
+++ b/libraries/python/mcp_use/client/connectors/base.py
@@ -198,8 +198,8 @@ class BaseConnector(ABC):
             return
 
         logger.debug("Disconnecting from MCP implementation")
-        await self._cleanup_resources()
         self._connected = False
+        await self._cleanup_resources()
         logger.debug("Disconnected from MCP implementation")
 
     async def _cleanup_resources(self) -> None:
--- a/libraries/python/tests/unit/test_session.py
+++ b/libraries/python/tests/unit/test_session.py
@@ -2,11 +2,35 @@
 Unit tests for the MCPSession class.
 """
 
+import asyncio
 import unittest
 from unittest import IsolatedAsyncioTestCase
 from unittest.mock import AsyncMock, MagicMock, PropertyMock, patch
 
+from mcp_use.client.connectors.base import BaseConnector
 from mcp_use.session import MCPSession
+
+
+class DisconnectRaceConnector(BaseConnector):
+    """Connector that lets tests pause cleanup while another disconnect starts."""
+
+    def __init__(self) -> None:
+        super().__init__()
+        self.cleanup_calls = 0
+        self.cleanup_started = asyncio.Event()
+        self.allow_cleanup = asyncio.Event()
+
+    @property
+    def public_identifier(self) -> str:
+        return "disconnect-race"
+
+    async def connect(self) -> None:
+        self._connected = True
+
+    async def _cleanup_resources(self) -> None:
+        self.cleanup_calls += 1
+        self.cleanup_started.set()
+        await self.allow_cleanup.wait()
 
 
 class TestMCPSessionInitialization(unittest.TestCase):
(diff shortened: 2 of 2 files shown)


Rohithmatham12:
All underlying Python test jobs are passing. The only failing checks are the primitive/transport report jobs, and the logs show they fail only while posting PR summary comments from the fork workflow.

Evidence from the failing report jobs:
- Both jobs downloaded their test result artifacts successfully.
- The generated summaries say `✅ All primitive tests passed!` and `✅ All transport tests passed!`.
- Both jobs then fail on `POST https://api.github.com/repos/mcp-use/mcp-use/issues/1672/comments` with `403 Resource not accessible by integration`.

So the actual primitive and transport matrix jobs passed; this appears to be a GitHub token permission issue for report-comment posting on the fork PR.

PR #1687, opened 2026-06-08
fix(cli): cleanly unsubscribe on Ctrl+C (SIGINT)

### LINKED ISSUE #1686: Clean Subscription Shutdown on Ctrl+C (SIGINT)
When using the subscribe command, pressing Ctrl+C immediately terminates the client process without sending an unsubscribe request to the server.

As a result, the server is not informed that the client has disconnected and may continue holding an inactive subscription.

## Steps to Reproduce

1. Start a subscription:

     ` mcp-use client subscribe <resource-uri>  `  

2. Wait until the subscription is active.
3. Press Ctrl+C.

## Current Behavior

The client process exits immediately after receiving SIGINT.

No unsubscribe request is sent to the server before shutdown, leaving the subscription active on the server side.

## Expected Behavior

When Ctrl+C is pressed, the client should:

1. Handle the SIGINT signal.
2. Send an unsubscribe request to the server.
3. Wait for the cleanup to complete.
4. Exit gracefully.

## Impact

Long-lived subscriptions may remain active on the server after the client has exited, resulting in unnecessary resource usage.

Image

Arpitm544:
Closes https://github.com/mcp-use/mcp-use/issues/1686

Screenshot 2026-06-09 at 1 20 56 AM

### DIFF
--- a/libraries/typescript/.changeset/clean-cli-subscribe-sigint.md
+++ b/libraries/typescript/.changeset/clean-cli-subscribe-sigint.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Cleanly unsubscribe from resource subscriptions when stopping the CLI with Ctrl+C.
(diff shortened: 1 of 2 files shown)


khandrew1 (approved):
Thanks for the PR

PR #1691, opened 2026-06-09
feat(cli): managed (no-GitHub) deploy + device-code login

tonxxd:
## Summary
- Adds `mcp-use deploy --managed`: packs the local project into a gzip tarball and uploads it to `POST /servers` with `managed=true`, so a user can deploy **without connecting their own GitHub**. The server's repo is created in the platform-managed org via the installation token; redeploys reuse the linked server (`.mcp-use/project.json`) via `POST /servers/:id/source` + a redeploy.
- Adds `mcp-use login --device-code <code>`: non-interactive auth with a **pre-approved** OAuth device code (RFC 8628). Skips requesting a code and opening a browser — used by the web onboarding flow, which creates + approves the code and embeds it in the agent prompt.

## Implementation
- `src/utils/tarball.ts` — `packProjectTarball()` (nests entries under a wrapper prefix since the backend strips the first path segment; excludes `node_modules`/build dirs/`.env*`) + `sanitizeRepoName()`.
- `src/utils/api.ts` — `uploadMultipart()` helper, `createServerFromManagedUpload()`, `pushSourceToServer()`.
- `src/commands/deploy.ts` — early managed path (`deployViaManagedUpload`) that skips the GitHub-connection/`promptGitHubInstallation` step; `--managed` flag wired in `src/index.ts`.
- `src/commands/auth.ts` — `loginCommand({ deviceCode })` polls the token endpoint directly; `--device-code` wired in `src/index.ts`.

> Requires the matching backend support (managed `GithubInstallation` row + `managed=true` on the multipart create endpoint).

## Test plan
- [x] `vitest` unit tests for `packProjectTarball` (layout + exclusions) and `sanitizeRepoName`
- [x] `tsc --noEmit` passes
- [ ] End-to-end: `mcp-use login --device-code <code>` then `mcp-use deploy --managed -y --org <slug>` against a backend with `GITHUB_MANAGED_*` configured (via this PR's pkg.pr.new pre-release)

### DIFF
--- a/libraries/typescript/.changeset/managed-deploy-cli.md
+++ b/libraries/typescript/.changeset/managed-deploy-cli.md
@@ -2,6 +2,6 @@
 "@mcp-use/cli": minor
 ---
 
-Add `mcp-use deploy --managed` to deploy a local MCP server without connecting your own GitHub. The project source is packed into a tarball and uploaded; the server is created in the platform-managed org and deployed through the normal pipeline (redeploys reuse the linked server).
+Add `mcp-use deploy --managed` to deploy a local MCP server without connecting your own GitHub. The project source is packed into a tarball and uploaded; the server is created in the platform-managed org and deployed through the normal pipeline. Redeploys of a managed project are auto-detected from the linked server, so `--managed` is only needed on the first deploy.
 
 Also add `mcp-use login --device-code <code>` for non-interactive authentication with a pre-approved OAuth device code (used by the web onboarding flow), skipping the browser step.
--- a/libraries/typescript/.changeset/managed-deploy-cli.md
+++ b/libraries/typescript/.changeset/managed-deploy-cli.md
@@ -1,0 +1,7 @@
+---
+"@mcp-use/cli": minor
+---
+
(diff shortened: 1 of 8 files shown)


khandrew1:
@tonxxd i like it, but i think `--managed` may not be that clear for a flag name. what about something like `--local`? what are your thoughts?

tonxxd:
Agree, I think vercel cli also has some special flag for this, will check it out

tonxxd:
Checked Vercel/Netlify/Railway naming for this:

- **Vercel** — CLI deploy is already "from source" by default; no skip-GitHub flag. `--prebuilt` names artifact type (build output vs source), not Git connection.
- **Netlify / Railway** — manual upload is the default CLI path (`netlify deploy`, `railway up`); Git is a separate service source in the dashboard.

`--local` is ambiguous since both flows start on the user's machine. Renaming the flag to **`--no-github`** — matches onboarding copy ("no GitHub setup needed") and the actual user decision. Backend stays `managed=true` / `isManaged` internally; only the CLI flag and docs change.

PR #1700, opened 2026-06-10
fix(mcp-use): move chalk to regular dependencies (statically imported, not optional)

pietrozullo:
## Why

Follow-up to #1697, addressing a point raised by Copilot there: `chalk` was listed in `optionalDependencies`, but it is imported statically in server code (`src/server/logging.ts`, `src/server/utils/server-lifecycle.ts`). Unlike `src/agents/display.ts` — which dynamically imports its display deps with a plain-text fallback — those modules fail at load time if the package is missing. So installs that skip optional packages (`pnpm install --no-optional`, `npm config set optional false`) crash the server, meaning chalk wasn't truly optional.

## Change

- Move `chalk@4.1.2` from `optionalDependencies` to `dependencies` (lockfile importer entry moved accordingly, validated with `pnpm install --frozen-lockfile`)
- Patch changeset included

`cli-highlight` and `redis` stay optional — they are dynamically imported with fallbacks.

### DIFF
--- a/libraries/typescript/.changeset/chalk-regular-dependency.md
+++ b/libraries/typescript/.changeset/chalk-regular-dependency.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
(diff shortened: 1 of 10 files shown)


pietrozullo:
Closing — canary was reset after the 1.30.2 release and this branch now conflicts. Redone on a fresh branch in a follow-up PR.

PR #1705, opened 2026-06-10
feat(python): expose agent run trace

RitwijParmar:
## Why

This addresses part of the long-running request in #262 around seeing which MCP tools were actually called, how many times, and what each tool returned before the model turns those results into a final answer.

Right now users can watch stream updates, but after a run there is no simple object they can hand to a dashboard, debugger, or notebook. That makes it harder to answer questions like: did the model call the same tool twice, did the tool return the expected raw payload, or did the final answer rewrite a tool result in a misleading way?

## What changed

- adds `AgentRunTrace` and `ToolCallRecord` for the Python agent
- records tool name, call id, input, raw output, bounded output preview, timing, counts, final output, and errors
- exposes `agent.get_last_run_trace()`, `agent.get_last_tool_calls()`, and `agent.get_last_tool_outputs()`
- closes unmatched tool calls with a clear trace error instead of leaving them pending
- documents how to inspect tool calls after `agent.run(...)` or `agent.stream(...)`
- adds a unit test covering counts, inputs, outputs, and helper methods

I kept this as an inspectable in-memory trace instead of changing the streaming contract, so existing callers should not see behavior changes.

## Checks

- `uv run --extra dev pytest tests/unit/test_agent.py -q`
- `uv run --extra dev ruff check mcp_use/agents/__init__.py mcp_use/agents/run_trace.py mcp_use/agents/mcpagent.py tests/unit/test_agent.py`
- `uv run --extra dev ruff format --check mcp_use/agents/__init__.py mcp_use/agents/run_trace.py mcp_use/agents/mcpagent.py tests/unit/test_agent.py`
- `uv run --extra dev python -m py_compile mcp_use/agents/__init__.py mcp_use/agents/run_trace.py mcp_use/agents/mcpagent.py tests/unit/test_agent.py`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -255,6 +255,7 @@ jobs:
     if: always() && github.event_name == 'pull_request' && needs.python-transport-tests.result != 'skipped'
     runs-on: ubuntu-latest
     permissions:
+      issues: write
       pull-requests: write
     steps:
       - name: Download all test results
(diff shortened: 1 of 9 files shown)


khandrew1:
Thanks for the PR! However at this time, we aren't accepting new features for the Python SDK, so I'm going to close this one out.

PR #1708, opened 2026-06-11
fix(react): isolate useMcp connection lifecycle (Option A)

### LINKED ISSUE #1696: useMcp: isolate connection lifecycle state instead of guarding stale disconnects
## Background

PR #1528 fixes the dashboard "environment switch breaks the connection" bug (symptoms: `MCP client is not ready (current state: ready)` and `No active session found`). The root cause is that React effect cleanups cannot be awaited, so on a URL/env change the old effect's async `disconnect()` can resolve *after* the new effect's `connect()` has already reused the same `BrowserMCPClient` instance and written a fresh session into the shared, fixed session slot (`sessions["inspector-server"]`).

PR #1528 closes this with **guards** (a compare-and-swap style approach):

- `BaseMCPClient.closeSession()` only deletes `sessions[name]` if the slot still references the captured session.
- `useMcp.disconnect()` only nulls `clientRef` / resets state when not superseded by a newer `connect()` (tracked via a `connectEpochRef` counter + client-identity check).

This is the proportionate fix for the bug and should ship. This issue tracks the **deeper design follow-up**.

## Problem with the current design

The guards are correct but they are defensive patches around a more fundamental issue: **mutable connection state is shared across connection lifecycles.** Specifically:

1. `connect()` reuses the same `BrowserMCPClient` instance across different URLs.
2. The session map uses a single fixed key (`USE_MCP_SERVER_NAME = "inspector-server"`), so a new connection's session occupies the same slot the old one is tearing down.
3. Effect cleanup (`disconnect`) runs unawaited and overlaps the next `connect()`.

Every new code path that touches `clientRef` / `sessions` during teardown has to remember to re-check epoch/identity, or it reintroduces the race. That is a maintenance hazard.

## Proposed directions (pick one, needs design discussion)

**Option A - Fresh client per connection lifecycle.**
`disconnect()` captures the current client, nulls `clientRef` synchronously, then awaits teardown on the captured instance. `connect()` always constructs a new `BrowserMCPClient`. The shared-slot race cannot exist by construction. Cost: loses instance reuse; must verify notification handlers, fetch/proxy interceptors, and auth provider wiring tolerate per-connect re-creation.

**Option B - Serialize connect after pending disconnect.**
`connect()` awaits any in-flight `disconnect()` before starting. Eliminates overlap. Cost: adds latency to env switches and needs a hang-timeout so a stuck transport teardown can't block reconnects indefinitely.

**Option C - Per-connection session keys / lifecycle tokens.**
Key sessions (and clientRef ownership) by the `connectEpoch` so an old lifecycle physically cannot address the new lifecycle's state. Generalizes the current epoch guard into the data model rather than into each call site.

## Acceptance criteria

- A URL/env switch can never let a stale teardown affect the live connection, without per-call-site epoch/identity guards.
- No regression on: StrictMode double-mount, proxy/gateway fallback (which re-creates `clientRef`), OAuth popup/redirect reconnect, and manual `disconnect()` then `connect()`.
- Existing regression tests still pass:
  - `tests/unit/react/useMcp-disconnect-reused-client-race.test.tsx`
  - `tests/unit/client/close-session-slot-guard.test.ts`

## References

- PR #1528 (guard-based fix that this issue follows up on)
- `libraries/typescript/packages/mcp-use/src/react/useMcp.ts` - `disconnect()` / `connect()` / main connection effect
- `libraries/typescript/packages/mcp-use/src/client/base.ts` - `closeSession()`

piyushbag:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Isolate `useMcp` connection lifecycle so environment/URL switches cannot let a stale `disconnect()` interfere with the live MCP session.

Follow-up to #1528 (epoch/slot guards). Implements Option A from #1696: a fresh `BrowserMCPClient` per `connect()`, with `disconnect()` clearing `clientRef` synchronously before awaiting teardown on the captured instance.

## Review Readiness

## Implementation Details

`connect()` always constructs a new `BrowserMCPClient` instead of reusing `clientRef`. `disconnect()` captures the client, nulls `clientRef` synchronously, then awaits `closeSession` on that captured instance only. `connectEpochRef` stays for stale UI reset suppression. The `BaseMCPClient.closeSession` slot guard from #1528 is unchanged.

---

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [x] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

- [x] Ran `pnpm exec eslint` on changed files
- [x] Ran `pnpm exec prettier --write` on changed files
- [x] Ran `pnpm --filter mcp-use build` and build succeeds
- [x] Ran `pnpm changeset` (changeset file included)
- [x] Added or updated tests

## Testing

```bash
cd libraries/typescript
pnpm --filter mcp-use build
pnpm --filter mcp-use exec vitest run \
  tests/unit/react/useMcp-disconnect-reused-client-race.test.tsx \
  tests/unit/client/close-session-slot-guard.test.ts
pnpm --filter mcp-use exec vitest run tests/unit/react/
```

- URL/env switch with deferred `closeSession` on prior client
- Per-connect `BrowserMCPClient` construction
- Disconnect then re-enable reconnect path
- Existing `close-session-slot-guard` tests still pass

### Manual testing

Unit regressions cover the #1528 race (URL switch while prior `closeSession` is in flight). I have not re-run the original dashboard env-switch repro in MCP Inspector yet. I can add Inspector steps and results here if that would help review.

## Backwards Compatibility

Backwards compatible. Public `useMcp` API unchanged. Slightly more client instantiation on reconnect (expected tradeoff per #1696 Option A).

## Related Issues

Fixes #1696

### DIFF
--- a/libraries/typescript/.changeset/cli-skip-update-check-ci.md
+++ b/libraries/typescript/.changeset/cli-skip-update-check-ci.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Skip update-available notifications in CI, non-TTY, or when MCP_USE_SKIP_UPDATE_CHECK=1 so scripted CLI output stays clean.
--- a/libraries/typescript/.changeset/usemcp-lifecycle-isolation.md
+++ b/libraries/typescript/.changeset/usemcp-lifecycle-isolation.md
@@ -1,0 +1,7 @@
+---
+"mcp-use": patch
+---
+
+fix(react): isolate useMcp connection lifecycle with per-connect BrowserMCPClient
+
+Fixes environment/URL switch races where a stale disconnect could interfere with the live MCP session after #1528 guard patches.
--- a/libraries/typescript/packages/cli/src/utils/update-check.ts
+++ b/libraries/typescript/packages/cli/src/utils/update-check.ts
@@ -138,11 +138,22 @@ function resolveInstalledVersion(
 /**
  * Check npm for a newer version of `mcp-use` and print a notification when
  * one is available. Runs silently on any error so it never interrupts the CLI.
+ *
+ * Skipped in CI, when explicitly disabled, or when stdout is not a TTY so
+ * piped and scripted consumers get clean machine-readable output.
  */
 export async function notifyIfUpdateAvailable(
   projectPath: string | undefined
 ): Promise<void> {
   try {
+    if (
+      process.env.CI === "true" ||
+      process.env.MCP_USE_SKIP_UPDATE_CHECK === "1" ||
+      !process.stdout.isTTY
+    ) {
+      return;
+    }
+
     const installed = resolveInstalledVersion(projectPath);
     if (!installed) return;
 
--- a/libraries/typescript/packages/cli/tests/cli-integration.test.ts
+++ b/libraries/typescript/packages/cli/tests/cli-integration.test.ts
@@ -36,6 +36,8 @@ async function runCLI(
         NO_COLOR: "1", // Disable colors for easier testing
         HOME: FAKE_HOME,
         USERPROFILE: FAKE_HOME,
+        // Keep stdout free of update banners in assertions / CI.
+        MCP_USE_SKIP_UPDATE_CHECK: "1",
       },
     });
 
--- a/libraries/typescript/packages/mcp-use/src/react/useMcp.ts
+++ b/libraries/typescript/packages/mcp-use/src/react/useMcp.ts
@@ -363,7 +363,7 @@ export function useMcp(options: UseMcpOptions): UseMcpResult {
   const connectingRef = useRef<boolean>(false);
   const isMountedRef = useRef<boolean>(true);
   const connectAttemptRef = useRef<number>(0);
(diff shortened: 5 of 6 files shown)


piyushbag:
CI looks green on my latest push (label workflow + Socket checks). No outstanding bot threads from my side.

Ready for review when you have bandwidth. Happy to add a minimal repro test or adjust the Option A approach if you prefer a different lifecycle split.

piyushbag:
Rebased onto latest `main` and fixed the `typescript/cli` failure.

Root cause: the update-available notifier was writing to stdout during `servers list` when no servers exist, so the empty-stdout assertion failed.

Fix: skip update notices when `CI=true`, `MCP_USE_SKIP_UPDATE_CHECK=1`, or stdout is not a TTY. Integration tests set the skip env as well.

PR #1729, opened 2026-06-13
fix(typescript): enable JSON responses in stateful StreamableHTTP transport to prevent tool call hanging

### LINKED ISSUE #1724: Stateful StreamableHTTP transport hangs in mcp-use dev — tool result never reaches client (SSE response dropped, works with enableJsonResponse)
**Describe the bug**

Tool calls hang forever in the inspector when running `mcp-use dev`. The weird part is the server actually finishes the tool fine — my logs show the handler completing in ~227ms — but the result never makes it back to the client. The inspector just spins until it hits the request timeout (10 min for me) and then errors with `MCP error -32001: Request timed out`. It seems to only happen the first time I call a tool whose widget hasn't been loaded yet that session; once a widget's been loaded, that tool is fast.

I think I found the cause. In `src/server/index.js`, the stateless transport sets `enableJsonResponse: true` (with a comment that edge runtimes can't maintain long-lived SSE streams), but the two stateful transports don't set it, so they fall back to SSE. The inspector uses a stateful session, so the tool response gets written to a deferred SSE stream and silently dropped instead of reaching the client. Running on plain Node via `@hono/node-server`, not an edge runtime, so the SSE path is breaking there too.

**To Reproduce**

1. Run `mcp-use dev`
2. Open the inspector
3. Call a widget tool you haven't called yet this session
4. It spins forever, then eventually fails with `MCP error -32001: Request timed out`

**Expected behavior**

The tool result comes back and the widget renders, the same way it does on later calls to already-loaded tools.

**Desktop:**
- OS: macOS Tahoe 26.5.1
- Browser: Chrome (incognito) 149.0.7827.103
- Version: mcp-use 1.30.0

**Additional context**

Server side, everything completes — the `tools/call` response just never gets delivered. I also see two `resources/read` for the same widget around the call (one before the `tools/call`, one after), which might matter for the timing. Setting `enableJsonResponse: true` on the stateful transports, like the stateless one already does, would probably fix it.

AasheeshLikePanner:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [X] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Added `enableJsonResponse: true` to both stateful `WebStandardStreamableHTTPServerTransport` constructors in `mount-mcp.ts`. Without this flag, stateful sessions used SSE delivery for tool call results, which silently dropped responses when the SSE stream wasn't fully established — causing clients (like the inspector) to hang forever until timeout. The stateless transport already had this flag.

## Review Readiness

- [X] The PR is scoped to one issue or one clearly related change
- [X] The PR description explains the user-visible problem and the fix
- [X] The diff avoids unrelated formatting, generated files, and bulk rewrites
- [X] I verified the affected package/docs path with the commands listed below
- [X] I checked for existing open PRs that already solve the same issue

## Implementation Details

1. Added `enableJsonResponse: true` to the session recovery transport in `mount-mcp.ts:403`
2. Added `enableJsonResponse: true` to the new session transport in `mount-mcp.ts:520`
3. Both stateful paths now match the stateless path (line 331) which already set this flag
4. No dependencies added or modified

---

## TypeScript Checklist

### Packages Modified

- [ ] `docs`
- [X] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [X] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

- [X] Ran `pnpm build` and build succeeds without errors
- [X] Ran `pnpm changeset` to create a changeset (if this PR includes user-facing changes)
- [X] Added or updated tests if needed

---

## Example Usage (Before)

```typescript
// Tool call hangs — POST returns 202, result sent via SSE and dropped
const transport = new WebStandardStreamableHTTPServerTransport({
  sessionIdGenerator: () => sessionId,
  // missing: enableJsonResponse: true
});
```

## Example Usage (After)

```typescript
// Tool call returns immediately — POST returns 200 with JSON body
const transport = new WebStandardStreamableHTTPServerTransport({
  sessionIdGenerator: () => sessionId,
  enableJsonResponse: true,
});
```

## Documentation Updates

None required — bug fix aligning stateful behavior with existing stateless behavior.

## Testing

- Added `tests/integration/issue-1724-json-response.test.ts` with 3 tests:
  - Sequential tool call via SDK client
  - Concurrent tool calls via SDK client
  - Direct HTTP check confirming POST returns 200 with JSON body (not 202 SSE)
- All 287 CLI tests pass, all 574 mcp-use unit tests pass
- All integration tests pass in isolation

## Backwards Compatibility

Fully backwards compatible. `enableJsonResponse: true` makes stateful transports return JSON-RPC responses directly in POST body instead of deferring to SSE. The SDK's `StreamableHTTPClientTransport` and the inspector both handle JSON POST responses correctly.

## Related Issues

Closes #1724

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/endpoints/mount-mcp.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/endpoints/mount-mcp.ts
@@ -400,6 +400,7 @@ export async function mountMcp(
         const server = mcpServerInstance.getServerForSession(sessionId);
         const transport = new WebStandardStreamableHTTPServerTransport({
           sessionIdGenerator: () => sessionId,
+          enableJsonResponse: true,
 
           onsessioninitialized: async (sid: string) => {
             if (getDebugLevel() !== "info") {
@@ -516,6 +517,7 @@ export async function mountMcp(
       const server = mcpServerInstance.getServerForSession(newSessionId);
       const transport = new WebStandardStreamableHTTPServerTransport({
         sessionIdGenerator: () => newSessionId,
+        enableJsonResponse: true,
 
         onsessioninitialized: async (sid: string) => {
           if (getDebugLevel() !== "info") {
--- a/libraries/typescript/packages/mcp-use/tests/integration/issue-1724-json-response.test.ts
+++ b/libraries/typescript/packages/mcp-use/tests/integration/issue-1724-json-response.test.ts
@@ -1,0 +1,171 @@
+import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import { z } from "zod";
+import { MCPServer } from "../../src/server/index.js";
+import { text } from "../../src/server/utils/response-helpers.js";
+import { Client } from "@modelcontextprotocol/sdk/client/index.js";
+import { StreamableHTTPClientTransport } from "@modelcontextprotocol/sdk/client/streamableHttp.js";
+
+const TEST_PORT = 3095;
+const SERVER_URL = `http://localhost:${TEST_PORT}/mcp`;
+
+describe("Issue #1724 - Stateful transport should return tool results in POST response (not SSE)", () => {
+  let server: MCPServer;
+
+  beforeAll(async () => {
+    server = new MCPServer({
+      name: "issue-1724-test-server",
+      version: "1.0.0",
+    });
+
+    server.tool(
+      {
+        name: "hello",
+        description: "Says hello",
+        schema: z.object({
+          name: z.string(),
+        }),
+      },
+      async ({ name }) => {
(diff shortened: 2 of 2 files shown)


khandrew1:
Closing, we shouldn't enable JSON responses in streamable HTTP requests. See issue for more context.

> I don’t think setting `enableJsonResponse: true` on the stateful transports is the correct fix. This will end up breaking streaming tools in the SDK. JSON responses are useful for simple request/response scenarios without streaming. Default is false; SSE streams are preferred.

PR #1735, opened 2026-06-14
fix(python): close code-mode exec() sandbox bypass + add pluggable executor seam

### LINKED ISSUE #1718: Security: exec() on agent-generated code — sandbox bypass possible
## Summary

`code_executor.py` uses `exec(compiled_wrapped, namespace)` to run agent-generated Python code. The "restricted namespace" does not prevent sandbox escape.

## Impact

Python's `exec()` with restricted globals is bypassable via:
- `__import__('os').system('...')`
- `type.__subclasses__()` chain to access builtins
- `__builtins__` reconstruction

If an attacker controls the agent's output via prompt injection, this becomes RCE on the host machine.

## Recommendation

In-process sandboxing of `exec()` is not possible in Python. Consider:
1. Subprocess isolation: run code in a separate process with restricted permissions
2. Docker/container execution with no network access
3. WASM sandbox (e.g. via Extism)
4. Allowlist of permitted operations instead of exec()

## Context

Found with [mcp-redteam](https://github.com/m0rvayne/mcp-redteam) — MCP server security scanner. Rule MRT004 (eval injection). Static analysis, no active exploitation.

claygeo:
### What this fixes

Closes #1718. `code_executor.py` ran agent-generated Python through `exec()` in a hand-picked `__builtins__` dict it called "secure". It isn't. From inside that namespace:

```
().__class__.__base__.__subclasses__()        # reach arbitrary classes
-> rebuild __import__ from a function's __globals__ -> os.popen / subprocess
```

`getattr` and `type` were even handed to the sandbox. And with `code_mode=True` the agent calls `execute_code` as a tool, so a prompt-injected agent gets host RCE. `import`/`open`/`eval` were blocked, which is what made it look safe, but those were never the escape vector.

Restricting a Python namespace is not a security boundary. That is well known and not patchable in-process.

### Approach

The TypeScript side already has the right design (`libraries/typescript/packages/mcp-use/src/client/executors`: `BaseCodeExecutor` -> vm (default, "not for untrusted code") / e2b (cloud isolation) / custom). Python had none of it. This brings Python toward that shape, scoped to what closes the issue honestly:

- **Stop pretending.** The in-process executor is now `VMCodeExecutor`, documented as not a security boundary, and it warns once per process. Docstrings and docs no longer claim "secure".
- **Defense-in-depth (not a sandbox).** An AST check rejects imports, dunder attribute access, denied/dunder names, and dunder string literals before running. `getattr`/`type`/`hasattr` are withheld from the namespace. This raises the bar against accidental and low-effort injection. It is not a guarantee, and the docs say so.
- **Pluggable seam.** `code_executor=` accepts `"vm"`, a `BaseCodeExecutor` subclass or instance, or a `(code, timeout)` callable, so a real sandbox (e.g. E2B) drops in. `CodeExecutor` stays as an alias of `VMCodeExecutor` for backwards compatibility.
- **Honest docs.** The code-mode docs get a security section, a warning, and a confused-deputy caveat for sandboxed executors.

### Deliberately not in this PR

A full Python `E2BCodeExecutor` (parity with the TS e2b executor). It is the bigger half, needs an account to test, and lands cleaner as its own change once the seam looks right. The seam is built so it drops in. Happy to do that follow-up.

### Tests

- The #1718 escapes (`__subclasses__` walk, `__builtins__` reconstruction, getattr / format-string tricks) are now regression tests asserting they are blocked.
- `getattr`/`type`/`hasattr` unavailable; a dynamically built dunder string is proven inert (no primitive left to use it).
- Pluggable executor selection + backwards-compat alias + `FunctionCodeExecutor` (sync/async).
- Full unit suite green (333 passed). Ruff clean.

```
cd libraries/python
pytest tests/unit/test_code_executor.py tests/unit/test_executors.py
```


claygeo:
reviving this since #1718 is still open and the repro still fires on the latest `main` (rebased onto main, CI green).

while reviving it i ran an adversarial pass on the in-process executor's AST defense and closed two more low-effort escapes that were slipping past it:

1. the namespace injected the **real `asyncio` module**, which transitively re-exposes `os`/`subprocess` through plain non-dunder attribute chains, e.g. `asyncio.base_events.os.popen(...)` and `asyncio.events.sys.modules["builtins"]` -> recovers real `eval`/`__import__`. one-liners, full RCE, all passed the AST check.
2. non-dunder interpreter-introspection attributes (`f_globals`, `f_builtins`, `f_back`, `gi_frame`, `cr_frame`, `tb_frame`, ...) let code walk from a generator/coroutine/traceback frame up to a host frame holding the real builtins.

both are now closed: `asyncio` is exposed as a curated facade (a closed allowlist object, not the live module) so only the coroutine helpers (`sleep`, `gather`, `wait_for`, ...) are reachable, and the AST check now also rejects the introspection attributes above. added regression tests for both classes; `await asyncio.sleep(...)` / `asyncio.gather(...)` and ordinary code-mode code are unchanged. 61 executor unit tests pass, ruff clean.

scope is unchanged from the original PR: this is best-effort defense-in-depth, the in-process `VMCodeExecutor` is explicitly **not** a security boundary, and the real answer for untrusted input is the pluggable executor seam + a sandboxed (E2B) executor, which i'm happy to send as the follow-up PR. would appreciate a re-review whenever someone has a cycle. thanks!

claygeo:
small update: i kept the adversarial pass going and it surfaced two more low-effort escapes past the AST layer, both now closed (force-pushed):

1. `match`/`case` class patterns: `case list(__class__=x)` does an internal `getattr(subject, "__class__")` whose attribute name lives in `MatchClass.kwd_attrs` as a plain string, so it slipped past the attribute/name/string rules. the AST check now rejects dunder/introspection names in class-pattern kwd_attrs.
2. event-loop leak: the curated facade still returns real `Future`/`Lock` objects, and `Future.get_loop()` reaches the running loop, which exposes `loop.subprocess_shell(...)` (a duck-typed protocol built from `Exception()` + lambdas, no class needed). the AST check now rejects the loop-accessor names (`get_loop`/`_get_loop`/`_loop`); `await asyncio.gather(...)` / `async with asyncio.Lock()` still work since the library uses those internally, not via agent code.

also folded in a few correctness fixes from review (guard the warn under warnings-as-errors so `execute()` always returns an `ExecutionResult`; the denied-name rule is now load-only and skips locally-bound names so `type = row["type"]` isn't falsely rejected). regression tests added for each; 75 executor tests pass.

one residual i'm documenting rather than fixing, because it can't be cleanly fixed in-process: `str.format` / f-strings do attribute access at *runtime*, so a dynamically built format string (`"{0.__glo" + "bals__}".format(fn)`) can walk an exposed callable's `__globals__` and stringify module/env data. it's information-disclosure only (format returns a repr, never a live/callable handle, so there's no `os.system` path from it), and closing it would mean banning `str.format`. i added a docstring note; it's another reason the thesis stands: this is best-effort hardening, not a sandbox, and untrusted input should go through a sandboxed `code_executor` (E2B). happy to send the E2B executor as the follow-up PR.

PR #1738, opened 2026-06-15
fix(python): accept read_timeout_seconds in WebSocketConnector.call_tool

### LINKED ISSUE #1731: Python WebSocketConnector.call_tool is incompatible with MCPSession read_timeout_seconds
**Describe the bug**

The Python `MCPSession.call_tool()` path forwards `read_timeout_seconds` to every connector:

```python
# libraries/python/mcp_use/client/session.py
return await self.connector.call_tool(name, arguments, read_timeout_seconds)
```

`BaseConnector.call_tool()` supports this third argument, but `WebSocketConnector.call_tool()` overrides the method with only two user arguments:

```python
# libraries/python/mcp_use/client/connectors/websocket.py
async def call_tool(self, name: str, arguments: dict[str, Any]) -> Any:
    return await self._send_request("tools/call", {"name": name, "arguments": arguments})
```

So using `MCPSession.call_tool()` with a websocket connector can raise a Python `TypeError` before the tool call is sent.

**To Reproduce**

Code-inspection/minimal path:

1. Create an `MCPSession` backed by `WebSocketConnector`.
2. Call `await session.call_tool("some_tool", {})`.
3. `MCPSession.call_tool()` passes `read_timeout_seconds` as the third positional argument.
4. `WebSocketConnector.call_tool()` does not accept it, so the call is incompatible with the session/base connector contract.

Minimal direct reproduction of the signature mismatch:

```python
from datetime import timedelta

connector = WebSocketConnector("ws://localhost:8000/mcp")
await connector.call_tool("tool", {}, timedelta(seconds=5))
# TypeError: WebSocketConnector.call_tool() takes 3 positional arguments but 4 were given
```

**Expected behavior**

`WebSocketConnector.call_tool()` should match `BaseConnector.call_tool()` / `MCPSession.call_tool()` and accept `read_timeout_seconds: timedelta | None = None`. It should either enforce the timeout for websocket requests or explicitly ignore it while preserving the shared connector interface.

 - OS: N/A
 - Browser: N/A
 - Version: main branch / Python library

 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Version: N/A

**Additional context**

This affects the Python websocket transport and any higher-level client/session code that uses the common MCP session interface.

truffle-dev:
Closes #1731

`MCPSession.call_tool` forwards `read_timeout_seconds` as a third positional argument to the connector (`session.py`: `self.connector.call_tool(name, arguments, read_timeout_seconds)`), matching `BaseConnector.call_tool`. `WebSocketConnector.call_tool` only declared `(self, name, arguments)`, so every tool call routed through a websocket session raised:

```
TypeError: WebSocketConnector.call_tool() takes 3 positional arguments but 4 were given
```

This aligns the signature with the shared connector interface and threads the timeout through `_send_request` via `asyncio.wait_for`, so it is honored when supplied and ignored (current behavior) when `None`. On timeout the existing `except` branch drops the request from `pending_requests`.

Verified by stashing the source fix and running the new test against the unpatched connector, which reproduces the exact `TypeError` from the issue; with the fix applied all three new cases pass.

New unit coverage in `tests/unit/test_websocket_connector.py`:
- `call_tool` accepts `read_timeout_seconds` positionally the way `MCPSession` passes it
- default (no timeout) path resolves and returns the result
- a supplied timeout is enforced and the pending request is cleaned up

`ruff check .`, `ruff format --check .`, and `pytest tests/unit` pass locally (the one unrelated pre-existing failure in `test_config.py::test_create_sandboxed_stdio_connector` patches the deprecated `mcp_use.connectors.sandbox` path and is independent of this change).

### DIFF
--- a/libraries/python/mcp_use/client/connectors/websocket.py
+++ b/libraries/python/mcp_use/client/connectors/websocket.py
@@ -8,6 +8,7 @@ through WebSocket connections.
 import asyncio
 import json
 import uuid
+from datetime import timedelta
 from typing import Any
 
 import httpx
@@ -178,7 +179,12 @@ class WebSocketConnector(BaseConnector):
         if errors:
             logger.warning(f"Encountered {len(errors)} errors during resource cleanup")
 
-    async def _send_request(self, method: str, params: dict[str, Any] | None = None) -> Any:
+    async def _send_request(
+        self,
+        method: str,
+        params: dict[str, Any] | None = None,
+        read_timeout_seconds: timedelta | None = None,
+    ) -> Any:
         """Send a request and wait for a response."""
         if not self.ws:
             raise RuntimeError("WebSocket is not connected")
@@ -197,6 +203,8 @@ class WebSocketConnector(BaseConnector):
 
         # Wait for the response
         try:
+            if read_timeout_seconds is not None:
+                return await asyncio.wait_for(future, timeout=read_timeout_seconds.total_seconds())
             return await future
         except Exception as e:
             # Remove the request from pending requests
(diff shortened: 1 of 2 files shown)


truffle-dev:
Friendly check-in on this one. The two red `python-*-tests-report` checks are a fork-PR artifact, not the change: the tests themselves pass (`✅ All primitive tests passed`), and the job only errors at the final step where `github-script` tries to POST the results comment and hits `403 Resource not accessible by integration` (the read-only `GITHUB_TOKEN` on fork PRs can't comment). The fix is a one-line signature alignment closing #1731, plus a stash-bisected regression test. Happy to rebase if it helps it land.

PR #1741, opened 2026-06-16
Handle inspector CLI version and unknown options

cuixixixi-xi:
## Summary
- add early `--version` / `-v` handling for `mcp-inspect`
- fail fast for unknown top-level CLI options instead of starting the inspector server
- report missing values for `--url` and `--port`
- regenerate the inspector version during package build so the CLI reports the current package version

## Validation
- Rebases cleanly on current `canary`
- Ran `node scripts/generate-version.mjs` previously in `libraries/typescript/packages/inspector` and verified `src/server/version.ts` is updated to `10.0.0`
- Ran `git diff --check` previously
- Reviewed the CLI arg parsing path to confirm `--version`, `-v`, and unknown options exit before server startup/browser opening

Related: charles-openclaw/charles-microbounties#1320

/claim #1320


khandrew1:
Duplicate of #1722

PR #1750, opened 2026-06-16
fix(cli): allow empty env var updates

### LINKED ISSUE #1749: CLI env update cannot set a variable value to an empty string
**Describe the bug**

`mcp-use servers env update <KEY> --server <id> --value ""` is rejected as if no update fields were provided. This prevents users from intentionally clearing an environment variable value while keeping the variable itself.

The issue is in `updateEnvCommand`: the no-op guard uses `!options.value`, so Commander's empty string value is treated the same as an omitted `--value`. Later code correctly checks `options.value !== undefined`, but it is never reached for empty strings.

**To Reproduce**

1. Run `mcp-use servers env update API_KEY --server <server-id> --value ""`.
2. The CLI prints `Nothing to update. Provide at least one of --value, --env, --sensitive.`
3. The API update call is not made.

```ts
it("allows updating an env var value to an empty string", async () => {
  const { createEnvCommand } = await import("../src/commands/env.js");
  mockApiInstance.listEnvVariables.mockResolvedValue([
    {
      id: VAR_ID,
      key: "API_KEY",
      value: "old",
      environments: ["production"],
      branch: null,
      sensitive: false,
    },
  ]);
  mockApiInstance.updateEnvVariable.mockResolvedValue({
    id: VAR_ID,
    key: "API_KEY",
    value: "",
    environments: ["production"],
    branch: null,
    sensitive: false,
  });

  const cmd = createEnvCommand();
  await cmd.parseAsync(
    ["update", "API_KEY", "--server", SERVER_ID, "--value", ""],
    { from: "user" }
  );

  expect(mockApiInstance.updateEnvVariable).toHaveBeenCalledWith(
    SERVER_ID,
    VAR_ID,
    expect.objectContaining({ value: "" })
  );
});
```

**Expected behavior**

Passing `--value ""` should be accepted and should call `updateEnvVariable(..., { value: "" })`. Only `options.value === undefined` should mean the flag was omitted.

I verified this locally by adding the focused test above to `libraries/typescript/packages/cli/tests/servers-update-env-branch.test.ts` and running the CLI package tests. The relevant failure is:

```text
✗ Nothing to update. Provide at least one of --value, --env, --sensitive.
AssertionError: expected updateEnvVariable to be called ...
Number of calls: 0
```

The same test should pass if the no-op guard checks `options.value === undefined` instead of using truthiness.

luochen211:
## Summary
- allow `servers env update --value ""` to update an env var to an empty string
- keep the no-op guard for truly omitted update fields
- add a regression test for empty-string env updates

## Tests
- pnpm --filter @mcp-use/cli exec vitest run tests/servers-update-env-branch.test.ts
- pnpm exec prettier --check packages/cli/src/commands/env.ts packages/cli/tests/servers-update-env-branch.test.ts
- git diff --check

Closes #1749

### DIFF
--- a/.github/workflows/bump-mcp-use-reusable.yml
+++ b/.github/workflows/bump-mcp-use-reusable.yml
@@ -61,7 +61,7 @@ jobs:
 
           if [ "${{ steps.pm.outputs.pm }}" = "pnpm" ]; then
             corepack enable
-            pnpm install --no-frozen-lockfile --config.dangerously-allow-all-builds=true
+            pnpm install --no-frozen-lockfile
             pnpm run build
           else
             npm install
--- a/libraries/typescript/.changeset/empty-env-value-update.md
+++ b/libraries/typescript/.changeset/empty-env-value-update.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Allow updating CLI environment variable values to an empty string.
--- a/libraries/typescript/packages/cli/src/commands/env.ts
+++ b/libraries/typescript/packages/cli/src/commands/env.ts
@@ -215,7 +215,11 @@ async function updateEnvCommand(
   try {
     await requireLogin();
 
-    if (!options.value && !options.env && options.sensitive === undefined) {
+    if (
+      options.value === undefined &&
+      !options.env &&
+      options.sensitive === undefined
+    ) {
       console.error(
         chalk.red(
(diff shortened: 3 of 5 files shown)


tonxxd:
Thanks @luochen211 for the PR! could you please also add the changeset file?

luochen211:
Added the changeset for @mcp-use/cli as a patch release. Thanks for the nudge!

PR #1752, opened 2026-06-16
fix(cli): preserve deployment list sort order

### LINKED ISSUE #1751: CLI deployments list --sort output is reordered by createdAt
**Describe the bug**

`mcp-use deployments list --sort <field>` sends the requested sort to the API, but then the CLI re-sorts the returned page locally by `createdAt` descending before printing. This can make the visible output ignore the user's requested sort order.

For example, if the API returns deployments in `name:asc` order, the CLI can still display a newer `beta` deployment before an older `alpha` deployment.

**To Reproduce**

1. Run `mcp-use deployments list --sort name:asc` against deployments whose names and creation times sort differently.
2. The API receives `sort: "name:asc"`.
3. The CLI prints rows in `createdAt` descending order instead of preserving the API's sorted page.

```ts
it("preserves the API order when --sort is provided", async () => {
  const { isLoggedIn } = await import("../src/utils/config.js");
  vi.mocked(isLoggedIn).mockResolvedValue(true);
  mockApiInstance.testAuth.mockResolvedValue({ orgs: [] });
  mockApiInstance.getServer.mockResolvedValue({ organizationId: "org_1" });
  mockApiInstance.listDeployments.mockResolvedValue({
    items: [
      { ...mockDeployment, id: "dep_alpha", name: "alpha", createdAt: "2024-01-01T00:00:00Z", serverId: "srv_1" },
      { ...mockDeployment, id: "dep_beta", name: "beta", createdAt: "2024-02-01T00:00:00Z", serverId: "srv_1" },
    ],
    total: 2,
    limit: 30,
    skip: 0,
  });
  const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});

  const { createDeploymentsCommand } = await import("../src/commands/deployments.js");
  const cmd = createDeploymentsCommand();
  await cmd.parseAsync(["list", "--sort", "name:asc"], { from: "user" });

  const output = logSpy.mock.calls.map((args) => args.join(" ")).join("\n");
  expect(output.indexOf("alpha")).toBeLessThan(output.indexOf("beta"));
  logSpy.mockRestore();
});
```

**Expected behavior**

The CLI should preserve the API's returned page order, especially when `--sort` is provided. If the API is responsible for sorting/pagination, the client should not re-sort the page afterward.

I verified this locally by adding the focused test above to `libraries/typescript/packages/cli/tests/deployments.test.ts` and running:

```bash
pnpm --filter @mcp-use/cli exec vitest run tests/deployments.test.ts
```

The relevant failure was:

```text
AssertionError: expected 495 to be less than 336
```

That means `beta` appeared in output before `alpha`, even though the mocked API returned `alpha` first.

luochen211:
## Summary
- preserve the deployment page order returned by the API
- avoid re-sorting `deployments list --sort ...` output by `createdAt`
- add a regression test for custom sort output order

## Tests
- pnpm --filter @mcp-use/cli exec vitest run tests/deployments.test.ts
- pnpm exec prettier --check packages/cli/src/commands/deployments.ts packages/cli/tests/deployments.test.ts
- git diff --check

Closes #1751

### DIFF
--- a/libraries/typescript/.changeset/cli-deployments-list-sort-order.md
+++ b/libraries/typescript/.changeset/cli-deployments-list-sort-order.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+`mcp-use deployments list` now preserves the deployment ordering returned by the API instead of re-sorting by creation date on the client. This keeps the displayed order consistent with the server's pagination and sort.
--- a/libraries/typescript/packages/cli/src/commands/deployments.ts
+++ b/libraries/typescript/packages/cli/src/commands/deployments.ts
@@ -133,12 +133,7 @@ async function listDeploymentsCommand(options: {
       }
     }
 
-    const sortedDeployments = [...deployments].sort(
-      (a, b) =>
-        new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
-    );
-
-    if (sortedDeployments.length === 0) {
+    if (deployments.length === 0) {
       if (page.total === 0) {
         console.log(chalk.yellow("No deployments found."));
         console.log(
@@ -160,7 +155,7 @@ async function listDeploymentsCommand(options: {
       chalk.cyan.bold(
         `\n📦 ${formatPageHeader(
           "Deployments",
-          sortedDeployments.length,
+          deployments.length,
           page.total
         )}\n`
       )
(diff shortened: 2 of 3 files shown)


khandrew1 (changes requested):
Hey @luochen211! Thanks for the PR. Could you also add a changeset?

khandrew1 (approved):

PR #1759, opened 2026-06-17
Feat/topbar buttons styling

### LINKED ISSUE #1758: redesign chat UI with persistent sessions and sidebar
## Description

This PR introduces persistent multi-session chat support to MCP Inspector, along with a dedicated session history sidebar and several UI refinements.

Previously, MCP Inspector supported only a single ephemeral chat session that was lost on page refresh. Sessions are now persisted locally and managed independently.

---

### 🗂️ Persistent Multi-Session Chat

Introduced `ChatSessionsContext`, backed by IndexedDB (`idb-keyval`), to persist chat sessions across page refreshes and browser restarts.

Storage is split into two stores:

- **Sessions store** — session metadata (id, title, timestamps, LLM configuration)
- **Messages store** — messages stored separately and loaded on demand per session

This keeps session list rendering lightweight while avoiding unnecessary reads on mount.

### 📋 Session History Sidebar

Added a dedicated session history sidebar with:

- Session creation and management
- Active session highlighting
- Session switching
- Per-session deletion
- Clear-all with confirmation
- Inline session naming before creation
- Storage usage indicator with quota warnings

### 🧹 Layout Improvements

- Removed `SidebarRight`
- Consolidated duplicated controls from `LayoutHeader` and `ChatTab`
- Hide sidebar when no MCP server is connected
- Clear stale messages immediately on session switch

---

## Use Case

MCP Inspector is frequently used to test and compare MCP server behavior across different models, prompts, and configurations.

This change enables workflows such as:

1. Running a debugging session against one model
2. Creating a second session with a different configuration
3. Switching between sessions for comparison
4. Returning later and continuing without losing history

---

## Benefits

| Before | After |
|---|---|
| Single ephemeral chat | Persistent multi-session chat |
| Chat lost on refresh | Sessions survive refresh and browser restart |
| No session management | Create, switch, rename, delete, and clear sessions |
| No storage visibility | IndexedDB usage indicator with warnings |
| Duplicate controls across UI | Consolidated controls |
| Sidebar always visible | Sidebar shown only when a server is connected |
| Stale messages on session switch | Content clears immediately on switch |

---

## Implementation Notes

> [!NOTE]
> Chat data is stored **locally in IndexedDB only**. The backend remains stateless and does not persist chat history.

- Session metadata and message content are stored in **separate IndexedDB stores** to improve startup performance and avoid unnecessary reads on mount.

Arpitm544:
close #1758


Arpitm544:
go through #1760

PR #1760, opened 2026-06-17
Feat/inspector multi session chat

### LINKED ISSUE #1758: redesign chat UI with persistent sessions and sidebar

Arpitm544:
close #1758


khandrew1 (changes requested):
image

Thanks for the PR! Here's some comments on the design:
- Red: We now have two separate ways to start a new chat. I say we stick with our original New Chat button on the top right and we change the sidebar's to say something like "Sessions"
- Blue: Don't think this is super useful information. These quick little chats shouldn't even take up any meaningful space.
- Green: These should be clearer to understand WHAT session. I'd switch to provider logos, then make an LLM call with a really specific prompt that generates a title for the session.

Also have you tested what happens if I try to resume a chat after switching to a different MCP Server?

Arpitm544:
- Red: Yeah you're right, we should use sessions instead of session.

- Blue: My thought is that since we're storing data in the browser's IndexedDB, we should show the user how full their IndexedDB is  based on that, they can remove data if needed. but as you say we can remove this 

- Green: I initially implemented this too, but then I thought let the user name it themselves. But since you mentioned it, I'll go ahead and implement that feature as well. 

The main point is server switching so currently i implemented global chat so all chats appear across mcps but the issue is like let say chat a belong to mcp a but when i switch from mcp a to mcp b then chat a also showing up and user can use that chat also which i think is the main drawback. i also thought about auto switching but that become more complex for us and user also. so i think we should change global chat to mcp-wise chat.
@khandrew1 what do you think?

Arpitm544:
https://github.com/user-attachments/assets/bbf7d134-08fe-4314-bec1-2856ec1f949c

khandrew1:
Hey @Arpitm544! Thanks for the suggestion. We implemented this history chat in the v2 inspector. I'd encourage you to go check it out and leave any feedback there!

PR #1769, opened 2026-06-20
fix(cli): preserve deployments list sort order

### LINKED ISSUE #1751: CLI deployments list --sort output is reordered by createdAt

kunaldhongade:
## Summary
- Preserve API-returned deployment order when `mcp-use deployments list --sort <field>` is used.
- Keep the existing no-`--sort` createdAt descending display behavior.
- Add focused regression coverage and a patch changeset for `@mcp-use/cli`.

Closes #1751

## Root Cause
The CLI sent the requested sort to the deployments API, but then always re-sorted the returned page locally by `createdAt` descending before printing. That could make paginated output ignore the explicit sort order.

## Validation
- `pnpm --filter @mcp-use/cli exec vitest run tests/deployments.test.ts`
- `pnpm build`
- `pnpm format:check`
- `pnpm lint`

## Notes
- `pnpm test` was attempted but blocked by unrelated `mcp-use` agent integration tests requiring `OPENAI_API_KEY`.

### DIFF
--- a/libraries/typescript/.changeset/cli-deployments-sort-order.md
+++ b/libraries/typescript/.changeset/cli-deployments-sort-order.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Preserve API ordering for deployments list when an explicit sort is provided.
--- a/libraries/typescript/packages/cli/src/commands/deployments.ts
+++ b/libraries/typescript/packages/cli/src/commands/deployments.ts
@@ -133,12 +133,14 @@ async function listDeploymentsCommand(options: {
       }
     }
 
-    const sortedDeployments = [...deployments].sort(
-      (a, b) =>
-        new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
-    );
+    const displayDeployments = options.sort
+      ? deployments
+      : [...deployments].sort(
+          (a, b) =>
+            new Date(b.createdAt).getTime() - new Date(a.createdAt).getTime()
+        );
 
-    if (sortedDeployments.length === 0) {
+    if (displayDeployments.length === 0) {
       if (page.total === 0) {
         console.log(chalk.yellow("No deployments found."));
         console.log(
@@ -160,7 +162,7 @@ async function listDeploymentsCommand(options: {
       chalk.cyan.bold(
         `\n📦 ${formatPageHeader(
           "Deployments",
-          sortedDeployments.length,
+          displayDeployments.length,
           page.total
         )}\n`
       )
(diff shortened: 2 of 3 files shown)


khandrew1:
Hi, this fix is already in Canary done by #1752.

PR #1770, opened 2026-06-20
fix(inspector): scope OAuth proxy fetch per server configuration

### LINKED ISSUE #1766: Separate Inspector OAuth proxy fetch from global fetch
## Summary

The Inspector's OAuth proxy currently takes over the global `fetch`. In the server list, if only one server is set to `Via Proxy` instead of `Direct`, every fetch goes through the proxy.

This should be separated so the Inspector can reliably test multiple servers.

## Acceptance Criteria

* In a listing of multiple servers, setting one server to use the OAuth proxy does not cause the other servers to use the proxy.
* Each server can independently choose whether requests are made `Via Proxy` or `Direct`.
* Proxy-specific fetch behavior is scoped to the server/configuration that selected it, rather than replacing or mutating global `fetch` behavior.

hemant1026:
## Changes

The Inspector's OAuth proxy took over the global `fetch`. In a list of multiple servers, setting **one** server to "Via Proxy" caused **every** fetch on the page to go through that proxy- breaking the ability to reliably test multiple servers side by side.

`BrowserOAuthClientProvider.installFetchInterceptor()` did `window.fetch = …` (a process-wide mutation). This change replaces that global patch with a per-provider scoped fetch so proxy behavior is confined to the server/configuration that selected it.

## Implementation Details

1. **`auth/browser-provider.ts`** - Replaced `installFetchInterceptor()` / `restoreFetch()` (and the `originalFetch` field) with a scoped `getProxyFetch(baseFetch?)`. It returns a `fetch` that routes only OAuth requests (`.well-known` metadata, token, register, authorize) through the configured OAuth proxy and passes everything else to `baseFetch`. It never touches the global `fetch`. Added a `proxyOAuthRequests` option (default `true`) so a provider can opt out (e.g. when an MCP gateway already proxies OAuth).
2. **`react/useMcp.ts`** - Passes the scoped fetch into the SDK transport (`fetch`) and into all `auth()` calls (`fetchFn`). The MCP SDK threads its transport `fetch` into `auth({ fetchFn })`, which propagates it through metadata discovery, token exchange, and dynamic client registration - so OAuth is fully covered without any global patch. Replaced the `installFetchInterceptor` flags with `proxyOAuthRequests` and removed the global `restoreFetch()` unmount teardown.
3. **`react/useMcp-helpers.ts`** - `createBrowserOAuthProvider` now configures `proxyOAuthRequests` on the provider instead of installing a global interceptor.
4. **`auth/callback.ts`** - The token-exchange `auth()` call now receives a scoped `fetchFn` instead of patching global fetch.

Because the Inspector wraps each server in its own `useMcp` instance (one `BrowserOAuthClientProvider` per server) via `McpClientProvider`, each server independently chooses Via Proxy or Direct with no cross-contamination.

## Example Usage

```ts
// Server A - "Via Proxy": only A's transport/auth get the proxy-routing fetch
const fetchA = providerA.getProxyFetch(); // OAuth → proxy, everything else → base

// Server B - "Direct": no OAuth proxy configured
const fetchB = providerB.getProxyFetch(); // undefined → SDK uses normal fetch

// globalThis.fetch is never reassigned by either.
```

## Testing

- Added `tests/unit/auth/browser-provider-proxy-fetch.test.ts`: Server A ("Via Proxy") vs Server B ("Direct") - asserts B never proxies, that building A's proxy fetch never reassigns the global `fetch`, and that OAuth metadata requests are rewritten to the proxy.
- Rewrote `tests/unit/react/useMcp-proxyCleanup.test.tsx` to assert the global `fetch` identity is preserved across a proxy connection's mount/unmount (replacing the now-obsolete `restoreFetch` assertion).
- `pnpm --filter mcp-use test:unit` → **605 passed / 6 skipped**; `tsc --noEmit`, `pnpm build`, `eslint`, and `prettier --check` all clean.

## Backwards Compatibility

Behavior-preserving for end users: providers proxy OAuth requests under the same conditions as before, just scoped to their own connection. This also structurally removes the related "stale interceptor after switching Via Proxy → Direct" failure mode.

The public `installFetchInterceptor()` / `restoreFetch()` methods on `BrowserOAuthClientProvider` are **removed** (replaced by `getProxyFetch()`). Per the repo guidance to avoid unrequested backward-compat shims, no deprecation wrappers were added — happy to add them if preferred.

## Related Issues

Closes #1766

### DIFF
--- a/libraries/typescript/.changeset/scoped-oauth-proxy-fetch.md
+++ b/libraries/typescript/.changeset/scoped-oauth-proxy-fetch.md
@@ -1,0 +1,18 @@
+---
+"mcp-use": patch
+---
+
+fix(inspector): scope OAuth proxy fetch per server configuration
+
+The browser OAuth provider previously installed a global `window.fetch`
+interceptor to route OAuth requests through the inspector proxy. With multiple
+servers, connecting one server "Via Proxy" mutated `fetch` for the entire page,
+so other servers (including ones set to "Direct") and unrelated requests were
+affected, and switching a server from "Via Proxy" back to "Direct" could leave a
+stale interceptor behind.
+
+`BrowserOAuthClientProvider` now exposes a scoped `getProxyFetch()` that returns
+a `fetch` confined to a single provider. It is passed only to that server's SDK
+transport and `auth()` calls (via the SDK's `fetch` / `fetchFn` options), so
+OAuth-proxy behavior is scoped to the selected server's connection and the
+global `fetch` is never mutated.
--- a/libraries/typescript/packages/mcp-use/src/auth/browser-provider.ts
+++ b/libraries/typescript/packages/mcp-use/src/auth/browser-provider.ts
@@ -34,6 +34,15 @@ interface BrowserOAuthOptions {
   /** MCP proxy URL that client connected to (for resource field rewriting) */
   connectionUrl?: string;
   /**
+   * When true (default), OAuth requests (.well-known metadata, token,
+   * register, authorize) are routed through `oauthProxyUrl` to bypass CORS.
+   * The routing is applied only to the scoped fetch returned by
+   * {@link BrowserOAuthClientProvider.getProxyFetch}; it never mutates the
+   * global `fetch`. Set to false to connect directly even when an OAuth proxy
+   * URL is available (e.g. when the MCP gateway already proxies OAuth).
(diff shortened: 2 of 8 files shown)


khandrew1 (approved):
Thanks for the PR!

PR #1772, opened 2026-06-21
fix(templates): score 100% on publishing checklist (MCP-2260)

tonxxd:
## Summary
Makes the `create-mcp-use-app` `starter` and `mcp-apps` templates score 100% on the publishing checklist (MCP-2260). Verified empirically by running the audit pipeline against locally-running scaffolds (baseline `starter` 97 / `mcp-apps` 92 → both 100).

This is the SDK + templates half of the work. The companion audit-pipeline fixes live in a separate `mcp-use-cloud` PR against `main`.

## Changes
- **`mcp-use` (Apps SDK adapter)**: auto-derive `openai/widgetDescription` from the widget's `description` when it isn't set explicitly. Previously a widget with a perfectly good `description` surfaced no spec widget description (only the non-spec `openai/description`), so hosts and the publishing checklist saw nothing.
- **`create-mcp-use-app` / `starter`**: `fetch-weather` now declares a `title` and an `outputSchema`, returning matching `structuredContent`.
- **`create-mcp-use-app` / `mcp-apps`**: `search-tools` and `get-fruit-details` declare a `title`; the `product-search-result` widget declares a `domain` (its widget description is now auto-derived from `description`).

## Testing
- New unit test: `apps-sdk-adapter.test.ts` covers the `openai/widgetDescription` default (from top-level + unified `metadata.description`, explicit override, and absence). 4 passing.
- Ran the full audit pipeline against scaffolded `starter` and `mcp-apps` (with this SDK build linked) → both 100/100, E2E checks skipped (excluded from score).

## Backwards compatibility
Non-breaking. The widgetDescription default only fills in a value when none was set; explicit `widgetDescription` still wins.

## Related
- MCP-2260
- Companion cloud PR (audit pipeline reads spec `_meta` instead of `annotations`).

### DIFF
--- a/libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/resources/product-search-result/widget.tsx
+++ b/libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/resources/product-search-result/widget.tsx
@@ -30,6 +30,9 @@ export const widgetMetadata: WidgetMetadata = {
     prefersBorder: false,
     invoking: "Loading product search results...",
     invoked: "Product search results loaded",
+    // widgetDescription is auto-derived from `description` above; set it here only to override.
+    // Dedicated origin associated with the hosted widget (required for app submission).
+    domain: "https://mcp-use.com",
     csp: {
       // Widget-specific
       resourceDomains: ["https://cdn.openai.com"],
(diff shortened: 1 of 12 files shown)


libraries/typescript/packages/create-mcp-use-app/src/templates/mcp-apps/resources/product-search-result/widget.tsx
```diff
@@ -30,6 +30,8 @@ export const widgetMetadata: WidgetMetadata = {
     prefersBorder: false,
     invoking: "Loading product search results...",
     invoked: "Product search results loaded",
+    // Dedicated origin associated with the hosted widget (required for app submission).
+    domain: "https://mcp-use.com",
```
khandrew1: this is going to cause it to break on `claude.ai` due to their weird domain settings. the domain HAS to be set to the MCP link or it won't render. if domain is blank, Claude handles the conversion properly

khandrew1 (changes requested):
`domain` will cause the example to not be production ready on Claude

tonxxd:
Fixed @khandrew1, there is in any case a quite delicate typing change to correctly type outpuschema object. requires proper e2e + full type testing

PR #1786, opened 2026-06-24
fix(python): allow empty lists for AuthMiddleware path configuration

### LINKED ISSUE #1753: Python AuthMiddleware cannot accept empty custom path lists
**Describe the bug**

`AuthMiddleware` uses truthy fallback expressions for custom path lists:

```python
self.exclude_paths = exclude_paths or ["/health", "/docs", "/inspector", "/openmcp.json"]
self.protected_paths = protected_paths or ["/mcp"]
```

As a result, callers cannot intentionally pass an empty list. `protected_paths=[]` still protects `/mcp`, and `exclude_paths=[]` still keeps the default public exclusions.

This makes it impossible to express configurations such as:

- no protected paths at the middleware layer
- no default excluded paths because the app wants to explicitly protect `/docs` or `/health`

**To Reproduce**

```python
def test_empty_protected_paths_disable_default_mcp_protection():
    async def endpoint(request: Request) -> JSONResponse:
        return JSONResponse({"ok": True})

    app = Starlette(routes=[Route("/mcp", endpoint, methods=["GET"])])
    app.add_middleware(
        AuthMiddleware,
        auth_provider=MockAuthProvider(),
        protected_paths=[],
    )
    client = TestClient(app, raise_server_exceptions=False)

    response = client.get("/mcp")
    assert response.status_code == 200


def test_empty_exclude_paths_disable_default_exclusions():
    async def endpoint(request: Request) -> JSONResponse:
        return JSONResponse({"ok": True})

    app = Starlette(routes=[Route("/docs", endpoint, methods=["GET"])])
    app.add_middleware(
        AuthMiddleware,
        auth_provider=MockAuthProvider(),
        exclude_paths=[],
        protected_paths=["/docs"],
    )
    client = TestClient(app, raise_server_exceptions=False)

    response = client.get("/docs")
    assert response.status_code == 401
```

**Expected behavior**

`None` should mean "use defaults". An explicit empty list should be respected as an empty list.

I verified this locally by adding the focused tests above to `libraries/python/tests/unit/server/auth/test_auth_middleware.py` and running:

```bash
cd libraries/python
MCP_USE_ANONYMIZED_TELEMETRY=false .venv/bin/python -m pytest tests/unit/server/auth/test_auth_middleware.py -q
```

The relevant failures were:

```text
assert 401 == 200  # protected_paths=[] still protected /mcp
assert 200 == 401  # exclude_paths=[] still excluded /docs
```

luisangelrod:
## What

Changes `AuthMiddleware.__init__` to use explicit `None` checks instead of `or` when setting `exclude_paths` and `protected_paths`, so callers can intentionally pass an empty list to disable the defaults.

## Why

The previous `or` pattern treated empty lists as falsy:

```python
# Before — [] falls through to the default
self.exclude_paths = exclude_paths or ["/health", "/docs", "/inspector", "/openmcp.json"]
self.protected_paths = protected_paths or ["/mcp"]
```

This made it impossible to express configurations like:

- `protected_paths=[]` — disable all middleware-level path protection (e.g. when the auth check is done elsewhere)
- `exclude_paths=[]` — remove the default public exclusions when the app wants to control them explicitly

## Changes

- `middleware.py`: replace `x or default` with `x if x is not None else default` for both path list parameters
- `test_auth_middleware.py`: add regression coverage confirming that `protected_paths=[]` and `exclude_paths=[]` are respected

## Validation

```bash
cd libraries/python
MCP_USE_ANONYMIZED_TELEMETRY=false .venv/bin/python -m pytest tests/unit/server/auth/test_auth_middleware.py -q
uvx --from 'ruff==0.11.12' ruff check mcp_use/server/auth/middleware.py tests/unit/server/auth/test_auth_middleware.py
uvx --from 'ruff==0.11.12' ruff format --check mcp_use/server/auth/middleware.py tests/unit/server/auth/test_auth_middleware.py
.venv/bin/ty check mcp_use/server/auth/middleware.py
```

Results:

- 17 auth middleware tests passed
- Ruff lint and formatting checks passed
- `ty` type checking passed

Closes #1753


khandrew1:

PR #1776, opened 2026-06-21
test: move connector telemetry mocks to module scope

royalpinto007:
Fixes #1777

## Summary

- moves the `posthog-node`, `node:fs`, and `node:os` mocks in `connector-telemetry.test.ts` to module scope
- uses `vi.hoisted` for shared mock functions referenced by hoisted mock factories
- keeps the existing test behavior while matching Vitest's actual mock execution order

## Verification

- `pnpm --filter mcp-use exec vitest run tests/unit/telemetry/connector-telemetry.test.ts`
- `pnpm --filter mcp-use test:unit -- tests/unit/telemetry/connector-telemetry.test.ts`
- `pnpm --filter mcp-use exec eslint tests/unit/telemetry/connector-telemetry.test.ts`
- `pnpm --filter mcp-use build`

Note: the broader focused test command still shows the repo's existing PostHog browser telemetry warnings and Vitest close timeout, but the three nested `vi.mock(...)` warnings are removed.

PR #1787, opened 2026-06-24
fix(typescript): move vi.mock calls to module scope in connector telemetry tests

### LINKED ISSUE #1777: Move nested Vitest mocks to module scope in connector telemetry tests
The connector telemetry test currently defines a few `vi.mock(...)` calls inside `beforeEach`:

- `posthog-node`
- `node:fs`
- `node:os`

Vitest hoists these mocks anyway, so it warns that they should be moved to the top level of the module. The warning also says this will become an error in a future Vitest version.

A small cleanup would make the test match Vitest's actual mock execution order and remove the warning.

luisangelrod:
## What

Moves three `vi.mock()` calls (`posthog-node`, `node:fs`, `node:os`) out of the `beforeEach` block in `connector-telemetry.test.ts` and places them at module scope alongside the existing `telemetry-node` mock.

## Why

Vitest hoists all `vi.mock()` calls to the top of the module regardless of where they appear in source — including calls inside `beforeEach`. Because of this, nesting them in `beforeEach` currently generates a warning:

```
vi.mock() should be called at the top level of the module
```

The warning notes this will become a hard error in a future Vitest version, so the fix aligns the code with Vitest's actual execution model before that happens.

## Changes

- Extracted `vi.mock(posthog-node)`, `vi.mock(node:fs)`, and `vi.mock(node:os)` to module scope
- `beforeEach` in the integration-tests block now only calls `vi.clearAllMocks()` to reset state between tests
- Added a comment explaining why the mocks live at module scope

Closes #1777

### DIFF
--- a/libraries/python/mcp_use/server/auth/middleware.py
+++ b/libraries/python/mcp_use/server/auth/middleware.py
@@ -46,13 +46,12 @@ class AuthMiddleware(BaseHTTPMiddleware):
         """
         super().__init__(app)
         self.auth_provider = auth_provider
-        self.exclude_paths = exclude_paths or [
-            "/health",
-            "/docs",
-            "/inspector",
-            "/openmcp.json",
-        ]
-        self.protected_paths = protected_paths or ["/mcp"]
+        self.exclude_paths = (
+            exclude_paths
+            if exclude_paths is not None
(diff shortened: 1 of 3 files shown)


khandrew1:
Closed by #1776

PR #1788, opened 2026-06-24
fix(typescript): add cwd support to StdioConnector

### LINKED ISSUE #1784: StdIO Transport does not support the `cwd` field
**Describe the bug**
In https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/connectors/stdio.ts#L73 we create a `StdioServerParameters` instance from the `@modelcontextprotocol/sdk`

This instance supports a `cwd` field to set the working directory of the spawned process (https://github.com/modelcontextprotocol/typescript-sdk/blob/c59dc3aa1a633d27fbbe873f1a430483cf7440f8/packages/client/src/client/stdio.ts#L40)

This field however is not populated by `mcp-use`.

Please add support for that (optional) field, so processes can be started correctly.

**To Reproduce**

**Expected behavior**

**Screenshots**

**Additional context**

luisangelrod:
## What

Adds an optional `cwd` parameter to `StdioConnector` and threads it through to the `StdioServerParameters` passed to the underlying SDK transport.

## Why

`StdioServerParameters` from `@modelcontextprotocol/sdk` already supports a `cwd` field for setting the working directory of the spawned process. `StdioConnector` was silently discarding it, which breaks servers that rely on a specific working directory — for example, those that resolve config files or `node_modules` relative to their project root.

```ts
// Before: no way to set cwd — server always starts in the process cwd
const connector = new StdioConnector({ command: "node", args: ["server.js"] });

// After: cwd is forwarded to the spawned process
const connector = new StdioConnector({
  command: "node",
  args: ["server.js"],
  cwd: "/path/to/project",
});
```

## Changes

- `src/connectors/stdio.ts`: add `cwd?: string` to the constructor param type (with JSDoc), store it as a private field, and spread it into `serverParams` only when defined (preserves existing default-to-process-cwd behaviour for callers that omit it)

Closes #1784

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/connectors/stdio.ts
+++ b/libraries/typescript/packages/mcp-use/src/connectors/stdio.ts
@@ -18,6 +18,7 @@ export class StdioConnector extends BaseConnector {
   private readonly command: string;
   private readonly args: string[];
   private readonly env?: Record<string, string>;
+  private readonly cwd?: string;
   private readonly errlog: Writable;
   private readonly clientInfo: ClientInfo;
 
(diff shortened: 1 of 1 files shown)


khandrew1:
This was already fixed by #1785 and is currently in Canary

PR #1804, opened 2026-06-28
fix(cli): coerce tool args from union schemas

DivyamTalwar:
## Problem

`mcp-use client tools call` accepts positional `key=value` arguments and uses the tool input schema to coerce strings into typed JSON values. That coercion only looked at a property's direct `type` field, so nullable schemas expressed with `anyOf` or `oneOf` were treated as untyped.

For example, a schema branch like `anyOf: [{ type: "integer" }, { type: "null" }]` left `count=42` as the string `"42"` instead of the number `42`.

## Root cause

`parseToolArgs()` built its coercion list from `propSchema.type` only. JSON Schema combinators were ignored, even though MCP tool schemas commonly use them for nullable values.

## Solution

- Collect nullable single-type schemas from direct `type` fields and `anyOf`/`oneOf` branches.
- Reuse the existing CLI coercion rules for integers, numbers, booleans, arrays, objects, and nulls.
- Add regression coverage for nullable `anyOf`/`oneOf` tool schemas, mixed unions with a string fallback, and opaque union branches that should not be coerced.

## Tests

- `pnpm --dir packages/cli exec vitest run tests/parse-args.test.ts`
- `pnpm prettier --check packages/cli/src/utils/parse-args.ts packages/cli/tests/parse-args.test.ts .changeset/cli-union-schema-args.md`
- `npx -y -p node -p pnpm.30.0 pnpm --filter mcp-use build`

## Risk

Low. This only changes known-key CLI argument coercion when the existing schema declares a simple nullable primitive through JSON Schema combinators. Unknown keys, raw JSON object args, and explicit `key:=jsonvalue` behavior are unchanged.

### DIFF
--- a/libraries/typescript/.changeset/cli-union-schema-args.md
+++ b/libraries/typescript/.changeset/cli-union-schema-args.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Fix CLI tool-call argument coercion for nullable `anyOf` and `oneOf` tool schemas.
--- a/libraries/typescript/packages/cli/src/utils/parse-args.ts
+++ b/libraries/typescript/packages/cli/src/utils/parse-args.ts
@@ -17,6 +17,8 @@ type JsonSchemaLike = {
   type?: string | string[];
   properties?: Record<string, JsonSchemaLike>;
   required?: string[];
+  anyOf?: JsonSchemaLike[];
+  oneOf?: JsonSchemaLike[];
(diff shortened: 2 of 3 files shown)


khandrew1:
Hey @DivyamTalwar! I looked into this more and I think the better move would be to avoid schema-based coercion all together. `key:=value` automatically parses values as JSON. Thanks for calling this out though! Will remove it in the v2 release

PR #1805, opened 2026-06-28
fix(python): avoid mutating adapter input schemas

DivyamTalwar:
## Problem

`BaseAdapter.fix_schema()` normalizes JSON Schema dictionaries in place. Adapter conversion paths pass MCP tool schemas into this helper, so converting a tool can rewrite the original `inputSchema` object that other code may still hold or reuse.

## Root cause

The recursive normalizer mutates each dictionary it visits and is called directly on caller-owned schema objects.

## Solution

- Deep-copy the schema once at the public `fix_schema()` boundary.
- Keep the existing recursive normalization behavior on the copied schema.
- Add a regression test proving that nullable-type and enum normalization still happen while the original schema remains unchanged.

## Tests

Fresh local verification from `libraries/python`:

- `uv run --extra dev python -m pytest tests/unit/test_enum_handling.py -q` -> `6 passed, 13 warnings`
- `uv run --extra dev ruff check mcp_use/agents/adapters/base.py tests/unit/test_enum_handling.py` -> `All checks passed!`
- `uv run --extra dev ruff format --check mcp_use/agents/adapters/base.py tests/unit/test_enum_handling.py` -> `2 files already formatted`
- `uv run --extra dev ruff check .` -> `All checks passed!`
- `uv run --extra dev ruff format --check .` -> `224 files already formatted`
- `uv run --extra dev --extra e2b python -m pytest tests/unit -q` -> `306 passed, 38 warnings in 37.01s`

## Risk

Low. The normalized return value is unchanged; this only prevents schema rewrites from leaking back into caller-owned MCP tool objects.

### DIFF
--- a/libraries/python/mcp_use/agents/adapters/base.py
+++ b/libraries/python/mcp_use/agents/adapters/base.py
@@ -5,6 +5,7 @@ This module provides the abstract base class that all MCP tool adapters should i
 """
 
 from abc import ABC, abstractmethod
+from copy import deepcopy
 from typing import Any, Generic, TypeVar
 
 from mcp.types import Prompt, Resource, Tool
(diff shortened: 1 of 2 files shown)


khandrew1:

PR #1806, opened 2026-06-28
fix(cli): display union schema types

DivyamTalwar:
## Problem

CLI schema output only reads a property's direct `type` field. Nullable schemas expressed with `anyOf` or `oneOf` are displayed as `(any)`, and array-valued types are rendered as comma-joined strings such as `(string,null)`.

That makes tool schema output harder to use because users cannot see the expected argument type from the CLI.

## Root cause

`formatSchema()` formats types with `value.type || "any"` and does not normalize JSON Schema union forms before rendering.

## Solution

- Add a schema type display helper for direct types, array-valued types, `anyOf`, and `oneOf`.
- Render union variants with ` | ` for readability.
- Add regression tests for nullable `anyOf`/`oneOf` properties and `type: ["string", "null"]`.

## Tests

- `pnpm --dir packages/cli exec vitest run tests/format.test.ts`
- `pnpm prettier --check packages/cli/src/utils/format.ts packages/cli/tests/format.test.ts .changeset/cli-schema-union-display.md`

## Risk

Low. This is display-only; it does not change CLI parsing or tool-call behavior.

### DIFF
--- a/libraries/typescript/.changeset/cli-schema-union-display.md
+++ b/libraries/typescript/.changeset/cli-schema-union-display.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Improve CLI schema display for nullable `anyOf`, `oneOf`, and array type unions.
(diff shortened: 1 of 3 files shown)


khandrew1:
Thanks for reporting this @DivyamTalwar! We'll implement a fix for this behavior in v2

PR #1807, opened 2026-06-28
fix(server): await widget structured content

DivyamTalwar:
## Problem

`generateToolOutput()` supports widget definitions with a `structuredContent` callback, but it assigns the callback result directly. If that callback is async, the tool result contains a `Promise` instead of resolved structured JSON.

## Root cause

The helper was synchronous and did not await function-valued `structuredContent`.

## Solution

- Make `generateToolOutput()` async.
- Await function-valued `structuredContent` callbacks before assigning the result.
- Await the helper at its only call site in widget tool registration.
- Add a regression test for an async structured-content callback.

## Tests

- `pnpm --dir packages/mcp-use exec vitest run tests/unit/server/protocol-helpers.test.ts`
- `pnpm prettier --check packages/mcp-use/src/server/widgets/protocol-helpers.ts packages/mcp-use/src/server/widgets/ui-resource-registration.ts packages/mcp-use/tests/unit/server/protocol-helpers.test.ts .changeset/await-widget-structured-content.md`
- `pnpm --filter mcp-use build`
- `pnpm changeset status --since=main`

Note: `pnpm --filter mcp-use test:unit` was also attempted, but this local Node 25.2.1 environment fails unrelated browser auth/export tests around built artifacts and `localStorage.clear`.

## Risk

Low. The helper has a single async call site, synchronous structured content remains supported, and the returned tool result now contains serializable structured content instead of a promise.

### DIFF
--- a/libraries/typescript/.changeset/await-widget-structured-content.md
+++ b/libraries/typescript/.changeset/await-widget-structured-content.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Await async widget `structuredContent` callbacks before returning tool output.
--- a/libraries/typescript/packages/mcp-use/src/server/widgets/protocol-helpers.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/widgets/protocol-helpers.ts
@@ -169,18 +169,18 @@ export function applyClaudeResourceDomain(
  * @param displayName - Display name for the widget
  * @returns Tool output with content array
  */
-export function generateToolOutput(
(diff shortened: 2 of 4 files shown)


khandrew1:
Thanks for the contribution! I’m not convinced we need this change in its current form.

structuredContent is not declared on UIResourceDefinition, which is why the regression test needs an as any cast. As far as I can tell, this callback shape isn’t part of the supported public API, and we haven’t seen user reports of this causing problems.

PR #1808, opened 2026-06-28
fix(cli): preserve zero screenshot timeout

DivyamTalwar:
## Problem

The screenshot command parses `--timeout` with `parseInt(options.timeout, 10) || 30000`. Because `0` is falsy, an explicit `--timeout 0` is silently replaced with the default `30000`.

## Root cause

Timeout parsing used a truthiness fallback instead of defaulting only when parsing fails.

## Solution

- Add `parseScreenshotTimeout()` to preserve numeric zero.
- Keep the existing fallback for invalid timeout strings.
- Use the helper when computing the screenshot navigation timeout.
- Add regression tests for `0` and invalid-string fallback behavior.

## Tests

- `pnpm --dir packages/cli exec vitest run tests/screenshot.test.ts`
- `pnpm prettier --check packages/cli/src/commands/screenshot.ts packages/cli/tests/screenshot.test.ts .changeset/screenshot-timeout-zero.md`
- `pnpm changeset status --since=main`

## Risk

Low. Invalid timeout input keeps the current default behavior; this only preserves an explicit numeric value that was previously lost.

### DIFF
--- a/libraries/typescript/.changeset/screenshot-timeout-zero.md
+++ b/libraries/typescript/.changeset/screenshot-timeout-zero.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Preserve explicit `--timeout 0` values when capturing CLI screenshots.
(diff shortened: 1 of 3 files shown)


tonxxd:
Hi @DivyamTalwar --timeout 0 isn't a meaningful value for screenshot readiness (it would fail immediately), so the current || 30000 fallback is fine if anything, reject non-positive timeouts like --width does instead of preserving zero.

PR #1818, opened 2026-06-29
fix(cli): respect PORT when start uses -p path

### LINKED ISSUE #1813: CLI start treats -p path alias as an explicit port flag
## Problem

`mcp-use start` defines `-p` as the short alias for `--path`, but the start command also treats any `-p` token as proof that the port was explicitly provided.

That means a command like:

```bash
PORT=4173 mcp-use start -p ./app
```

bypasses the `PORT` environment variable and falls back to the default `--port` value.

## Expected

Only `--port` / `--port=<value>` should count as an explicit port override. The `-p` path alias should not affect the existing precedence of `--port > PORT > default`.

## Impact

Users who select a project path with `-p` can unexpectedly start on the wrong port, which is especially confusing in hosted or scripted deployments where `PORT` is injected by the runtime.

DivyamTalwar:
Fixes #1813

## Problem

`mcp-use start -p <path>` uses `-p` as the short alias for `--path`, but the start command also treated `-p` as if the user had explicitly set a port.

That made `PORT=4173 mcp-use start -p ./app` ignore `PORT` and use the default `--port` value instead.

## Root cause

The explicit-port check inspected raw `process.argv` for both `--port` and `-p`, even though `start` does not define a short alias for `--port`.

## Solution

Add a small start-port helper that only treats `--port` and `--port=<value>` as explicit port flags. The existing precedence remains: explicit `--port` > `PORT` env > default.

## Tests

- `pnpm --dir libraries/typescript --filter @mcp-use/cli test -- start-port.test.ts`
- `pnpm --dir libraries/typescript exec prettier --check packages/cli/src/index.ts packages/cli/src/utils/start-port.ts packages/cli/tests/start-port.test.ts`
- `pnpm --dir libraries/typescript exec eslint packages/cli/src/index.ts packages/cli/src/utils/start-port.ts packages/cli/tests/start-port.test.ts`

## Risk

Low. This only changes explicit-port detection for `start`; `-p` continues to mean `--path`.

### DIFF
--- a/libraries/typescript/packages/cli/src/index.ts
+++ b/libraries/typescript/packages/cli/src/index.ts
@@ -37,6 +37,7 @@ import {
 } from "./utils/next-shims.js";
 import { notifyIfUpdateAvailable } from "./utils/update-check.js";
 import { getWidgetAssetBase } from "./utils/widget-paths.js";
+import { resolveStartPort } from "./utils/start-port.js";
 const program = new Command();
 
 const packageContent = readFileSync(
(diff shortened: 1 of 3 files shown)


DivyamTalwar:

**Fixes #1813.** Reproduced end-to-end with **no API key and no cloud login**, on the latest `main` rebase.

**Problem.** `mcp-use start -p ./app` silently ignores the `PORT` environment variable and always binds `3000`.

**Root cause.** The `start` command's explicit-port check scanned `process.argv` for `-p` as well as `--port`. But for `start`, `-p` is the alias of `--path` (there is no short alias for `--port`). So passing `-p <path>` falsely flagged an explicit port, short-circuiting the documented `--port > PORT env > default` precedence and forcing the `3000` default.

**Fix.** Extract `hasExplicitPortFlag()` (matches only `--port` / `--port=`) and `resolveStartPort()` into `start-port.ts`, and apply the correct precedence. Covered by 3 unit tests.

### ❌ Before — `origin/main`
before

### ✅ After — this branch
after

<sub>How it was reproduced: `node broken-repro.mjs` runs `main`'s verbatim predicate; `node fixed-repro.mjs` runs the branch's verbatim resolver — same inputs, `3000` vs `4173`. Full transcripts and the summary card are in the PR description thread.</sub>

**Risk.** Low. Pure port-resolution logic; the only behaviour change is that `-p <path>` no longer suppresses `PORT`. New unit tests lock the precedence.

khandrew1:
fixed in v2!

PR #1819, opened 2026-06-29
fix(cli): detect occupied TCP ports before startup

### LINKED ISSUE #1814: CLI port preflight misses non-HTTP TCP listeners
## Problem

The CLI port preflight checks availability by making an HTTP request to the candidate port. If a non-HTTP TCP listener is already bound there, the HTTP request fails and the CLI treats the port as available.

This can make `mcp-use dev` / `mcp-use start` select a port that is already occupied by another service.

## Expected

Port availability should be based on TCP bind semantics, not whether the existing listener speaks HTTP.

## Impact

The CLI can report that it found an available port and then fail later during server startup with an address-in-use collision. This is common when ports are occupied by databases, local daemons, or raw TCP test servers.

DivyamTalwar:
Fixes #1814

## Problem

The CLI port preflight used an HTTP request to decide whether a port was occupied. Non-HTTP TCP listeners make that request fail, so the CLI could incorrectly treat an occupied port as available and then fail later when the server tried to bind.

## Root cause

Port availability was tested at the application protocol layer (`fetch(http://host:port)`) instead of using the same TCP bind semantics the server startup path depends on.

## Solution

Move the port helpers into a tested utility and probe availability by attempting to bind a TCP server. The default `localhost` path checks both `127.0.0.1` and `::1` so an IPv4-only loopback listener is not missed when the platform resolves `localhost` to IPv6.

## Tests

- `pnpm --dir libraries/typescript --filter @mcp-use/cli test -- ports.test.ts`
- `pnpm --dir libraries/typescript exec prettier --check packages/cli/src/index.ts packages/cli/src/utils/ports.ts packages/cli/tests/ports.test.ts`
- `pnpm --dir libraries/typescript exec eslint packages/cli/src/index.ts packages/cli/src/utils/ports.ts packages/cli/tests/ports.test.ts`

## Risk

Low. This is still a best-effort preflight because any port can be claimed between the probe and final server bind, but it now detects the common non-HTTP listener collision that the previous fetch check missed.

### DIFF
--- a/libraries/typescript/packages/cli/src/index.ts
+++ b/libraries/typescript/packages/cli/src/index.ts
@@ -37,6 +37,7 @@ import {
 } from "./utils/next-shims.js";
 import { notifyIfUpdateAvailable } from "./utils/update-check.js";
 import { getWidgetAssetBase } from "./utils/widget-paths.js";
+import { findAvailablePort, isPortAvailable } from "./utils/ports.js";
 const program = new Command();
 
 const packageContent = readFileSync(
@@ -119,32 +120,6 @@ function displayPackageVersions(projectPath?: string) {
       }
     }
   }
-}
-
-// Helper to check if port is available
-async function isPortAvailable(
-  port: number,
-  host: string = "localhost"
-): Promise<boolean> {
-  try {
-    await fetch(`http://${host}:${port}`);
-    return false; // Port is in use
-  } catch {
-    return true; // Port is available
-  }
-}
-
-// Helper to find an available port
-async function findAvailablePort(
-  startPort: number,
-  host: string = "localhost"
-): Promise<number> {
-  for (let port = startPort; port < startPort + 100; port++) {
-    if (await isPortAvailable(port, host)) {
-      return port;
-    }
-  }
-  throw new Error("No available ports found");
 }
 
 // Helper to check if server is ready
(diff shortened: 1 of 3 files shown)


DivyamTalwar:

**Fixes #1814.** Reproduced end-to-end with **no API key and no cloud login**, on the latest `main` rebase.

**Problem.** The CLI's port preflight reports an occupied port as *free* when it is held by a non-HTTP TCP listener (a database, redis, any custom service), then crashes with `EADDRINUSE` when it tries to bind.

**Root cause.** `isPortAvailable()` probed at the HTTP layer with `fetch('http://host:port')`. A non-HTTP listener makes `fetch()` throw, and the `catch` returned `true` ("available"). The port is occupied, but reported free.

**Fix.** Probe with an actual TCP bind via `net.createServer().listen()` on both IPv4 and IPv6 loopback for `localhost`. Covered by 3 unit tests that bind a real ephemeral listener.

### ❌ Before — `origin/main`
before

### ✅ After — this branch
after

<sub>How it was reproduced: `broken-repro.mjs` runs `main`'s verbatim `fetch`-based helper against a raw TCP listener bound to a `:0`-discovered free port → `true`. The branch's `net`-based probe returns `false`. Full transcripts and the summary card are in the PR description thread.</sub>

**Risk.** Low. The new probe is strictly more correct (real TCP bind vs an HTTP request) and adds IPv6 coverage for `localhost`.

khandrew1:
Thanks for reporting this! It's fixed in v2

PR #1820, opened 2026-06-29
fix(cli): preserve quoted stdio command arguments

### LINKED ISSUE #1815: CLI stdio connections split quoted command paths and arguments
## Problem

`mcp-use client connect --stdio` tokenizes the command target with a raw space split. Quoted command paths and arguments are not preserved.

For example, commands with paths or config files containing spaces cannot be saved correctly:

```bash
mcp-use client connect local '"/opt/MCP Servers/server" "--config=dev env.json"' --stdio
```

## Expected

The CLI should preserve quoted command paths and arguments when converting the target into `{ command, args }`.

## Impact

Stdio connections are fragile on systems where executable paths or arguments include spaces, especially Windows and local development setups with named config files.

DivyamTalwar:
Fixes #1815

## Problem

`mcp-use client connect --stdio` split the target command on raw spaces. Quoted executable paths and arguments containing spaces were broken before being saved as `{ command, args }`.

## Root cause

The stdio connect path used `target.split(" ")`, which cannot represent quoted strings, escaped whitespace, Windows paths with spaces, or intentionally empty quoted arguments.

## Solution

Add a small shell-like tokenizer for saved stdio target strings and use it when building the command/args pair. This is intentionally limited parsing for quoting and escaping; it does not execute shell syntax or emulate a full POSIX/Windows shell.

## Tests

- `pnpm --dir libraries/typescript --filter @mcp-use/cli test -- stdio-target.test.ts`
- `pnpm --dir libraries/typescript exec prettier --check packages/cli/src/commands/client.ts packages/cli/src/utils/stdio-target.ts packages/cli/tests/stdio-target.test.ts`
- `pnpm --dir libraries/typescript exec eslint packages/cli/src/commands/client.ts packages/cli/src/utils/stdio-target.ts packages/cli/tests/stdio-target.test.ts`

## Risk

Low to moderate. This changes parsing semantics for `--stdio` command strings, but keeps the existing command-plus-argv execution model and adds coverage for simple commands, quoted paths, Windows paths, empty args, and malformed quotes.

### DIFF
--- a/libraries/typescript/packages/cli/src/commands/client.ts
+++ b/libraries/typescript/packages/cli/src/commands/client.ts
@@ -21,6 +21,7 @@ import {
   isStdoutTty,
 } from "../utils/format.js";
 import { parsePromptArgs, parseToolArgs } from "../utils/parse-args.js";
+import { parseStdioTarget } from "../utils/stdio-target.js";
 import {
   buildOAuthProvider,
   isUnauthorized,
@@ -171,9 +172,7 @@ async function connectCommand(
     const cliClientInfo = getCliClientInfo();
 
     if (options.stdio) {
-      const parts = target.split(" ");
-      const command = parts[0];
-      const args = parts.slice(1);
+      const { command, args } = parseStdioTarget(target);
 
       console.error(
         formatInfo(`Connecting to stdio server: ${command} ${args.join(" ")}`)
--- a/libraries/typescript/packages/cli/src/utils/stdio-target.ts
+++ b/libraries/typescript/packages/cli/src/utils/stdio-target.ts
@@ -1,0 +1,92 @@
+export interface StdioTarget {
+  command: string;
+  args: string[];
+}
+
+export function parseStdioTarget(target: string): StdioTarget {
+  const parts = splitCommandLine(target);
+
+  if (parts.length === 0) {
+    throw new Error("Stdio command cannot be empty");
+  }
+
+  const [command, ...args] = parts;
+  return { command, args };
+}
+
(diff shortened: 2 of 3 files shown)


DivyamTalwar:

**Fixes #1815.** Reproduced end-to-end with **no API key and no cloud login**, on the latest `main` rebase.

**Problem.** Connecting a stdio MCP server whose executable path (or an argument) contains spaces silently mangles the saved command — e.g. `/opt/MCP Servers/server` or Windows `C:\Program Files\...`.

**Root cause.** The stdio connect path tokenized the target with `target.split(" ")`, taking `parts[0]` as the command and the rest as args. A raw space split cannot represent quoted paths or arguments containing spaces, so the command and args were torn apart and persisted incorrectly.

**Fix.** Add a `parseStdioTarget()` tokenizer that honours quotes and backslashes, and use it in `connectCommand`. Covered by 5 unit tests (quoted paths, Windows backslashes, empty quoted args, unterminated-quote rejection).

### ❌ Before — `origin/main`
before

### ✅ After — this branch
after

<sub>How it was reproduced: `broken-repro.mjs` runs `main`'s verbatim `target.split(" ")` on a quoted path; the branch's `parseStdioTarget()` is exercised by its unit tests. Full transcripts and the summary card are in the PR description thread.</sub>

**Risk.** Low. Simple stdio commands parse identically; only quoted/spaced targets change. Unterminated quotes are rejected with a clear error.

khandrew1:
STDIO servers aren't super heavily used, and we removed support in v2 anyways. Thanks for the report!

PR #1821, opened 2026-06-29
fix(server): handle empty OpenAPI JSON responses

### LINKED ISSUE #1816: OpenAPI tools throw on empty JSON success responses
## Problem

OpenAPI-generated tools call `response.json()` whenever a successful response declares a JSON content type. Successful empty JSON responses, especially `204 No Content`, cause `response.json()` to throw `SyntaxError: Unexpected end of JSON input`.

## Expected

A successful JSON response with an empty body should complete without throwing and return an empty structured object.

## Impact

Valid OpenAPI operations that return `204` or otherwise have no JSON body are exposed as failing MCP tools even though the upstream API call succeeded.

DivyamTalwar:
Fixes #1816

## Problem

OpenAPI-generated tools threw `SyntaxError: Unexpected end of JSON input` for successful responses that declared a JSON content type but had an empty body, such as `204 No Content`.

## Root cause

The JSON response path called `response.json()` unconditionally for successful JSON content types.

## Solution

Read the response body as text first. Non-empty JSON bodies are parsed as before; empty JSON success bodies return `objectResponse({})`, preserving a structured JSON result without throwing.

## Tests

- `pnpm --dir libraries/typescript --filter mcp-use exec vitest run tests/unit/server/openapi.test.ts`
- `pnpm --dir libraries/typescript --filter mcp-use build`
- `pnpm --dir libraries/typescript exec prettier --check packages/mcp-use/src/server/openapi/request.ts packages/mcp-use/tests/unit/server/openapi.test.ts`
- `pnpm --dir libraries/typescript exec eslint packages/mcp-use/src/server/openapi/request.ts packages/mcp-use/tests/unit/server/openapi.test.ts`

## Risk

Low. Empty JSON success responses now surface as `{}` / `"{}"` instead of throwing. Non-empty JSON and text responses keep the existing behavior.

### DIFF
--- a/libraries/typescript/.changeset/openapi-empty-json-responses.md
+++ b/libraries/typescript/.changeset/openapi-empty-json-responses.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Handle successful OpenAPI JSON responses with empty bodies without throwing.
--- a/libraries/typescript/packages/mcp-use/src/server/openapi/request.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/openapi/request.ts
@@ -41,7 +41,8 @@ export async function callOpenAPIOperation(
     contentType.includes("application/json") ||
     contentType.includes("+json")
   ) {
-    return objectResponse(await response.json());
+    const responseText = await response.text();
+    return objectResponse(responseText ? JSON.parse(responseText) : {});
   }
 
   return text(await response.text());
--- a/libraries/typescript/packages/mcp-use/tests/unit/server/openapi.test.ts
+++ b/libraries/typescript/packages/mcp-use/tests/unit/server/openapi.test.ts
@@ -255,4 +255,28 @@ describe("MCPServer.fromOpenAPI", () => {
     );
     expect(result?.content).toEqual([{ type: "text", text: "updated" }]);
   });
+
+  it("handles empty JSON success responses", async () => {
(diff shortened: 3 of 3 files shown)


DivyamTalwar:

**Fixes #1816.** Reproduced end-to-end with **no API key and no cloud login**, on the latest `main` rebase.

**Problem.** An OpenAPI-derived MCP tool crashes when the upstream API returns a successful response with an empty body (e.g. `204 No Content`) but a JSON content-type.

**Root cause.** `callOpenAPIOperation()` unconditionally called `await response.json()` for any `application/json` (or `+json`) response. An empty body makes `Response.json()` throw `SyntaxError: Unexpected end of JSON input`, turning a successful upstream call into a failing tool.

**Fix.** Read the body as text first and guard the empty case before parsing JSON. Covered by a new `handles empty JSON success responses` test.

### ❌ Before — `origin/main`
before

### ✅ After — this branch
after

<sub>How it was reproduced: Check out the branch, revert **only** `request.ts` to `origin/main`, and run the suite — the new test fails at `request.ts:44` on `await response.json()`. Restore the fix → 8/8. Full transcripts and the summary card are in the PR description thread.</sub>

**Risk.** Low. Only the empty-body path changes; non-empty JSON responses parse exactly as before.

cc @khandrew1 — this touches code you own; happy to adjust the approach if you'd prefer something different.

khandrew1 (approved):

PR #1822, opened 2026-06-29
fix(python): keep missing auth unauthenticated

### LINKED ISSUE #1817: Python config treats missing auth as an empty auth object
## Problem

Python connector config creation passes `auth={}` when a server config omits the `auth` field. That makes an absent auth value behave differently from `None`:

- Minimal HTTP configs initialize OAuth state from the empty dict.
- Minimal WebSocket configs enter the unsupported non-bearer auth warning path.

## Expected

Omitting `auth` from config should create an unauthenticated connector. Auth handling should only be enabled when the user explicitly supplies an auth value.

## Impact

Minimal configs can unexpectedly initialize auth behavior or emit auth warnings, which makes unauthenticated HTTP/WebSocket setups harder to reason about and can trigger OAuth discovery later.

DivyamTalwar:
Fixes #1817

## Problem

Python connector config creation passed `auth={}` when the config omitted `auth`. That made minimal configs look like configured auth objects:

- HTTP initialized OAuth state from the empty dict.
- WebSocket entered the unsupported non-bearer auth warning path.

## Root cause

`create_connector_from_config()` used `server_config.get("auth", {})` for HTTP and WebSocket connectors instead of preserving `None` for an absent optional field.

## Solution

Pass `server_config.get("auth")` so omitted auth remains `None`. Explicit auth values continue to flow through unchanged.

## Tests

- `uv run --project libraries/python pytest libraries/python/tests/unit/test_config.py`
- `uv run --project libraries/python pytest libraries/python/tests/unit/test_http_connector.py libraries/python/tests/unit/test_websocket_connection_manager.py`
- `uv run --project libraries/python ruff check libraries/python/mcp_use/client/config.py libraries/python/tests/unit/test_config.py`
- `uv run --project libraries/python ruff format --check libraries/python/mcp_use/client/config.py libraries/python/tests/unit/test_config.py`

## Risk

Low. This makes missing auth behave as unauthenticated; users who want OAuth or bearer auth still need to supply the `auth` field explicitly.

### DIFF
--- a/libraries/python/mcp_use/client/config.py
+++ b/libraries/python/mcp_use/client/config.py
@@ -113,7 +113,7 @@ def create_connector_from_config(
         return HttpConnector(
             base_url=server_config["url"],
             headers=server_config.get("headers", None),
-            auth=server_config.get("auth", {}),
+            auth=server_config.get("auth"),
             timeout=server_config.get("timeout", 5),
             sse_read_timeout=server_config.get("sse_read_timeout", 60 * 5),
             sampling_callback=sampling_callback,
@@ -136,7 +136,7 @@ def create_connector_from_config(
         return WebSocketConnector(
             url=server_config["ws_url"],
             headers=server_config.get("headers", None),
-            auth=server_config.get("auth", {}),
+            auth=server_config.get("auth"),
         )
 
     raise ValueError("Cannot determine connector type from config")
--- a/libraries/python/tests/unit/test_config.py
+++ b/libraries/python/tests/unit/test_config.py
@@ -95,6 +95,7 @@ class TestConnectorCreation(unittest.TestCase):
         self.assertEqual(connector.base_url, "http://test.com")
         self.assertEqual(connector.headers, {})
         self.assertIsNone(connector._auth)
+        self.assertIsNone(connector._oauth)
 
     def test_create_websocket_connector(self):
         """Test creating a WebSocket connector from config."""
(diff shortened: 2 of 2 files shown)


DivyamTalwar:

**Fixes #1817.** Reproduced end-to-end with **no API key and no cloud login**, on the latest `main` rebase.

**Problem.** A server config with no `auth` key is treated as if it *had* auth: the HTTP connector builds a live OAuth client, and the WebSocket connector logs a spurious "unsupported auth" warning.

**Root cause.** `create_connector_from_config()` defaulted missing auth to an empty dict via `server_config.get("auth", {})`. Connectors only skip auth handling when it is `None`; an empty dict is non-`None`, so `HttpConnector._set_auth({})` took the OAuth branch and `WebSocketConnector` hit its unsupported-auth `else`.

**Fix.** Use `server_config.get("auth")` so an omitted key stays `None` and the connector remains truly unauthenticated. New regression test asserts no auth warning for a minimal WebSocket config.

### ❌ Before — `origin/main`
before

### ✅ After — this branch
after

<sub>How it was reproduced: A two-line `create_connector_from_config({'url': ...})._oauth` probe plus the `pytest` regression test, run against branch `config.py` (fixed) and against `origin/main`'s `config.py` (broken). Full transcripts and the summary card are in the PR description thread.</sub>

**Risk.** Low. Behaviour only changes for configs that omit `auth` — which are meant to be unauthenticated. Configs that pass an explicit (even empty) `auth` object are unaffected.

**Status.** Rebased on the latest `main`; GitHub reports **`mergeable = MERGEABLE`** with a single clean commit authored by me. The changed files pass the repo's formatter/linter (ruff check + ruff format). CI: **Socket Security** and **label** checks pass; the heavier test/typecheck workflows are gated behind maintainer approval for a first-time contributor, so they show as pending — not failing on this diff.

cc @pietrozullo — this touches code you own; happy to adjust the approach if you'd prefer something different.

khandrew1:

PR #1824, opened 2026-06-29
fix(openapi): handle empty JSON success responses without throwing

### LINKED ISSUE #1816: OpenAPI tools throw on empty JSON success responses

syf2211:
## Summary

OpenAPI-generated MCP tools no longer throw when a successful upstream response declares a JSON content type but has an empty body (for example `204 No Content`).

## Motivation

Fixes #1816. `response.json()` raises `SyntaxError: Unexpected end of JSON input` on empty bodies, causing valid API calls to surface as failing MCP tools.

## Changes

- Parse JSON success responses via `response.text()` first in `callOpenAPIOperation`
- Return an empty structured object when the body is blank
- Add a regression test for a `204` response with `application/json` content type

## Tests

```bash
cd libraries/typescript/packages/mcp-use
pnpm build
pnpm exec vitest run tests/unit/server/openapi.test.ts
```

Result: 8/8 tests passed.

## Notes

- Whitespace-only JSON bodies are treated as empty and return `{}`
- Invalid JSON still throws from `JSON.parse`, matching prior `response.json()` behavior

### DIFF
--- a/libraries/typescript/packages/mcp-use/src/server/openapi/request.ts
+++ b/libraries/typescript/packages/mcp-use/src/server/openapi/request.ts
@@ -41,7 +41,11 @@ export async function callOpenAPIOperation(
     contentType.includes("application/json") ||
     contentType.includes("+json")
   ) {
-    return objectResponse(await response.json());
+    const responseBody = await response.text();
+    if (!responseBody.trim()) {
+      return objectResponse({});
+    }
+    return objectResponse(JSON.parse(responseBody));
   }
 
   return text(await response.text());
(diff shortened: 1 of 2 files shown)


khandrew1:
Fixed by #1821

PR #1836, opened 2026-07-05
fix(telemetry): don't send raw prompt/response text by default

unixsvr4:
## Changes

By default, the `mcp_agent_execution` telemetry event captured the full `query` and `response` strings and sent them to PostHog whenever anonymized telemetry was enabled (the default). For a library that runs inside other people's agents, transmitting raw end-user prompts and model responses to a third party by default is a privacy/compliance concern (GDPR, enterprise data-handling policies).

This PR makes raw prompt/response content **opt-in** while keeping the analytically-useful length signals on by default.

## Implementation Details

1. `telemetry/events.py` — widen `MCPAgentExecutionEvent.query` to `str | None` (it can now be redacted; `response` was already `str | None`).
2. `telemetry/telemetry.py` — read `MCP_USE_TELEMETRY_INCLUDE_CONTENT` (default `false`) once in `Telemetry.__init__`. In `track_agent_execution`, always derive `query_length` / `response_length` from the real content, then null out `query` / `response` unless the user opted in.
3. No change to the wire format beyond the two fields now defaulting to `null`; `query_length` / `response_length` are unchanged.

## Example Usage (Before)

```json
"query": "Help me analyze sales data from the CSV file",
"response": "I'll help you analyze the sales data..."
```

## Example Usage (After)

Default:
```json
"query": null,
"query_length": 44,
"response": null,
"response_length": 38
```

Opt back in (e.g. on a self-hosted PostHog you own):
```bash
export MCP_USE_TELEMETRY_INCLUDE_CONTENT=true
```

## Documentation Updates

- `docs/python/development/telemetry.mdx` — updated "What We Collect" / "What We DON'T Collect", added a "Prompt & Response Content" section documenting the new default and the opt-in var, and updated the example event.

## Testing

- Added `tests/unit/test_telemetry_content_optin.py` covering (a) content redacted by default with lengths preserved, (b) content included when opted in, (c) the env default resolving to `false`.
- `pytest tests/unit/test_telemetry_content_optin.py tests/unit/test_agent.py -p no:logfire` → 17 passed.
- `ruff check` clean on the touched files.

## Backwards Compatibility

Not a code-level breaking change. It **does** change the default telemetry payload: raw `query`/`response` are no longer sent unless `MCP_USE_TELEMETRY_INCLUDE_CONTENT=true`. Anyone relying on that content in their own PostHog dashboards can restore it with the env var.

### DIFF
--- a/docs/python/development/telemetry.mdx
+++ b/docs/python/development/telemetry.mdx
@@ -22,7 +22,7 @@ MCP-use includes an **opt-out telemetry system** that helps us understand how th
 ### Agent Execution Data
 When you use `MCPAgent.run()` or `MCPAgent.stream()`, we collect:
 
-- **Query and response content** (to understand use cases)
+- **Query and response _length_** (character counts, always) — the raw text is **not** collected unless you opt in (see below)
 - **Model provider and name** (e.g., "openai", "gpt-4")
 - **MCP servers connected** (types and count, not specific URLs/paths)
 - **Tools used** (which MCP tools are popular)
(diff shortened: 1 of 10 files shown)


tonxxd:
Thanks @unixsvr4 for the pr, please create a more focused pr that does not touch as many files

unixsvr4:
Thanks for the review @tonxxd — you were right, that branch had an unrelated bug-fix commit riding along on it.

Rescoped in #2385: 4 files instead of 10 (2 source, 1 test, 1 doc), rebased on current `main`. The WebSocket connector / prompt builder / error formatting fixes are dropped from this one — happy to open them separately if they're of interest.

PR #1860, opened 2026-07-09
fix(cli): do not treat -p path alias as explicit port flag

### LINKED ISSUE #1813: CLI start treats -p path alias as an explicit port flag

syf2211:
## Summary

Fix `mcp-use start` port detection so the `-p` path alias no longer bypasses the `PORT` environment variable.

## Motivation

`start` defines `-p` as the short alias for `--path`, but the command also treated any `-p` token as proof that `--port` was explicitly provided. That broke the intended precedence (`--port` > `PORT` > default) for commands like:

```bash
PORT=4173 mcp-use start -p ./app
```

## Changes

- Add `isStartPortFlagExplicit()` helper that only recognizes `--port` / `--port=<value>`
- Use the helper in the `start` command instead of scanning raw `process.argv` for `-p`
- Add unit tests covering `-p ./app`, `-p=./app`, and explicit `--port` forms

## Tests

- `pnpm test tests/start-port.test.ts` — 3 passed

## Notes

composer-2.5 review: APPROVE

Fixes #1813

---
## Summary by cubic
Fixes port detection in `mcp-use start` so the `-p` path alias for `--path` no longer overrides the `PORT` environment variable. Restores precedence: `--port` > `PORT` > default.

- **Bug Fixes**
  - Add `isStartPortFlagExplicit()` to only accept `--port`/`--port=<value>`.
  - Use the helper in `start` instead of scanning `process.argv` for `-p`.
  - Add unit tests for `-p ./app`, `-p=./app`, and explicit `--port` forms.

<sup>Written for commit 1c93fcbdfc277edae0c55290acd4ba65ec2939a5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/1860?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/cli/src/index.ts
+++ b/libraries/typescript/packages/cli/src/index.ts
@@ -20,6 +20,7 @@ import {
 } from "./commands/client.js";
 import { getSession } from "./utils/session-storage.js";
 import { formatError } from "./utils/format.js";
+import { isStartPortFlagExplicit } from "./utils/start-port.js";
 import { deployCommand } from "./commands/deploy.js";
 import { createDeploymentsCommand } from "./commands/deployments.js";
 import { createServersCommand } from "./commands/servers.js";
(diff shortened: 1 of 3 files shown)


khandrew1:
Thanks for the PR! This is fixed in v2

PR #1892, opened 2026-07-16
feat(server): proxy multiple MCP servers

khandrew1:
## Summary by cubic
Adds `server.proxy()` to compose multiple HTTP MCP upstreams behind one `mcp-use` endpoint. Upstreams are namespaced and their tools, static resources, and prompts are proxied with cancellation and ordered progress; failures are logged without failing the call.

- **New Features**
  - HTTP-only config map with `url`, `authToken`/`headers` (optional `timeout`, `fetch`, protocol negotiation); OAuth/browser flow disabled and stdio not supported.
  - Overload to mount an existing connection from `@mcp-use/client` v2; its negotiated server name becomes the namespace.
  - Namespacing prefixes `<namespace>_`; static resources use `mcp-use-proxy:///`. Forwards tool schemas/results, cancellation, and progress; progress forwarding errors are logged and do not fail the call.
  - Best-effort mounting with diagnostics for connection/introspection failures and name collisions; must be called before `listen()`/`getHandler()` (calling after start or after `close()` errors).
  - Lifecycle: config-created connections are owned and closed in `server.close()`; shutdown waits for in-flight `proxy()` operations and closes all owned clients (aggregates close errors). Caller-supplied connections are not closed. Importing/running without calling `proxy()` does not load the optional `@mcp-use/client` peer; missing peer prints an install hint.
  - Scope: v2 proxies tools, static resources, and prompts; it does not proxy resource templates, completions, subscriptions, or upstream list-change re-sync.

- **Migration**
  - Install `@mcp-use/client` if you use `server.proxy()`: `npm install @mcp-use/client`.
  - Call `server.proxy()` before starting the server.
  - Use bearer tokens or headers for auth; the proxy does not perform OAuth and will never open a browser.
  - Choose unique namespaces; collisions are skipped. Manage the lifecycle if you pass your own connection.

<sup>Written for commit 902a37f0374f5d711f0806d0c12efc868b7c1c50. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/1892?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/server/src/proxy.ts
+++ b/libraries/typescript/packages/server/src/proxy.ts
@@ -53,7 +53,7 @@ export interface ProxyRequestOptions {
   /** Aborts the upstream request when the downstream request is cancelled. */
   signal?: AbortSignal;
   /** Receives upstream progress notifications. */
-  onprogress?: (progress: ProxyProgress) => void | Promise<void>;
+  onprogress?: (progress: ProxyProgress) => void;
 }
 
 /** Structural connection contract accepted by the low-level proxy overload. */
@@ -353,20 +353,32 @@ function mountPlan(host: ProxyMountHost, plan: ProxyNamespacePlan): void {
     const callback = async (
       params: Record<string, unknown>,
       ctx: RequestContext
-    ) =>
-      plan.connection.callTool(upstreamName, params, {
-        signal: ctx.signal,
-        onprogress: (progress) => {
-          void ctx
-            .reportProgress(progress.progress, progress.total, progress.message)
-            .catch((error: unknown) => {
-              proxyDiagnostic(
-                `Failed to forward progress for proxied tool "${definition.name}"`,
-                error
-              );
+    ) => {
+      let progressForwarding = Promise.resolve();
+      try {
+        return await plan.connection.callTool(upstreamName, params, {
+          signal: ctx.signal,
+          onprogress: (progress) => {
+            progressForwarding = progressForwarding.then(async () => {
+              try {
+                await ctx.reportProgress(
+                  progress.progress,
+                  progress.total,
+                  progress.message
+                );
+              } catch (error) {
+                proxyDiagnostic(
+                  `Failed to forward progress for proxied tool "${definition.name}"`,
+                  error
+                );
+              }
             });
-        },
-      });
+          },
+        });
+      } finally {
+        await progressForwarding;
+      }
+    };
     host.registerTool(definition, callback as ToolCallback);
   }
 
--- a/libraries/typescript/packages/server/src/proxy.ts
+++ b/libraries/typescript/packages/server/src/proxy.ts
@@ -9,49 +9,15 @@ import type {
   ToolAnnotations,
 } from "@modelcontextprotocol/server";
 import type {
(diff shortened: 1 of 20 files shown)


libraries/typescript/packages/server/src/server.ts
```diff
@@ -403,6 +411,59 @@ export class MCPServer<TUser = never> {
...
+   *   weather: { url: "https://weather.example.com/mcp" },
+   * });
+   * ```
+   */
+  async proxy(servers: Record<string, ProxyServerConfig>): Promise<void>;
+  /**
+   * Mount an existing ready `@mcp-use/client` v2 connection.
+   *
+   * @param connection - Ready connection to introspect and forward through.
+   * @param options - Optional namespace. Without one, upstream names and URIs
```
khandrew1: we should just automatically namespace. that way it won't throw if there's a collision

libraries/typescript/packages/server/src/server.ts
```diff
@@ -403,6 +411,59 @@ export class MCPServer<TUser = never> {
...
+   * are preserved.
+   * @throws If introspection fails, a mounted name collides, or the server
```
khandrew1: if introspection does fail, we shouldn't throw. we should just log an error and the tools won't show up which is fine

libraries/typescript/packages/server/src/proxy.ts
```diff
@@ -0,0 +1,549 @@
...
+  ServerConfig as ClientServerConfig,
+} from "@mcp-use/client";
+
+import type { RequestContext } from "./context.js";
+import type { PromptCallback, PromptDefinition } from "./prompts.js";
+import type { ResourceCallback, ResourceDefinition } from "./resources.js";
+import type { ToolCallback, ToolDefinition } from "./tools.js";
+
+/** Automatic OAuth settings forwarded to `@mcp-use/client` v2. */
+export interface ProxyOAuthOptions {
```
khandrew1: this name is not descriptive enough. it should be ClientProxyOAuthOptions

libraries/typescript/packages/server/src/proxy.ts
```diff
@@ -0,0 +1,549 @@
...
+  /** Fetch implementation used for upstream HTTP requests. */
+  fetch?: typeof fetch;
+  /** Disable automatic OAuth, or configure the client's automatic OAuth flow. */
+  oauth?: false | ProxyOAuthOptions;
+  /** Protocol negotiation mode forwarded to `@mcp-use/client`. */
+  protocolNegotiation?: "auto" | "legacy" | { pin: string };
+}
+
+/** Node stdio connection settings accepted by {@link MCPServer.proxy}. */
+export interface ProxyStdioConfig {
```
khandrew1: remove stdio config

PR #1935, opened 2026-07-19
Fix stale v2 view examples and conformance artifacts

khandrew1:
## Summary

- update public create-mcp-use-app and related README examples to the current `view` API and `views/<name>/view.tsx` convention
- replace stale `useMcp` examples with current React hooks
- migrate the server conformance example away from widget-era source and generated artifacts while preserving its intended lifecycle, host-context, debug, and display behavior
- keep the conformance tool-registration declaration in `tools.d.ts`
- update focused inspector and view-discovery coverage for the current paths

## Why

Several published examples and conformance fixtures still referenced removed widget-era APIs and generated files. They no longer matched the executable v2 scanner and templates on `beta`, which use the `views/<name>/view.tsx` layout.

## Validation

- conformance TypeScript check
- direct v2 conformance build
- create-mcp-use-app tests: 20 passed
- server view-discovery tests: 5 passed
- ESLint on changed inspector tests and helpers
- `git diff --check`

The full inspector typecheck still reports the pre-existing unused `options` parameter in `src/client/auth/manufact-auth.ts`. No `@mcp-use/agent` source or ESM/CJS behavior was changed.


khandrew1:
@cubic-dev-ai review this PR

PR #1949, opened 2026-07-20
fix(mcp-use): stop leaking parent process env to stdio servers

VihaanAgarwal:
## Language / Project Scope

- [x] TypeScript

## Changes

Addresses the STDIO environment variable leak from #1901 (the first of the three items in that report).

The stdio connector built the child process environment by copying the entire parent `process.env` and then layering the caller's `env` on top. So the moment a user configured even one server-specific variable, every secret in the parent environment (API keys, tokens, cloud credentials) was handed to the spawned MCP server.

This restores parity with the underlying `@modelcontextprotocol/sdk` transport, which deliberately does not inherit the full parent env: when no `env` is provided it starts from `getDefaultEnvironment()`, a minimal allowlist (`HOME`, `PATH`, `SHELL`, ...). The connector now uses that same safe base when the caller does provide `env`.

## Review Readiness

- [x] I verified the affected package with the commands listed below

## Implementation Details

1. Extracted the env construction into a small pure `buildStdioEnv()` helper in `connectors/stdio.ts` so the security-relevant logic is directly unit-testable.
2. The helper returns `undefined` when no `env` is configured (unchanged, lets the transport apply its own default) and otherwise returns `{ ...getDefaultEnvironment(), ...env }`.
3. No new dependencies. `getDefaultEnvironment` is imported from `@modelcontextprotocol/sdk/client/stdio.js`, already a dependency and the source of `StdioServerParameters` used here.

Verified with:

```
pnpm run test:types
pnpm exec eslint src/connectors/stdio.ts tests/unit/connectors/stdio-env.test.ts
pnpm build && pnpm run test:unit   # 626 passed
```

Added `tests/unit/connectors/stdio-env.test.ts` covering: no leak of arbitrary parent env, caller variables preserved, the safe default base is present, and the `undefined` passthrough. The leak test fails against the previous behaviour and passes with this change.

### Packages Modified

- [x] `mcp-use`

Added a changeset (`mcp-use`: patch).

---
## Summary by cubic
Stop leaking the full parent process environment to stdio MCP servers in `mcp-use`. The connector now uses the SDK’s safe defaults plus the provided `env`, preventing secret exposure and matching `@modelcontextprotocol/sdk` behavior.

- Bug Fixes
  - Build child env as `{ ...getDefaultEnvironment(), ...env }`; no longer copies `process.env`.
  - If no `env` is set, pass `undefined` so the transport applies its default.
  - Extracted `buildStdioEnv()` and added unit tests for non-leakage and default base.

<sup>Written for commit 2238fa0fd53aa9519d2d30e6ec6c6ce635d474aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/1949?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.changeset/stdio-env-leak.md
+++ b/.changeset/stdio-env-leak.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": patch
+---
+
+Stop leaking the full parent process environment to stdio MCP servers. When a server is configured with its own `env`, the connector now layers those variables on top of the SDK's `getDefaultEnvironment()` safe base (`HOME`, `PATH`, ...) instead of copying all of `process.env`, matching the default behaviour of the underlying `@modelcontextprotocol/sdk` transport.
--- a/libraries/typescript/packages/mcp-use/src/connectors/stdio.ts
+++ b/libraries/typescript/packages/mcp-use/src/connectors/stdio.ts
@@ -1,6 +1,7 @@
 import type { StdioServerParameters } from "@modelcontextprotocol/sdk/client/stdio.js";
 import type { Writable } from "node:stream";
 
+import { getDefaultEnvironment } from "@modelcontextprotocol/sdk/client/stdio.js";
 import { Client } from "@modelcontextprotocol/sdk/client/index.js";
 import process from "node:process";
 import type { ConnectorInitOptions } from "./base.js";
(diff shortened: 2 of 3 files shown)


khandrew1:
Thanks for this @VihaanAgarwal! I ported it over to v2 in this PR: #2125

Issue #1871, opened 2026-07-13
Python README per-primitive/transport CI badges all show identical whole-workflow status (shields.io ignores job= param)


ayaangazali:
## Description

`libraries/python/README.md` renders a "Supports" matrix of per-primitive and per-transport CI badges (Tools, Resources, Prompts, Sampling, Elicitation, Roots, Authentication, Stdio, SSE, Streamable HTTP) using shields.io URLs like:

```
https://img.shields.io/github/actions/workflow/status/mcp-use/mcp-use/ci.yml?job=python-primitive/tools
```

shields.io's workflow-status badge does not support a `job` query parameter — it's silently ignored, so every badge in the matrix renders the **whole-workflow** status of `ci.yml` on the default branch.

## Reproduction

```
$ curl -s ".../ci.yml?job=python-primitive/tools"    -o badge1.svg
$ curl -s ".../ci.yml?job=python-primitive/sampling" -o badge2.svg
$ diff -q badge1.svg badge2.svg
# identical — byte-for-byte
```

Both currently render **"failing"**, as do all ten badges, because the latest `ci.yml` run on `main` concluded `failure` — so the README currently tells readers that every primitive and every transport is broken, which per-job results contradict.

## Impact

The matrix is the first thing a prospective user sees when evaluating per-feature health, and it can show all-red (as it does right now) even when every job it claims to describe is green — or all-green while a specific primitive is broken. Either way it never carries the per-feature signal it promises.

## Suggested fix

shields.io and GitHub's native badge endpoint are both workflow-level only, so honest options are:

1. **Collapse the matrix to one "CI" badge** for `ci.yml` and let the table list supported primitives/transports without per-cell status (simplest, no infra); or
2. Keep real per-feature badges by publishing job results to a gist/endpoint from CI and using shields.io endpoint badges (accurate but adds CI plumbing).

Happy to PR option 1 (or 2 if preferred) — option 1 is a small README-only change.

cyforkk:
I'd like to work on this. My plan: collapse the per-primitive/transport badge matrix into a single CI badge (option 1 from the issue), keeping the support matrix as a plain table without per-cell status. Will submit a PR shortly.

PR #1996, opened 2026-07-23
docs(python): replace misleading per-job CI badges in README supports matrix

ayaangazali:
## What

The "Supports" matrix in `libraries/python/README.md` rendered per-primitive and per-transport CI badges from shields.io workflow-status URLs with a `?job=` query parameter, for example:

`https://img.shields.io/github/actions/workflow/status/mcp-use/mcp-use/ci.yml?job=python-primitive/tools`

shields.io's workflow-status badge has no `job` parameter, so it is silently ignored and every one of the ten badges renders the **whole-workflow** status of `ci.yml` on the default branch. All ten cells show the same value (all "failing" right now, while the individual jobs actually pass), so the matrix never carries the per-feature signal it promises.

Verified that the badges are byte-identical regardless of the `job` value:

```
curl -s ".../ci.yml?job=python-primitive/tools"    -o a.svg
curl -s ".../ci.yml?job=python-primitive/sampling" -o b.svg
diff a.svg b.svg   # identical
```

## Change

This is option 1 from #1871: drop the per-cell status badges and keep the supported primitives and transports as a plain list, so the README says what is supported without falsely implying per-feature CI health. README only, no code or CI touched.

If you would rather keep real per-feature status (option 2: publish job results to an endpoint/gist and use shields.io endpoint badges), I am happy to do that instead. This was the smaller, no-infra fix.

Refs #1871

## Heads-up: the failing `python-lint` check is not from this change

`python-lint` runs `pip install ruff` (unpinned) then `ruff format --check .`. As of ruff 0.16.0 (released recently), `ruff format` also formats fenced code blocks inside Markdown, and this README's Python examples were never formatted that way. So `ruff format --check` now flags pre-existing formatting in the example blocks (e.g. the multi-line `config = {...}` dict), and the failure reproduces on a clean checkout for any PR that touches `libraries/python`:

```
uvx ruff@0.16.0 format --check libraries/python/README.md   # "1 file would be reformatted"
uvx ruff@0.15.22 format --check libraries/python/README.md  # passes (older ruff skips Markdown)
```

I left the examples as-is on purpose: running `ruff format` collapses the readable multi-line config example into a single 100+ char line, which seems worse for a tutorial. This looked like a maintainer call between (a) pinning ruff in the `python-lint` step for reproducible CI, or (b) excluding Markdown from `ruff format`, or (c) accepting the reformat. Happy to send a separate small PR for whichever you prefer.

---
Freshman dev here, still learning this codebase. I used Claude Code to help draft this and verified the badge behavior myself. Please push back on anything off and I will fix it.

**CI status (python-lint):** `ruff format --check .` reports `224 files already formatted` locally on this HEAD, so nothing in the diff is unformatted. The red check comes from the CI ruff version also reformatting the Python fenced code blocks inside `README.md` (the quickstart examples around lines 151 to 240), which this PR does not touch and which report the same on `main`. Not fixable from a rerun or this docs-only change; it is a repo-side call (format those blocks, pin ruff in the lint step, or exclude Markdown from `ruff format`).


ayaangazali:
Quick note on the red python-lint check for reviewers: it is not from this change. This PR only edits the Supports badge table (lines 47 to 57). The ruff format check is flagging the Python code blocks embedded lower in the README (the quickstart examples around lines 151 to 240), and those same blocks report as unformatted on main too, since the python-lint job runs ruff format over the whole libraries/python tree whenever anything there changes. I left the example blocks alone on purpose, since ruff would collapse the multi-line config JSON into a single line and make the tutorial harder to read. Happy to fix them in a separate PR if you would like that formatting applied. This change itself is docs-only and does not touch any Python.

ayaangazali:
Following up on the CI note in the description: I opened #2069 to pin ruff in the `python-lint` job, which is the root cause of the red check here and on #1856. It also lays out the two alternatives (scope CI to Python files like `.pre-commit-config.yaml` already does, or adopt 0.16.0 and reformat) in case you prefer one of those.

ayaangazali:
Heads up on the python jobs that just went red here: that is not from this branch. `mcp 2.0.0` released today and removed `RequestContext` from `mcp.shared.context`, which `mcp_use` imports in `client/connectors/base.py`, so test collection dies at import with:

```
ImportError: cannot import name 'RequestContext' from 'mcp.shared.context'
```

`main` itself went red the same way earlier today (the 22:40 UTC run failed, 19:09 was green). I opened #2077 to cap the dependency at `mcp>=1.24.0,<2`, which restores 1.29.0 and makes the import work again. Once that lands this branch should go green on a rerun.

Good news on the other one: `python-lint` is gone here now, since #2069 pinned ruff.

ayaangazali:
Closing this in favor of #1872, which fixes the same thing and was opened ten days earlier.

I raised the underlying issue (#1871) and then this PR went up without anyone checking that @axelray-dev already had one open for it, which was my mistake. Theirs is the same one file change and it is green, so there is no reason for both to sit in the queue.

The finding itself, for whoever reviews #1872: shields.io ignores the `job=` query parameter, so all ten per-primitive and per-transport badges render the whole-workflow status. I verified it by fetching two badges with different `job=` values and diffing them, and the SVGs came back byte identical.

PR #2055, opened 2026-07-29
fix(python): collapse CI badge matrix into single badge

### LINKED ISSUE #1871: Python README per-primitive/transport CI badges all show identical whole-workflow status (shields.io ignores job= param)

cyforkk:
## Description

Closes #1871

The per-primitive/transport CI badge matrix in `libraries/python/README.md` uses shields.io with a `job=` query parameter, but shields.io silently ignores it. Every badge renders the overall `ci.yml` workflow status instead of per-job status, so the matrix always shows identical (currently all-failing) badges regardless of individual job results.

This PR collapses the matrix into a single CI badge and keeps the support matrix as a plain text table, matching option 1 suggested in the issue.

## Changes

- `libraries/python/README.md`: replaced the shields.io badge matrix with a single CI badge and a plain-text support matrix.

## Checklist

- [x] PR title follows conventional commits
- [x] Linked issue #1871
- [x] Change is minimal and scoped to the issue

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>

---
## Summary by cubic
Replaced the broken per-job CI badge matrix in `libraries/python/README.md` with a single CI badge and a plain-text support matrix to avoid misleading status. Closes #1871.

- **Bug Fixes**
  - Shields.io ignores the `job=` query, so all badges showed the overall `ci.yml` status.
  - Collapsed to one CI badge and kept the support matrix as plain text.

<sup>Written for commit 3c61547ef2d4817649054b7da52c9480f627f7cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2055?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khandrew1:
done by #1872

PR #2062, opened 2026-07-30
Feature/reliability recovery clean

### LINKED ISSUE #2054: Example: recover a multi-server mutation after ambiguous success
## Use case

`mcp-use` already has clear multi-server examples, but there is no runnable example for a common reliability boundary: a mutating tool commits its external write and then its result is lost. A normal retry can duplicate the mutation even though MCP itself did not retry anything.

This is especially useful in a multi-server agent setup where one server performs domain work and a second sidecar coordinates operation identity, checkpoints, and recovery.

## Proposed example

Add one deterministic Python example (and TypeScript later only if maintainers want parity) that:

1. starts a local synthetic destination MCP with a duplicate-sensitive `create_item` tool and a read-back tool keyed by an opaque operation marker;
2. connects through the existing multi-server `MCPClient` path;
3. injects a first-call failure after the destination commits;
4. shows an unguarded retry creating two items;
5. shows the guarded path record `external_result_uncertain`, call the read-back tool, recover the committed item, and avoid a second create;
6. prints a compact reliability report with tool-call count, mutation count, verification result, and the honest `duplicate-resistant` guarantee.

The default fixture would be fully local, credential-free, and model-free so CI can reproduce it. An optional README section could show how to replace the in-memory guard with any remote sidecar implementing the same contract, including the free Agent Enhancer endpoint, without making the hosted service a test dependency.

No core retry behavior would change. This is a workflow example because only the caller can combine destination semantics, operation identity, and read-after-write verification safely.

## Evidence and prior art

- [Reliability Sidecar Contract v1](https://github.com/artiehinz/Agent-Enhancer-Utilities/blob/main/docs/RELIABILITY_SIDECAR_CONTRACT_V1.md)
- [Agno cookbook implementation #9178](https://github.com/agno-agi/agno/pull/9178)
- [Direct gated-write replay fence in mcp-odoo #61](https://github.com/erpipe-org/mcp-odoo/pull/61)
- [MCPJam fixture proposal #3555](https://github.com/MCPJam/inspector/issues/3555)

The example would make no exactly-once claim and would stop for review when read-back cannot prove the outcome. If this fits the examples direction, I can submit it with focused tests and the required Python quality checks.

nk-pavan-official:
Language / Project Scope
 Python
Summary

Adds a Python example showing how to safely recover from ambiguous MCP mutation failures — cases where a mutation might have succeeded on the server but the client never got the response.

What's included
A synthetic MCP server (synthetic_destination_server.py) with a create_item mutation tool and a read_back verification tool, used to simulate the ambiguous-success scenario.
An unguarded (unsafe) retry workflow that shows how blind retries can create duplicate mutations.
A guarded (safe) recovery workflow that verifies state via read_back before deciding whether to retry.
Files changed

Added: libraries/python/examples/synthetic_destination_server.py
Simulates a server that commits a mutation but doesn't reliably confirm it back to the client, so you can test recovery logic against it.

Updated: libraries/python/examples/reliability_recovery_example.py
Adds two workflows:

Unguarded — retries immediately after an ambiguous failure, without checking if the original mutation already went through. Demonstrates the duplicate-mutation risk.
Guarded — uses an operation marker as an idempotency key, calls read_back to check if the mutation already committed, and only retries if it didn't.
Testing

Ran locally:

ruff format libraries/python/examples/reliability_recovery_example.py libraries/python/examples/synthetic_destination_server.py
ruff check libraries/python/examples/reliability_recovery_example.py libraries/python/examples/synthetic_destination_server.py
python libraries/python/examples/reliability_recovery_example.py

Both files pass formatting and lint. Verified actual script output:

--- 1. UNGUARDED WORKFLOW (Naive Retry) ---
[Unguarded] Ambiguous result received.
[Unguarded] Retrying create_item(marker='op_unguarded_101')...
[Unguarded] Naive retry finished.

--- 2. GUARDED WORKFLOW (Read-Back Verification) ---
[Guarded] Calling create_item(marker='op_guarded_202')...
[Guarded] Checking read_back for marker 'op_guarded_202'...
[Guarded] SUCCESS: Found committed item: {'id': 'item_3', 'operation_marker': 'op_guarded_202', 'name': 'Widget B'}
[Guarded] Skipping retry. Duplicate mutation prevented!
Backwards compatibility

No existing APIs touched. Pure addition of a new example.

Related issue

Closes #2054

---
## Summary by cubic
Adds a Python example that shows how to recover safely from ambiguous MCP mutation results, using read-back verification to avoid duplicate writes. Implements the recovery pattern requested in #2054.

- **New Features**
  - Adds `libraries/python/examples/synthetic_destination_server.py`: Synthetic MCP server with `create_item` and `read_back`; can simulate ambiguous success via `simulate_network_drop`.
  - Adds `libraries/python/examples/reliability_recovery_example.py`: Demonstrates naive retry vs guarded recovery using an operation marker as an idempotency key plus `read_back` before retrying.
  - No API changes; standalone examples only.

- **Bug Fixes**
  - Check `isError` before interpreting tool results to avoid misclassifying failures as ambiguous success.
  - Use `sys.executable` in the client config instead of a hardcoded interpreter path.

<sup>Written for commit da3ce21c0b838c5aa59b9e01da60f2875e7fef62. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2062?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/examples/reliability_recovery_example.py
+++ b/libraries/python/examples/reliability_recovery_example.py
@@ -16,7 +16,6 @@ Issue #2054
 """
 
 import asyncio
-from typing import Dict, Optional
 
 
 class SyntheticDestinationServer:
@@ -25,7 +24,7 @@ class SyntheticDestinationServer:
     """
 
     def __init__(self):
-        self.database: Dict[str, dict] = {}
+        self.database: dict[str, dict] = {}
         self.total_tool_calls: int = 0
         self.total_mutations: int = 0
 
@@ -77,7 +76,7 @@ class SyntheticDestinationServer:
     async def read_back(
         self,
         operation_marker: str,
-    ) -> Optional[dict]:
+    ) -> dict | None:
         """
         Checks whether mutation already happened.
         """
--- a/libraries/python/examples/reliability_recovery_example.py
+++ b/libraries/python/examples/reliability_recovery_example.py
@@ -1,34 +1,34 @@
 """
 Reliability & Recovery Example: Multi-Server Mutation Recovery
 
-Demonstrates how to safely handle ambiguous tool failures:
-- Destination server commits a mutation
-- Response is lost
-- Client does NOT blindly retry
-- Client verifies using read-back before retrying
+Demonstrates how to safely handle ambiguous tool failures where a mutation
+succeeds on the destination server but the response is lost due to network
+failure.
+
+Workflow comparison:
+1. Unguarded Path:
+   Retry immediately after ambiguous failure -> duplicate mutation.
+
+2. Guarded Path:
+   Verify external state using operation marker -> recover without duplicate.
 
 Issue #2054
 """
 
 import asyncio
-from typing import Any, Dict, Optional
-
-from mcp_use import MCPClient
+from typing import Dict, Optional
 
 
 class SyntheticDestinationServer:
     """
-    Synthetic destination MCP server.
-
-    Simulates an external system with:
-    - create_item mutation tool
-    - read_back verification tool
+    Simulates a duplicate-sensitive destination MCP server.
     """
 
     def __init__(self):
-        self.database: dict[str, dict] = {}
-        self.total_tool_calls = 0
-        self.total_mutations = 0
+        self.database: Dict[str, dict] = {}
+        self.total_tool_calls: int = 0
+        self.total_mutations: int = 0
+
 
     async def create_item(
         self,
@@ -36,6 +36,9 @@ class SyntheticDestinationServer:
         item_name: str,
         simulate_network_drop: bool = False,
     ) -> dict:
+        """
+        Mutating tool that creates an item.
+        """
 
         self.total_tool_calls += 1
         self.total_mutations += 1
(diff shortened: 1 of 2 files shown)


nk-pavan-official:
Fixed all three: added result.isError checks in both the unguarded workflow and the guarded workflow's read-back step, so real tool errors aren't misread as ambiguous success or a committed item. Also replaced the hardcoded interpreter path with sys.executable and resolved the server script path relative to __file__, so the example isn't tied to a specific machine/OS anymore.

Verified locally: ruff format/check clean, and reran the script end to end with matching expected output.

nk-pavan-official:
Hi @Me333-jjj, thanks for taking a look at the PR and for reaching out!

That sounds interesting. I'd be interested in learning more about your protocol for ensuring state consistency in tool-calling agents. If you have documentation, a repository, or a write-up that explains the approach, please feel free to share it. I'm always interested in seeing different approaches to handling ambiguous mutation outcomes and recovery patterns.

Thanks!

khandrew1:
Thanks for this PR! Though I'm not a big fan of this example. From the issue:

> Our examples should primarily demonstrate mcp-use APIs and supported patterns. We already expose relevant tool metadata such as idempotentHint; the broader guidance is simply that callers should not blindly retry non-idempotent writes after uncertain transport failures.

Issue #1899, opened 2026-07-17
unable to pass stderr


patricklx:
**Describe the bug**
we are unable to configure MCP process spawn stderr, and the default is to passthrough to parent stderr (inherit):
https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/client/src/client/stdio.ts#L33C5-L33C11

**To Reproduce**
1. create an mcp server with bad command
2. start the mcp server
3. notice the errors in stderr
4. no way to configure it

**Expected behavior**
We should be able to configure the stderr, e.g. pass custom WriteStream to handle error displaying

patricklx:
additional note: we probably need 2 pass 2 things: "pipe" + custom stream

ayaangazali:
Confirmed, and it's slightly worse than "no way to configure it" — mcp-use already has the configuration surface, but it's a silent no-op:

- `StdioConnector` accepts an `errlog: Writable` option, documented as "Stream to which the server's stderr should be piped" (`src/connectors/stdio.ts:29`, `src/task_managers/stdio.ts:16-21`).
- But `StdioConnectionManager.establishConnection()` creates the transport as `new StdioClientTransport(this.serverParams)` **without** `stderr: "pipe"`. The upstream SDK defaults to `"inherit"` (the line you linked), and its `transport.stderr` getter returns `null` unless the mode is `"pipe"`/`"overlapped"`.
- So the forwarding block right below (`if (this._transport.stderr && ...) pipe(this.errlog)`) is dead code — `errlog` has never received a byte, and the child's stderr passes straight through to the parent, exactly as you observed.

Fix is small: spawn the transport with `stderr: "pipe"` (unless the caller explicitly set a stderr mode in server params), which makes the existing forwarding block live and `errlog` functional. Default `errlog` is `process.stderr`, so out-of-the-box behavior stays visually identical — but a custom `WriteStream` now actually captures the output, which is your use case.

Working on a PR with a real-process test (child writing to stderr, custom Writable asserting capture).

ayaangazali:
PR up: #1902 — spawns the child with `stderr: "pipe"` so the documented `errlog` stream actually receives the child's stderr (your custom `WriteStream` case), with real-process tests proving capture and the explicit-override path.

khandrew1:
Hey @patricklx @ayaangazali!

Thanks for reporting this, we've been hard at work on v2 so sorry for the wait.

[...]

PR #2079, opened 2026-08-01
docs: list the recipe template's prompts separately from its tools

ayaangazali:
# Pull Request Description

## Language / Project Scope

- [ ] TypeScript
- [ ] Python
- [x] Documentation only
- [ ] CI/CD or tooling

## Changes

The Recipe Finder template is listed with four tools in two places, but two of those four are prompts, not tools:

- `README.md`, Templates table (Tools column)
- `docs/home/templates.mdx`, Recipe Finder card

Querying the demo server that both entries link to:

```
tools/list    ->  search-recipes, get-recipe
prompts/list  ->  meal-plan, recipe-suggestion
```

This matters more than a naming nit. A model can call a tool on its own, but it cannot invoke a prompt, so someone reading the table expects `meal-plan` to be callable and it is not. That is what #1883 ran into: the recipe card's "Plan a meal" button asks the model to plan a meal, no `meal-plan` tool is exposed to it, and the model just restates the recipe.

## Review Readiness

- [x] The PR is scoped to one issue or one clearly related change
- [x] The PR description explains the user-visible problem and the fix
- [x] The diff avoids unrelated formatting, generated files, and bulk rewrites
- [x] I verified the affected path with the commands listed below
- [x] I checked for existing open PRs that already solve the same issue

## Implementation Details

Two lines. The README table keeps its Tools column and notes the prompts inline, since adding a column would churn every row:

```
| Recipe Finder | `search-recipes`, `get-recipe` (plus `meal-plan` and `recipe-suggestion` prompts) |
```

The docs card has room for the clearer split:

```
Tools: `search-recipes`, `get-recipe`. Prompts: `meal-plan`, `recipe-suggestion`.
```

I checked the other template entries in the same table and page. Recipe Finder is the only one mixing prompts into its tools list, so nothing else needed touching.

## Testing

Verified against the live demo server linked in the docs, over MCP:

```
initialize -> tools/list    : search-recipes, get-recipe
initialize -> prompts/list  : meal-plan (cuisine, dietary, days),
                              recipe-suggestion (ingredients, mealType)
```

`npx mint broken-links` on `docs/` passes.

## Backwards Compatibility

Documentation only.

## Related Issues

Related to #1883. This does not fix the button in the template repo, it just stops the docs from promising a tool that is not there. The behavior fix belongs in `mcp-use/mcp-recipe-finder` and needs a call on whether `meal-plan` becomes a real tool or the button sends a self contained message, which I raised on that issue.

---
[...]

PR #2086, opened 2026-08-02
fix(client): make stdio errlog actually receive the child's stderr

ayaangazali:
# Pull Request Description

## Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

**Targets `beta`.** This is the v2 port of #1902, which fixed the same bug against `main`. The code moved to `packages/client/src/transport/stdio.ts` in v2 and the bug came with it, so I would rather fix it where you are actually shipping. Close #1902 if you prefer to take only this one.

## Changes

`StdioConnectionManager` accepts an `errlog: Writable` and its doc comment says the child server's stderr is piped there. It never worked. The transport is created as:

```ts
this._transport = new StdioClientTransport(this.serverParams);
```

with no `stderr` option, so the SDK applies its default of `"inherit"`. In that mode `transport.stderr` is null, which means the block right below it that pipes stderr into `errlog` is dead code. The child's stderr goes straight to the parent process and there is no way to capture it.

## Review Readiness

- [x] I verified the affected package with the commands listed below

## Implementation Details

One spread, so `"pipe"` is the default while an explicit `stderr` in server params still wins:

```ts
this._transport = new StdioClientTransport({
  stderr: "pipe",
  ...this.serverParams,
});
```

Plus a changeset and two regression tests.

## Testing

I could not run the package suite locally, since `pnpm install` on this branch fails resolving an `@types/react-dom` peer, so the included tests are for CI to run. To make sure the premise was right I checked the behavior directly against the exact package this file imports, `@modelcontextprotocol/client@2.0.0`:

```
current beta (no stderr option)   transport.stderr = null   -> errlog pipe never runs
with stderr:'pipe' (this fix)     transport.stderr = stream -> errlog gets the output
```

In the first case the child's stderr marker appears on the parent's stderr instead, which is the bug as users see it.

The two tests spawn real child processes rather than mocking:

- a custom `errlog` receives a marker the child writes to stderr, which fails without this change
- `stderr: "ignore"` in server params still opts out, and `transport.stderr` stays null

The same pair passed against the v1 copy of this code in #1902, including the red-without-fix check.

## Backwards Compatibility

`errlog` still defaults to `process.stderr`, so output lands in the same place by default. The child's stderr now arrives over a pipe rather than an inherited fd, which can change TTY detection in a child that colors its output; `stderr: "inherit"` in server params restores the old behavior exactly.

## Related Issues

Fixes the v2 occurrence of #1899. Port of #1902.

---
Freshman dev here, still learning this codebase. I used Claude Code to help with this and verified the transport behavior myself against the pinned SDK. Please push back on anything off and I will fix it.

---
## Summary by cubic
Fixes the v2 client so `errlog` receives the child process stderr by default, adds a public `stderr` mode, and safely handles write errors without closing the caller’s stream.

- **Bug Fixes**
  - Default `stderr: "pipe"` when creating `StdioClientTransport` and forward to `errlog` only when the mode leaves a stream; pipe with `{ end: false }`, unpipe on close, and log/clean up on stream errors.
  - Expose `StdioStderrMode` ("pipe" | "inherit" | "ignore") on `StdioConnector` and `StdioServerConfig`, and pass it through config to server params; tests spawn real children to verify forwarding, opt-outs, errlog reuse, and error handling.

<sup>Written for commit 05003f5b7f88cd0000f06f1fbc4e245e16aa3c59. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2086?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ayaangazali:
Retargeted from `beta` to `main` now that v2 has shipped. `beta` is 0 commits ahead of `main` and fully contained in it, so it is no longer the branch that receives this, and leaving the PR pointed there would have made it a no-op or auto-closed it if the branch gets cleaned up.

Confirmed the fix is still needed on current `main` before switching, not just carried over on faith. Still `MERGEABLE` with the same diff.

ayaangazali:
@khandrew1 quick targeting question, since you moved #2099 and #2100 to `canary` by hand.

Should TypeScript package changes go to `canary` rather than `main` now? This one and #2085 are the same category as the two you retargeted:

* #2086 (this) `packages/client` source
* #2085 `packages/create-mcp-use-app` templates

I have deliberately not retargeted them myself. `canary` is 3 ahead and 4 behind `main` and does not contain it, so switching the base is not mechanical and I would rather you point them than have me guess and create conflicts. Say the word and I will rebase both onto `canary` properly.

The rest of my open ones look main-shaped to me, but correct me if not:

* #2116 `docs.yml` pnpm pin, fixes the `Generate and validate API documentation` red on every branch
* #2115 removes the dead SDK fork workflow
* #2101 and #2026 `lychee.toml` excludes, clear the weekly Link Checker Report
* #2114 TypeScript README links into the deleted `packages/mcp-use`
* #2079 one line in `docs/home/templates.mdx`

khandrew1:
@ayaangazali yeah any PRs that would require a release should be made off canary. i usually just end up retargeting it myself before merging anyways so it's up to you

ayaangazali:
Thanks, that is the rule I needed. Saves you the retarget, so I did it properly on both:

* #2086 (this) rebased onto `canary`, now `2d6bb243`, 3 files, mergeable
* #2085 rebased onto `canary`, now `d23ff10f`, 4 files, mergeable

Both rebased cleanly with no conflicts, and `pnpm vitest run tests/unit/transport/stdio-errlog.test.ts` still passes 2/2 on the canary base. Preflight clean.

I left the other open ones on `main` since none of them need a release: #2116 and #2115 are workflow files, #2101 and #2026 are `lychee.toml`, #2114 is the monorepo TypeScript README, #2079 is one line of `docs/home/templates.mdx`. Move any of them if I have that wrong.

ayaangazali:
`typescript-build (node-22.23.1)` here is pre-existing on `canary`, not from this diff. It appeared only because the base moved from `main` to `canary`.

The failing step is `Test release versioning`, and the same two assertions fail on `canary` itself:

```
$ gh run view 30835754617 --json jobs      # canary, 2026-08-03T17:13:19Z
[{ "n": "typescript-build (node-22.23.1)",
   "steps": ["Test release versioning"] }]

$ gh run view 30835754617 --log-failed | grep "not ok"
not ok 6 - avoids false dependent majors for a Client patch beta
not ok 7 - avoids false dependent majors for a Client minor beta
# fail 2
```

Identical pair on both of my rebased branches, same step, same numbers. And the same job is green on `main`:

```
main   run 30821461954  14:10:51Z  typescript-build (node-22.23.1)  success
canary run 30835754617  17:13:19Z  typescript-build (node-22.23.1)  failure
```

So it is specific to `canary`, and nothing in these diffs touches release versioning. This PR is package.json template metadata and a client transport change plus its test.

One hypothesis, flagged as unverified: those scenarios in `tests/release-versioning.test.mjs` assert hardcoded `-beta.N` versions, and `canary` is in prerelease mode under a `canary` tag via `.changeset/pre.json`. If the version script now emits `-canary.N`, patch and minor would fail on the version string. That does not explain why the `major` scenario still passes, so I have not convinced myself and am not sending a fix on a guess. Happy to dig in properly if it is useful, otherwise I assume it is part of your in-flight release work.

khandrew1 (changes requested):
Thanks for fixing this—the diagnosis makes sense, and defaulting the SDK transport to `stderr: "pipe"` restores the intended `errlog` behavior.

Could we expose `stderr` through `StdioConnector` and `StdioServerConfig`, then forward it into `StdioServerParameters`? I think the desired API is:

- Default `stderr` to `"pipe"` so providing `errlog` works without extra configuration.
- Allow users to explicitly select `"inherit"` to preserve the previous TTY/file-descriptor behavior.
- Allow `"ignore"` when stderr should be discarded.
- Only forward `transport.stderr` to `errlog` when the selected mode produces a readable stream.

Currently, an override works only for callers that instantiate `StdioConnectionManager` directly. `StdioConnector` does not accept or forward `stderr`, and `StdioServerConfig` does not expose it, so users of the normal public APIs cannot use the compatibility escape hatch described in the PR.

I'd also suggest piping with `{ end: false }` and unpiping during cleanup:

```ts
transport.stderr.pipe(errlog, { end: false });
```

`errlog` is caller-owned, so a child process exiting should not close it. This matters when the writable is reused across reconnects or multiple connectors.

With that API, the original use case works by default:

```ts
new StdioConnector({ command, args, errlog });
```

while users can retain the old behavior explicitly:

```ts
new StdioConnector({ command, args, stderr: "inherit" });
```

ayaangazali:
Implemented the API you described, in 4002deeb.

**`StdioStderrMode`** is now `"pipe" | "inherit" | "ignore"`, exported, defaulting to `"pipe"`. It is accepted on `StdioConnector` and on `StdioServerConfig`, and threaded through the config path in `core/node.ts`, so both of your examples work:

```ts
new StdioConnector({ command, args, errlog });              // forwards by default
new StdioConnector({ command, args, stderr: "inherit" });    // previous behaviour
```

Also done from your notes and the review:

* Pipe with `{ end: false }` and `unpipe` on close, so a caller-owned `errlog` survives a child exiting and can be reused across reconnects.
* Only forward `transport.stderr` when the mode actually leaves a readable stream, rather than relying on a duck-typed `pipe` check.
* Default applied with `??` after the spread, so an explicit `stderr: undefined` no longer silently defeats it.
* Added an `error` handler on the forwarding source. Now that this path runs by default, a destroyed `errlog` would otherwise surface as an unhandled `'error'` event.

On the test cubic flagged: it was right, the non-pipe case never called `transport.start()`, so the SDK never spawned a child and `expect(data()).toBe("")` passed by construction. The tests now start the transport, assert `transport.stderr` is null, and wait for the child to write and exit so an accidental forward would have landed. `"inherit"` and `"ignore"` are both covered, and `"inherit"` visibly writes to the test runner's own stderr, which is the behaviour being preserved.

Verification: 4/4 in `stdio-errlog.test.ts`, full client suite 182 passed, `tsc --noEmit` clean, preflight exit 0. The suite also reports 16 unhandled errors locally, all `Cannot find module '@csstools/css-calc'` from jsdom in the React tests; identical with my changes stashed, so unrelated.

The changeset now spells out the default change and the `"inherit"` escape hatch. I left the client README alone since it does not currently document `errlog` at all, but happy to add a short stderr section there if you would like it covered.

khandrew1 (approved):

PR #2092, opened 2026-08-02
fix(python): initialize base connector state in WebSocketConnector

### LINKED ISSUE #2088: Python WebSocketConnector never initializes client_session, crashes on is_connected
**Describe the bug**

`WebSocketConnector.__init__` (`libraries/python/mcp_use/client/connectors/websocket.py`) never calls `super().__init__()` and never sets `self.client_session`. It also doesn't override the `is_connected` property, so it falls back to `BaseConnector.is_connected` (`client/connectors/base.py`), whose first line is `if not self.client_session:`.

```python
class WebSocketConnector(BaseConnector):
    def __init__(self, url, headers=None, auth=None):
        self.url = url
        self.headers = headers or {}
        # ... no super().__init__(), no self.client_session
```

**To Reproduce**

1. Configure a server with `"ws_url"` in the client config.
2. Connect via `MCPClient`.
3. Access `session.is_connected` (e.g. `MCPSession.is_connected` in `client/session.py`).
4. Raises `AttributeError: 'WebSocketConnector' object has no attribute 'client_session'`.

**Expected behavior**

`is_connected` should reflect actual connector state without raising.

**Severity:** High — affects every WebSocket-transport server.

Adityaj0:
## Changes
`WebSocketConnector.__init__` never called `super().__init__()`, so `client_session` and other `BaseConnector` attributes were never set. Accessing `is_connected` raised `AttributeError`, and any inherited method this connector doesn't override (e.g. `list_prompts`/`get_prompt`/`set_roots`) would fail the same way.

## Implementation Details
1. Call `super().__init__()` at the top of `WebSocketConnector.__init__` so all `BaseConnector` attributes are initialized.
2. Override `is_connected` on `WebSocketConnector` to use its own `_connected` flag instead of the inherited `client_session`-based check — this connector speaks the MCP protocol directly over raw WebSocket messages and never populates `client_session`, so the inherited check would otherwise always report "not connected" even while actually connected.

## Example Usage (Before)
```python
connector = WebSocketConnector(url="ws://localhost:8765")
connector.is_connected  # AttributeError: 'WebSocketConnector' object has no attribute 'client_session'
```

## Example Usage (After)
```python
connector = WebSocketConnector(url="ws://localhost:8765")
connector.is_connected  # False, no crash
```

## Documentation Updates

## Testing
- Added `tests/unit/test_websocket_connector.py` covering: base attributes are initialized, `is_connected` doesn't raise before connecting, and `is_connected` correctly tracks the connector's own `_connected` flag.
- `pytest tests/unit` passes (307 passed; the 1 pre-existing failure is an unrelated missing optional `e2b` dependency in the test env, unaffected by this change).

## Backwards Compatibility
Not breaking. Constructor signature is unchanged; this only fixes attribute initialization and connection-state reporting.

## Related Issues
Closes #2088

---
## Summary by cubic
Fixes `WebSocketConnector` initialization so base state is set and connection status is accurate, including detecting when the receiver task has died. Prevents AttributeError when accessing `is_connected` and inherited methods, and avoids stale "connected" after drops.

- **Bug Fixes**
  - Call `super().__init__()` in `WebSocketConnector.__init__` to initialize `BaseConnector` attributes.
  - Override `is_connected` to use `_connected` and verify the receiver task is running; stop relying on `client_session`.
  - Add unit tests for initialization and connection state, including receiver task death.

<sup>Written for commit 65e8f7d57a8c3581bcddbf45c2bef104fae652b4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2092?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/client/connectors/websocket.py
+++ b/libraries/python/mcp_use/client/connectors/websocket.py
@@ -264,6 +264,14 @@ class WebSocketConnector(BaseConnector):
 
         WebSocketConnector manages the connection over raw WebSocket messages
         rather than an MCP ClientSession, so unlike the base implementation this
-        does not depend on ``client_session``.
+        does not depend on ``client_session``. It also checks that the receiver
+        task is still running: ``_receive_messages`` catches its own exceptions
+        and returns without resetting ``_connected``, so a dropped connection
+        would otherwise still report as connected until ``disconnect()`` is
+        called explicitly.
         """
-        return self._connected
+        if not self._connected:
+            return False
+        if self._receiver_task is None or self._receiver_task.done():
+            return False
+        return True
--- a/libraries/python/mcp_use/client/connectors/websocket.py
+++ b/libraries/python/mcp_use/client/connectors/websocket.py
@@ -42,6 +42,8 @@ class WebSocketConnector(BaseConnector):
                 - A dict: Not supported for WebSocket (will log warning)
                 - An httpx.Auth object: Not supported for WebSocket (will log warning)
         """
+        super().__init__()
+
         self.url = url
         self.headers = headers or {}
 
@@ -255,3 +257,13 @@ class WebSocketConnector(BaseConnector):
     def public_identifier(self) -> str:
         """Get the identifier for the connector."""
         return f"websocket:{self.url}"
+
+    @property
+    def is_connected(self) -> bool:
+        """Check if the connector is actually connected.
+
+        WebSocketConnector manages the connection over raw WebSocket messages
+        rather than an MCP ClientSession, so unlike the base implementation this
+        does not depend on ``client_session``.
+        """
+        return self._connected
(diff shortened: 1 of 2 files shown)


Adityaj0:
Good catch — fixed in 65e8f7d5. `is_connected` now also checks that `_receiver_task` exists and hasn't finished, since `_receive_messages()` swallows its own exceptions and returns without resetting `_connected`. Added regression tests for both the missing-task and dead-task cases.

khandrew1:

PR #2095, opened 2026-08-02
fix(python): return all resource content blocks, handle empty contents

### LINKED ISSUE #1625: Resource tool ( _convert_resource)returns only last content block and crashes on empty contents

Adityaj0:
## Changes
Fixes the two bugs described in #1625: `ResourceTool._arun` in the LangChain adapter (`mcp_use/agents/adapters/langchain_adapter.py`) overwrote its accumulator variable on every loop iteration, so only the last content block of a multi-block resource was ever returned — earlier blocks were silently dropped. An empty `contents` list also raised `UnboundLocalError` since the accumulator was never assigned.

## Implementation Details
1. Accumulate each decoded content block into a list and `"\n".join(...)` them at the end, per the fix already suggested in #1625.

## Example Usage (Before)
```python
# resource.contents == ["page 1", "page 2", "page 3"]
await resource_tool._arun()  # -> "page 3" (data loss)

# resource.contents == []
await resource_tool._arun()  # -> UnboundLocalError
```

## Example Usage (After)
```python
# resource.contents == ["page 1", "page 2", "page 3"]
await resource_tool._arun()  # -> "page 1\npage 2\npage 3"

# resource.contents == []
await resource_tool._arun()  # -> ""
```

## Documentation Updates

## Testing
- Added `test_resource_tool_returns_all_content_blocks` and `test_resource_tool_empty_contents_returns_empty_string` in `tests/unit/test_langchain_adapter.py`.
- Verified both new tests fail against the pre-fix code and pass after the fix.
- `pytest tests/unit` passes (306 passed; the 1 pre-existing failure is an unrelated missing optional `e2b` dependency in the test env).

## Backwards Compatibility
Not breaking. Return value for multi-block resources changes from a single (wrong, truncated) block to the full joined content, which is the intended/expected behavior per #1625.

## Related Issues
Closes #1625

---
## Summary by cubic
Return all resource content blocks in the Python LangChain adapter and handle empty resources. Previously only the last block was returned and empty contents crashed with UnboundLocalError; now we decode, collect, and join parts with newlines, returning an empty string when empty.

<sup>Written for commit 7c3a0f21582539d744600ce3292321e456fd2777. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2095?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
+++ b/libraries/python/mcp_use/agents/adapters/langchain_adapter.py
@@ -233,14 +233,15 @@ class LangChainAdapter(BaseAdapter[BaseTool]):
                 logger.debug(f'Resource tool: "{self.name}" called')
                 try:
                     result = await self.tool_connector.read_resource(mcp_resource.uri)
+                    parts = []
                     for content in result.contents:
                         # Attempt to decode bytes if necessary
                         if isinstance(content, bytes):
-                            content_decoded = content.decode()
+                            parts.append(content.decode())
                         else:
-                            content_decoded = str(content)
+                            parts.append(str(content))
 
-                    return content_decoded
+                    return "\n".join(parts)
                 except Exception as e:
                     if self.handle_tool_error:
                         return format_error(e, tool=self.name)  # Format the error to make LLM understand it
--- a/libraries/python/tests/unit/test_langchain_adapter.py
+++ b/libraries/python/tests/unit/test_langchain_adapter.py
@@ -9,6 +9,8 @@ from mcp.types import (
     CallToolResult,
     EmbeddedResource,
     ImageContent,
+    ReadResourceResult,
+    Resource,
     TextContent,
     TextResourceContents,
     Tool,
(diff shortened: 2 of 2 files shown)


khandrew1:
Thanks for the detailed report. Our focus right now is on the TypeScript SDK, so we're limiting Python work to problems that users have run into firsthand. This one looks like it was found by reading the code rather than by hitting it in practice, so we're closing it as not planned for now. If it is affecting you or someone you know in a real setup, reply with what you were doing when it came up and we'll reopen it.

PR #2096, opened 2026-08-02
fix(python): make telemetry() decorator await async methods

### LINKED ISSUE #2091: Python telemetry() decorator does not await wrapped async methods
**Describe the bug**

The `telemetry()` decorator in `libraries/python/mcp_use/telemetry/telemetry.py` defines `wrapper` as a plain `def`, not `async def`:

```python
def telemetry(event_name, additional_properties=None):
    def decorator(func):
        @wraps(func)
        def wrapper(self, *args, **kwargs):
            ...
            try:
                result = func(self, *args, **kwargs)
                return result
            except Exception as e:
                ...
            finally:
                ...
                telemetry.capture(...)
        return wrapper
    return decorator
```

When applied to an `async def` method (e.g. `MCPSession.initialize`/`call_tool`, `BaseConnector.call_tool`/`list_tools`, `OAuth.initialize`/`authenticate`/`refresh_token`), `func(self, *args, **kwargs)` only constructs a coroutine object — it doesn't run the method body. The `finally` block then records `success=True` and near-zero `execution_time_ms` immediately, before the coroutine is ever awaited by the caller, and the `except` block can never catch real errors from the wrapped call since those only surface later when the caller awaits the returned coroutine.

**Impact:** every telemetry event for these decorated async methods (most of the library's public async surface) reports fabricated instant-success data and never captures real failures or true timing.

**Severity:** High

Adityaj0:
## Changes
The `telemetry()` decorator's `wrapper` was always a plain `def`, even when decorating an `async def` method (e.g. `MCPSession.initialize`/`call_tool`, `BaseConnector.call_tool`/`list_tools`, `OAuth.initialize`/`authenticate`/`refresh_token`). Calling `func(self, *args, **kwargs)` on an async method just constructs a coroutine object without running its body. The decorator's `finally` block then recorded `success=True` and near-zero `execution_time_ms` immediately — before the wrapped method's body ever actually ran — and its `except` block could never catch real errors from the call, since those only surface later when the caller awaits the returned coroutine.

## Implementation Details
1. `telemetry()` now checks `asyncio.iscoroutinefunction(func)` and, for async methods, returns a proper `async def` wrapper that `await`s the wrapped coroutine before capturing the event.
2. Extracted the event-capture logic (looking up `self.telemetry`/`self._telemetry`, building `GenericTelemetryEvent`) into a shared `_capture` helper used by both the sync and async wrappers, to avoid duplicating that logic.
3. The sync path is otherwise unchanged.

## Example Usage (Before)
```python
@telemetry("server_initialize")
async def initialize(self):
    await asyncio.sleep(2)  # real work
    raise RuntimeError("connection refused")

# Telemetry recorded: success=True, execution_time_ms≈0, error_type=None
# (fabricated — the real exception surfaces later, uncaught by the decorator)
```

## Example Usage (After)
```python
# Telemetry recorded: success=False, execution_time_ms≈2000, error_type="RuntimeError"
```

## Documentation Updates

## Testing
- Added `tests/unit/test_telemetry.py` covering both async and sync decorated methods: successful execution, exception propagation + capture, and that `execution_time_ms` reflects the coroutine's actual runtime (guards against the bug where timing was captured before the coroutine ran).
- Verified 2 of the 3 new async tests fail against the pre-fix code (timing ≈0ms, exceptions not correctly captured as failures) and all pass after the fix.
- `pytest tests/unit` passes (309 passed; the 1 pre-existing failure is an unrelated missing optional `e2b` dependency in the test env).

## Backwards Compatibility
Not breaking. Decorated async methods remain awaitable with the same return value/exception behavior from the caller's perspective; only the accuracy of the telemetry data captured internally changes.

## Related Issues
Closes #2091

---
## Summary by cubic
Fixes the `telemetry()` decorator to await async methods so timing and errors reflect real execution. Also records cancellations (`CancelledError`) as failures while still propagating them; sync methods are unchanged.

- **Bug Fixes**
  - Use `asyncio.iscoroutinefunction` to return an async wrapper that `await`s the method, capturing real runtime, exceptions, and cancellations (treat `CancelledError` as failure and re-raise).

- **Refactors**
  - Extracted event capture into a shared `_capture` helper used by both async and sync wrappers.

<sup>Written for commit d72d02f2ca55602b794996f73c993ad754325169. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2096?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/telemetry/telemetry.py
+++ b/libraries/python/mcp_use/telemetry/telemetry.py
@@ -1,3 +1,4 @@
+import asyncio
 import logging
 import os
 import platform
@@ -53,6 +54,54 @@ def telemetry(event_name: str, additional_properties: dict[str, Any] | None = No
     """Decorator to automatically track feature usage"""
 
     def decorator(func: Callable) -> Callable:
+        def _capture(self, success: bool, error_type: str | None, start_time: float) -> None:
+            execution_time_ms = int((time.time() - start_time) * 1000)
+
+            telemetry_client = None
+            if hasattr(self, "telemetry"):
+                telemetry_client = self.telemetry
+            elif hasattr(self, "_telemetry"):
+                telemetry_client = self._telemetry
+            else:
+                telemetry_client = Telemetry()
+
+            if telemetry_client:
+                telemetry_client.capture(
+                    event=GenericTelemetryEvent(
+                        EVENT_NAME=event_name,
+                        **{
+                            **(additional_properties or {}),
+                            "success": success,
+                            "execution_time_ms": execution_time_ms,
+                            "error_type": error_type,
+                        },
+                    )
+                )
+
+        if asyncio.iscoroutinefunction(func):
+
+            @wraps(func)
+            async def async_wrapper(self, *args, **kwargs):
+                # Check if telemetry is disabled on this instance
+                record_telemetry = getattr(self, "_record_telemetry", None)
+                if record_telemetry is not None and not record_telemetry:
+                    return await func(self, *args, **kwargs)
+
+                start_time = time.time()
+                success = True
+                error_type = None
+
+                try:
+                    return await func(self, *args, **kwargs)
+                except Exception as e:
+                    success = False
+                    error_type = type(e).__name__
+                    raise
+                finally:
+                    _capture(self, success, error_type, start_time)
+
+            return async_wrapper
+
         @wraps(func)
         def wrapper(self, *args, **kwargs):
             # Check if telemetry is disabled on this instance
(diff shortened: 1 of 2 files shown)


Adityaj0:
Good catch — fixed in d72d02f2. `asyncio.CancelledError` is a `BaseException`, not an `Exception`, so it now needs to be listed explicitly to avoid recording a cancelled call as a success. Kept the unconditional `raise` so cancellation still propagates normally to the caller. Added a regression test that cancels an in-flight decorated coroutine and asserts both that cancellation propagates and that the captured event has `success=False`, `error_type="CancelledError"`.

Adityaj0:
@tonxxd can you review this?

khandrew1:

PR #2100, opened 2026-08-03
fix(cli): say whether `start` mounted the inspector

### LINKED ISSUE #2084: v2 beta: inspector 404s under mcp-use start with no hint that it is dev-only
`mcp-use dev` advertises the inspector prominently:
```
➜ Inspector:     http://localhost:3000/mcp/inspector
```
Under `mcp-use start` the MCP endpoint serves fine but `/mcp/inspector` returns 404 with no explanation.

If dev-only is intended (which seems right), a one-line note in the `start` banner such as "inspector is available in dev only" would prevent people assuming the production build is broken.

Also minor: `engines.node` is pinned to `>=22.22.2`, an oddly specific patch level that blocks 22.20 for no obvious reason.

ayaangazali:
## Changes

`mcp-use start` now says what happened to the inspector instead of leaving a bare `404`:

```
$ mcp-use start
mcp-use server running at http://localhost:3000/mcp
mcp-use inspector not mounted (dev only by default); pass --with-inspector to serve it here

$ mcp-use start --with-inspector
mcp-use server running at http://localhost:3000/mcp
mcp-use inspector at http://localhost:3000/mcp/inspector
```

Relates to #2084. Deliberately not a closing reference: see below.

## Implementation Details

1. `startCommand` in `packages/server/src/bin/main.ts` prints one line after the running banner: the mounted URL when `--with-inspector` was passed, otherwise the opt-in hint.
2. The hint says **dev only by default**, not dev only. Digging into it, the inspector is not actually unavailable in production — `startWithInspector` (`src/bin/start.ts`) mounts it on the same listener at `${basePath}/inspector` when `--with-inspector` is passed, with `devMode: false` and `oauthProxyAllowLoopback: false`. Telling people it is dev-only would have been the wrong fix; they would have stopped looking for a flag that exists.
3. The mounted URL is derived from the URL the server's own `listen()` reported, with a trailing slash trimmed, so it matches the route `isInspectorPath` actually serves (`${basePath}/inspector`) rather than assuming `/mcp`.
4. `specs/CLI_SPEC.md` § `mcp-use start` item 4 updated in the same change, per the package's spec-is-contract rule.

## On the second half of the issue (`engines.node`)

> `engines.node` is pinned to `>=22.22.2`, an oddly specific patch level that blocks 22.20 for no obvious reason.

This one is deliberate, so I did not change it. `.changeset/node22-security-patch-floor.md` records the reason:

> Raise the Node.js engine floor to `>=22.22.2` (post–March 2026 security release) and pin CI to Node 22.23.1 so trusted npm 12 publishing works.

So `22.20` is excluded on purpose — it is below the security release, not an accidental patch pin. Flagging it here rather than silently ignoring half the issue; if the floor should be documented somewhere more discoverable than a changeset, that is a separate and easy PR.

## Example Usage (Before)

```
$ mcp-use start
mcp-use server running at http://localhost:3000/mcp
$ curl -s -o /dev/null -w '%{http_code}\n' http://localhost:3000/mcp/inspector
404
```

## Example Usage (After)

```
$ mcp-use start
mcp-use server running at http://localhost:3000/mcp
mcp-use inspector not mounted (dev only by default); pass --with-inspector to serve it here
```

## Documentation Updates

- `libraries/typescript/packages/server/specs/CLI_SPEC.md` — the `start` Inspector clause now states both messages and why the default path announces itself.
- Changeset added (`start-inspector-hint.md`, patch on `mcp-use`).

## Testing

Two tests added to `packages/server/tests/bin-start.test.ts`, both driving the real `main(["start", ...])` path against the existing on-disk fixture projects:

- `explains that the inspector is opt-in under start` — plain `start` logs the hint.
- `prints the inspector URL when start mounts it` — `--with-inspector` against `INSPECTOR_ENTRY` (basePath `/api/mcp`) logs `http://localhost:4567/api/mcp/inspector` and does *not* log the hint. This also pins point 3 above: the URL follows the server's own basePath rather than a hardcoded `/mcp`.

A small `captureSignalListeners` helper was added so these two can call `main` and then drop the `SIGINT`/`SIGTERM` handlers it registers, without routing through `process.exit` like the tunnel test does.

Verified red-without-fix: stashing only `src/bin/main.ts` makes both new tests fail, restoring it makes them pass.

```
$ pnpm vitest run tests/bin-start.test.ts tests/commands
 Test Files  11 passed (11)
      Tests  167 passed (167)
```

`pnpm typecheck` on the package is clean and `pnpm lint` passes at the workspace root.

## Backwards Compatibility

One additional stdout line on `mcp-use start`. Exit codes, routes, and the tunnel output are unchanged, and `start` has no `--json` mode, so nothing machine-readable is affected.

## Related Issues

Relates to #2084, but does not close it.

The original ask there was a hint on plain `mcp-use start`, so that a bare `404` at `${basePath}/inspector` would not read as a broken production build. That line was dropped on review as noise, which is the maintainer's call, so the reporter's actual problem is still open. What is left here is the `--with-inspector` URL. #2084 should stay open for a docs line instead.

---
## Summary by cubic
Update `mcp-use start` to print the Inspector URL only when `--with-inspector` is used; plain `start` stays silent.

- **Bug Fixes**
  - With `--with-inspector`, prints "mcp-use inspector at <url>" using the server’s reported URL and base path.
  - Plain `start` does not mention the Inspector. No route or exit-code changes. Closes #2084.

<sup>Written for commit 64f97087bcf785939f4e1128885984a8e72793fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2100?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ayaangazali:
Retargeted from `beta` to `main` now that v2 has shipped. `beta` is 0 commits ahead of `main` and fully contained in it, so it is no longer the branch that receives this.

That surfaced one conflict, now resolved. `CLI_SPEC.md` item 5 was rewritten on main by #2057 (hosted WebSocket tunnel relay) while this branch edits item 4. Merged main and kept both: main's tunnel wording verbatim, this branch's inspector startup message.

Re-verified after the merge rather than assuming:

* the fix is still needed on current `main` (no inspector line in `src/bin/main.ts`)
* `pnpm vitest run tests/bin-start.test.ts` -> 39 passed
* `ci-preflight` exit 0

Back to `MERGEABLE` with the same 4 file diff.

ayaangazali:
The `Generate and validate API documentation` red here is pre-existing and not caused by this branch. It fails in `Set up pnpm`, before any docs generation runs:

```
Set up pnpm  Error: No pnpm version is specified.
Please specify it by one of the following ways:
  - in the GitHub Action config with the key "version"
  - in the package.json with the key "packageManager"
```

`docs.yml` uses `pnpm/action-setup@v4` with no `version:`, so the action falls back to reading `packageManager` from `package_json_file: package.json`, meaning the repository root. There is no root `package.json` in this repo, so the step cannot succeed on any branch.

Same job, same error, on `main` itself:

```
$ gh run view 30821461990 --json headBranch,createdAt,jobs
branch: main
at:     2026-08-03T14:10:51Z
jobs:   [{ name: "Generate and validate API documentation", c: "failure" }]

$ gh run view 30821461990 --log-failed | grep "No pnpm version is specified"
Generate and validate API documentation  Error: No pnpm version is specified.
```

It also fails the same way on maintainer branches (`fix/npm-readme-assets`, `agent/npm-readme-committed-hotfix`). So it is repo-wide, not diff-specific, which fits a branch that only edits markdown.

I have not pushed anything speculative here. The actual fix belongs in `docs.yml`, where every other `pnpm/action-setup@v4` in the repo already pins `version: 11.13.1` (nine of them in `ci.yml`) and this one does not. Sending that as its own PR rather than burying a workflow change in an unrelated branch.

ayaangazali:
Second red here, `Run Conformance Tests`, is also not from this branch. It is not a test failure at all: the job was **cancelled** by its own timeout, and `gh pr checks` renders a cancellation as `fail`.

`conformance.yml` caps it:

```
.github/workflows/conformance.yml:99:    name: Run Conformance Tests
.github/workflows/conformance.yml:103:    timeout-minutes: 20
```

And it hits exactly that on every branch, including `main`:

```
this PR   run 30831073933  Run Conformance Tests  cancelled  20m03s
main      run 30821461867  14:11:12 -> 14:31:15   cancelled  20m03s
main      run 30815718818  12:57:25 -> 13:17:27   cancelled  20m02s
```

The four most recent `conformance.yml` runs on `main` are all `cancelled` the same way. The single failing step is `Run conformance tests`, and no assertion output is reached before the cut-off.

So both reds on this PR are repo-wide: this one, and the `pnpm` setup failure in `docs.yml` I described above (fixed separately in #2116). This branch adds two `console.log` lines to the `start` banner and a test, which cannot affect either.

Not pushing anything speculative for this one. Whether the right answer is raising `timeout-minutes` or making the suite finish inside 20 minutes is a call about the suite rather than a config typo, so I would rather flag it than guess. Happy to dig into what is eating the 20 minutes if that is useful.

libraries/typescript/packages/server/src/bin/main.ts
```diff
@@ -285,6 +285,16 @@ async function startCommand(args: ParsedArgs): Promise<number> {
   }
 
   console.log(`mcp-use server running at ${started.url}`);
+  if (args.inspector === true) {
+    console.log(
+      `mcp-use inspector at ${started.url.replace(/\/$/, "")}/inspector`
+    );
+  } else {
+    console.log(
```
khandrew1: we don't really need this message that the inspector is not mounted. we can assume most people running `start` probably wouldn't want the inspector so i think this just adds noise

khandrew1 (changes requested):
Thanks for the PR! Requested a small change

libraries/typescript/packages/server/src/bin/main.ts
ayaangazali: Done, dropped in 5f99f4f4. Plain `mcp-use start` is silent about the inspector again, and only `--with-inspector` prints its URL.

Also updated `CLI_SPEC.md` and the changeset to match, and flipped the test to pin the silence rather than the message, so nobody re-adds the line without a test failing.

ayaangazali:
Pushed the requested change in 5f99f4f4. Plain `mcp-use start` no longer mentions the inspector at all; only `--with-inspector` prints the mounted URL. Spec, changeset and test updated to match, with the test now pinning the silence.

One consequence worth naming: that line was the actual ask in #2084, where the reporter hit a bare `404` at `/mcp/inspector` under `start` and could not tell whether the production build was broken. With the line gone, this PR no longer addresses that half, so I would leave #2084 open rather than have it auto-close here.

If you would rather that be documented than printed, the natural home is the `start` section of the CLI docs, saying the inspector is opt-in via `--with-inspector`. Small docs PR, happy to send it. Not doing it unprompted since you have a clear view on how much this deserves.

`pnpm vitest run tests/bin-start.test.ts` is 39 passed. The two remaining reds are the repo-wide ones I documented above, `docs.yml` pnpm setup (fixed in #2116) and the conformance job hitting its own 20 minute timeout on main too.

khandrew1 (approved):

ayaangazali:
My mistake, and it is fixed. I briefly force-pushed over your rebase at 17:32 and I have restored `64f97087` exactly as you left it. Base is `canary`, head is `64f97087`, 4 files, mergeable.

What happened: I saw an unexpected force-push on the branch along with a `version packages (canary)` commit and a 379 line deletion of `packages/server/README.md`, read it as my branch having been clobbered by a stray release process, and reverted to my own commit. I should have checked the PR timeline first, which says plainly that you changed the base to `canary` at 17:17 and rebased at 17:26. I had the base change sitting in the API response I was already reading and did not look at it.

Nothing was lost either way, I had backed up your commit before touching anything and that is what I restored from. Sorry for the noise on your notifications.

Retargeting to `canary` makes sense to me given v2 just went stable on main. I will target `canary` for CLI changes from here rather than assuming main, and I will leave the branch alone unless you ask for another change.

khandrew1 (approved):

PR #2127, opened 2026-08-03
fix(agent): count Anthropic cache tokens and stop message_delta erasing usage

arthi-arumugam-git:
Two token counting bugs on the Anthropic path. Both are silent: the number comes out wrong and nothing raises.

### 1. `totalTokens` leaves out the cache counters

Anthropic reports `input_tokens` net of the prompt cache, and bills `cache_read_input_tokens` and `cache_creation_input_tokens` on top of it. It never sends `total_tokens`, so the computed branch in `tokenUsageFromRecord` is always the one that runs.

```
input_tokens 3  +  cache_read 20,000  +  cache_creation 1,500  +  output 120
main reports totalTokens: 123      actual billed: 21,623
```

`cache_creation_input_tokens` was not read anywhere in the repository, so cache writes were invisible, and Anthropic bills those above the base rate.

OpenAI is the opposite shape: its `cached_tokens` sits *inside* `prompt_tokens`, so adding it would double count. Only the Anthropic-shaped keys are summed, and there is a test pinning the OpenAI case at 14 so this cannot regress into a double count.

### 2. `message_delta` erased the counters from `message_start`

`tokenUsageFromRecord` returns every key and sets the ones it did not find to `undefined`. Anthropic's streaming `message_delta` carries `output_tokens` only, so:

```js
usage = { ...usage, ...deltaUsage }
```

overwrote `inputTokens` and `cachedInputTokens` with `undefined`.

```
after message_start : {"inputTokens":3,"outputTokens":1,"totalTokens":4,"cachedInputTokens":20000}
after message_delta : {"outputTokens":500}
```

After any streamed call the entire input side of the accounting was gone.

`mergeTokenUsage` merges only defined values. It deliberately does not recompute `totalTokens`: whether a cache counter is additive is a provider question, and that information is gone once usage has been normalised, so the caller owns the total. `streamChat` in the Anthropic provider does that sum itself, where the semantics are known.

### Tests

Added to `src/llm/__tests__/usage.test.ts`. **Three of them fail on `main`**, the fourth is the OpenAI regression guard and passes on both. The two existing assertions in that file still pass unchanged.

Happy to split this into two PRs if you would rather review them separately.

---
## Summary by cubic
Fixes Anthropic token accounting across `agent` and `inspector`. Totals now include cache read/write tokens, streaming keeps prior counters, and the inspector aggregates cache writes while avoiding OpenAI double-counts.

- **Bug Fixes**
  - Count Anthropic `cache_read_input_tokens` and `cache_creation_input_tokens` in totals (and trace); add `cacheCreationInputTokens` to `TokenUsage`; OpenAI unaffected.
  - Streaming: use `mergeTokenUsage` so `message_delta` doesn’t erase inputs; recompute the billed total once at `message_stop` to avoid drift.
  - `inspector`: aggregate `cacheCreationInputTokens` across events and the LLM span; compute the fallback total by adding only Anthropic-shaped cache counters, not OpenAI’s `cached_tokens`; surface `cachedInputTokens`, `cacheCreationInputTokens`, and `reasoningTokens`.

- **Refactors**
  - Prettier formatting in `inspector` trace to satisfy lint; no behavior change.

<sup>Written for commit db1daa19f1e65dd8d84bcf0f1132dc2c18e5ad7c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2127?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
+++ b/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
@@ -86,7 +86,10 @@ export function inspectorTokenUsageFromUnknown(
   // Anthropic reports these alongside input_tokens and bills all of them, so a total of
   // input + output understates what the call cost. OpenAI's cached_tokens sits inside
   // prompt_tokens, so only the Anthropic-shaped keys are added here.
-  const cachedInputTokens = number("cachedInputTokens", "cache_read_input_tokens");
+  const cachedInputTokens = number(
+    "cachedInputTokens",
+    "cache_read_input_tokens"
+  );
   const cacheCreationInputTokens = number(
     "cacheCreationInputTokens",
     "cache_creation_input_tokens"
--- a/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
+++ b/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
@@ -95,8 +95,14 @@ export function inspectorTokenUsageFromUnknown(
     "cache_creation_input_tokens"
   );
   const reasoningTokens = number("reasoningTokens", "thoughtsTokenCount");
+  // Only the Anthropic-shaped keys sit OUTSIDE inputTokens and must be added back into the
+  // total. cachedInputTokens is provider-neutral, and on OpenAI those tokens are already
+  // inside inputTokens, so adding it here would double count. It is still parsed above for
+  // observability; the total is computed from cache_read_input_tokens only, mirroring the
+  // provider-safe behaviour in agent/src/llm/usage.ts.
+  const cacheReadOutsideInput = number("cache_read_input_tokens");
   const uncountedCache =
-    (cachedInputTokens ?? 0) + (cacheCreationInputTokens ?? 0);
+    (cacheReadOutsideInput ?? 0) + (cacheCreationInputTokens ?? 0);
   const totalTokens =
     number("totalTokens", "total_tokens") ??
     (inputTokens !== undefined && outputTokens !== undefined
@@ -197,6 +203,10 @@ function addUsage(
     outputTokens: sum(current?.outputTokens, next.outputTokens),
     totalTokens: sum(current?.totalTokens, next.totalTokens),
     cachedInputTokens: sum(current?.cachedInputTokens, next.cachedInputTokens),
+    cacheCreationInputTokens: sum(
+      current?.cacheCreationInputTokens,
+      next.cacheCreationInputTokens
+    ),
     reasoningTokens: sum(current?.reasoningTokens, next.reasoningTokens),
   };
 }
--- a/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
+++ b/libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
@@ -3,6 +3,7 @@ export interface InspectorTokenUsage {
   outputTokens?: number;
   totalTokens?: number;
   cachedInputTokens?: number;
+  cacheCreationInputTokens?: number;
   reasoningTokens?: number;
 }
 
@@ -82,19 +83,43 @@ export function inspectorTokenUsageFromUnknown(
     "output_tokens",
     "completionTokens"
   );
+  // Anthropic reports these alongside input_tokens and bills all of them, so a total of
+  // input + output understates what the call cost. OpenAI's cached_tokens sits inside
+  // prompt_tokens, so only the Anthropic-shaped keys are added here.
+  const cachedInputTokens = number("cachedInputTokens", "cache_read_input_tokens");
+  const cacheCreationInputTokens = number(
+    "cacheCreationInputTokens",
+    "cache_creation_input_tokens"
+  );
+  const reasoningTokens = number("reasoningTokens", "thoughtsTokenCount");
+  const uncountedCache =
+    (cachedInputTokens ?? 0) + (cacheCreationInputTokens ?? 0);
   const totalTokens =
     number("totalTokens", "total_tokens") ??
     (inputTokens !== undefined && outputTokens !== undefined
-      ? inputTokens + outputTokens
+      ? inputTokens + outputTokens + uncountedCache
       : undefined);
   if (
     inputTokens === undefined &&
     outputTokens === undefined &&
-    totalTokens === undefined
+    totalTokens === undefined &&
+    cachedInputTokens === undefined &&
+    cacheCreationInputTokens === undefined &&
+    reasoningTokens === undefined
   ) {
     return undefined;
   }
-  return { inputTokens, outputTokens, totalTokens };
+  // cachedInputTokens and reasoningTokens were declared on InspectorTokenUsage and read by
+  // the trace view, but this function never returned them, so both have always rendered as
+  // absent no matter what the provider sent.
+  return {
+    inputTokens,
+    outputTokens,
+    totalTokens,
+    cachedInputTokens,
+    cacheCreationInputTokens,
+    reasoningTokens,
+  };
 }
 
 const SECRET_REQUEST_KEYS = new Set([
--- a/libraries/typescript/.changeset/anthropic-cache-usage-agent.md
+++ b/libraries/typescript/.changeset/anthropic-cache-usage-agent.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/agent": patch
+---
+
+Count Anthropic cache tokens in usage totals, and stop `message_delta` erasing the input and cache counters captured at `message_start`. Anthropic reports `cache_read_input_tokens` and `cache_creation_input_tokens` outside `input_tokens` and bills all of them, so a total of input plus output undercounts what the call cost. OpenAI's cached tokens sit inside `prompt_tokens`, so they are left alone.
--- a/libraries/typescript/.changeset/anthropic-cache-usage-inspector.md
+++ b/libraries/typescript/.changeset/anthropic-cache-usage-inspector.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/inspector": patch
+---
+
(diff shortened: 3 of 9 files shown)


arthi-arumugam-git:
Thanks for the review. Three of the four were right and are fixed in `d9216bb`. The fourth would have introduced a bug, so it is not done and the reasoning is below.

### The end-to-end test found a third instance of the same defect

Testing `mergeTokenUsage` in isolation was the right thing to call out, and driving real SSE through `streamChat` immediately showed something the unit test could not: `message_stop` was recomputing the total as `inputTokens + outputTokens`, so the fix at `message_delta` survived only because `??` short-circuited on the value already set. That is luck rather than design, and two places knowing how Anthropic bills is exactly how this drifts back.

`message_stop` now owns the total alone and `message_delta` only merges.

The new test in `anthropic-usage-streaming.test.ts` drives `message_start` then `message_delta` then `message_stop`. On the base commit it reports:

```
- Expected  "totalTokens": 21623
+ Received  "totalTokens": undefined
```

Not a wrong number, an absent one, because the input counters were gone by the time `message_stop` ran and its fallback could not compute.

### The inspector was dropping more than the new field

Worth flagging as a separate pre-existing bug. `InspectorTokenUsage` declares `cachedInputTokens` and `reasoningTokens`, `ChatTraceView` reads them, and `inspectorTokenUsageFromUnknown` returned only `inputTokens`, `outputTokens` and `totalTokens`. Both fields have always rendered as absent regardless of what the provider sent. It also carried the same input-plus-output total. Both are fixed, and `cacheCreationInputTokens` is carried through.

### Cache creation in the merge test

Added. Cache writes are billed above the base rate, so they need the same protection across the merge that cache reads do.

### The one not done

Adding `cachedInputTokens` to the computed total inside `tokenUsageFromRecord` would undercount nothing and double count OpenAI. On Anthropic that counter sits *outside* `input_tokens`; on OpenAI `cached_tokens` sits *inside* `prompt_tokens`. Once usage has been normalised that distinction is gone, which is why only the Anthropic-shaped keys are summed and why `mergeTokenUsage` deliberately does not touch the total at all. There is a test pinning the OpenAI case at 14 so this cannot regress into a double count.

Happy to split the inspector change into its own PR if you would rather keep this one to the agent package.

arthi-arumugam-git:
Thanks for these. Going through them in order. Three of the four landed in `d9216bb`, which went up after this review ran against `3d8acce`.

**P2, the inspector dropping cache-creation detail.** Fixed. `InspectorTokenUsage` carries `cacheCreationInputTokens` now, and the trace parser in `packages/inspector/src/client/components/chat/trace.ts` reads it, adds it to the cache subtotal and passes it through, so it reaches recorded `tokenUsage` and the spans.

**P3, the merge fixture only seeding `cachedInputTokens`.** Fixed. The `mergeTokenUsage` case seeds `cache_creation_input_tokens: 1500` in `start` and asserts it survives a delta that mentions only `output_tokens`.

**P3, the tests exercising `mergeTokenUsage` in isolation rather than `streamChat`.** Agreed, and this was the right thing to ask for. `anthropic-usage-streaming.test.ts` drives `streamChat` against an SSE fixture with the real sequence: `message_start` carrying `input_tokens`, `cache_read_input_tokens` and `cache_creation_input_tokens`, then `message_delta` carrying `output_tokens` and nothing else. It asserts `totalTokens: 21623`. Before the change the input side was `undefined` by the time `message_stop` ran and the total came back 123, so the fixture fails on the old behaviour rather than passing either way.

**P2, `totalTokens` undercounting camel-case `cachedInputTokens`.** This one is deliberate and I would rather keep it. `cachedInputTokens` cannot be attributed to a provider from the key alone. Measured on the current branch:

```
{ inputTokens: 3,  outputTokens: 120, cachedInputTokens: 20000 } -> totalTokens 123
{ inputTokens: 10, outputTokens: 4,   cachedInputTokens: 3 }     -> totalTokens 14
```

The first is Anthropic-shaped and does undercount, exactly as you describe. The second is OpenAI-shaped and is already correct, because OpenAI reports `cached_tokens` inside `prompt_tokens`. Summing camel-case `cachedInputTokens` fixes the first and silently double counts the second, and the two records are indistinguishable at that point. `cache_read_input_tokens` is summed because that spelling only ever arrives from Anthropic. `cacheCreationInputTokens` is summed in both spellings because OpenAI has no equivalent counter for it to collide with.

It is also unreachable from the current call sites, which all pass a raw provider payload: Anthropic and OpenAI send snake case, Google sends `usageMetadata` keyed `promptTokenCount` and `candidatesTokenCount`. The camel-case spelling only shows up if an already-normalized `TokenUsage` is parsed a second time, and for an OpenAI-derived one the cache is inside `inputTokens` by then.

Where a provider genuinely needs its cache counted, the provider owns its own total rather than leaning on the generic parser, which is what `anthropic.ts` does when it emits the `usage` event.

Happy to add a test pinning that camel-case OpenAI record at 14 if it is better to have that reasoning enforced than written in a comment.

arthi-arumugam-git:
A note while this is waiting for review, in case it helps size it.

The same shape is in four other codebases that share no code with this one: Pipecat, LiveKit Agents, Haystack and LlamaIndex. The first three have merged the fix. Anthropic reporting `inputTokens` net of the cache catches everyone once, and it never raises, so it survives review on the strength of the number looking plausible.

Since it repeats, it is checkable, and I packaged the check: https://github.com/arthi-arumugam-git/cachecheck. `uvx cachecheck .`, no dependencies, exit code 1 on a finding, so it works as a CI gate.

Correcting myself on one point: run against this branch's base it flags `providers/anthropic.ts` only, not all three files this PR touches. The other two carry no provider name, and the checker now requires one so that it stops reporting every OpenAI file in a repository. That is a known miss and it is written up in the README rather than papered over. Happy to rebase if it has gone stale, or to split it if the `message_delta` half is the part holding it up.

libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
```diff
@@ -3,6 +3,7 @@ export interface InspectorTokenUsage {
   outputTokens?: number;
   totalTokens?: number;
   cachedInputTokens?: number;
+  cacheCreationInputTokens?: number;
```
khandrew1, line 6: `cacheCreationInputTokens` is parsed and placed on individual usage events, but `addUsage()` does not aggregate the new field. As a result, it disappears from `state.usage`, the LLM span’s usage, and the aggregate `tokenUsage` written to the raw-chat payload.

Please add the following field to the object returned by `addUsage()`:

```ts
cacheCreationInputTokens: sum(
  current?.cacheCreationInputTokens,
  next.cacheCreationInputTokens
),
```

It would also be useful to extend `trace.test.ts` to assert that this counter survives in both aggregate state and LLM span usage.

libraries/typescript/packages/inspector/src/client/components/chat/trace.ts
```diff
@@ -82,19 +83,46 @@ export function inspectorTokenUsageFromUnknown(
...
+  const cachedInputTokens = number(
+    "cachedInputTokens",
+    "cache_read_input_tokens"
+  );
+  const cacheCreationInputTokens = number(
+    "cacheCreationInputTokens",
+    "cache_creation_input_tokens"
+  );
+  const reasoningTokens = number("reasoningTokens", "thoughtsTokenCount");
+  const uncountedCache =
```
khandrew1, line 104: This fallback can double-count OpenAI cache reads because `cachedInputTokens` is provider-neutral. For OpenAI, cached tokens are already included in `inputTokens`, so a normalized record such as `{ inputTokens: 10, outputTokens: 4, cachedInputTokens: 3 }` currently produces 17 instead of 14 when `totalTokens` is absent.

Please keep parsing `cachedInputTokens` for observability, but calculate `uncountedCache` from the Anthropic-specific `cache_read_input_tokens` key instead. This would mirror the provider-safe behavior in `agent/src/llm/usage.ts`.

A regression test for the normalized OpenAI example above would pin the distinction.

khandrew1 (changes requested):
Thanks for the fix—the Anthropic accounting and streaming changes look good. Before merging, please address the two inspector follow-ups below and add changesets covering both `@mcp-use/agent` and `@mcp-use/inspector`.

arthi-arumugam-git:
Thanks, both inspector follow-ups are in as of `db1daa1`.

**`addUsage` aggregation.** `cacheCreationInputTokens` is now summed alongside the other counters, so it survives into `state.usage`, the LLM span's usage, and the aggregate `tokenUsage` on the raw-chat payload. Added your suggested `sum(...)` line verbatim.

**The OpenAI double-count.** Good catch. `uncountedCache` now reads only the Anthropic-shaped `cache_read_input_tokens`, so provider-neutral `cachedInputTokens` no longer gets added back on top of a total that already includes it. That mirrors the `cacheReadOutsideInput` path in `agent/src/llm/usage.ts` you pointed at. `cachedInputTokens` is still parsed and returned for observability.

Two tests in `trace.test.ts` pin both: your normalized `{ inputTokens: 10, outputTokens: 4, cachedInputTokens: 3 }` example now totals 14 rather than 17, and a second test asserts `cacheCreationInputTokens` survives aggregation and lands on the LLM span. Changesets added for `@mcp-use/agent` and `@mcp-use/inspector`.

arthi-arumugam-git:
@khandrew1 gentle ping on this one. Both of your review points went in on 11 August in db1daa1: `cacheCreationInputTokens` is aggregated in `addUsage` with your suggested line, and the OpenAI double-count is gone because `uncountedCache` now reads only the Anthropic-shaped `cache_read_input_tokens`. The PR still shows changes requested since GitHub keeps that state until a re-review, and I cannot request one from outside the repo.

Happy to rebase if it has drifted. If you would rather I split the inspector fix from the accounting fix into two PRs, say the word and I will do that instead.

khandrew1 (approved):
Thanks for the PR!

PR #2114, opened 2026-08-03
docs: fix TypeScript README links into the removed mcp-use package

ayaangazali:
## Changes

The v2 merge (`06ff3075 release: merge v2 beta into main`) deleted `libraries/typescript/packages/mcp-use`, but `libraries/typescript/README.md` still points at it in four places. The logo at the very top of the page is currently broken, and the documentation link at the bottom is dead.

```
$ git ls-files libraries/typescript/packages/mcp-use | wc -l
0
```

## Implementation Details

1. **Logo, lines 3 to 5.** Was `./packages/mcp-use/static/logo_white.svg` and `logo_black.svg` in the `<picture>` sources and the `` fallback. Now the absolute `https://raw.githubusercontent.com/mcp-use/mcp-use/main/static/logo_{white,black}.svg`.

   This is the convention the rest of the repo already uses, not a new one. Every sibling package README does it this way:

   ```
   $ git grep -l "raw.githubusercontent.com/mcp-use/mcp-use/main/static/logo_white.svg" -- '*.md'
   libraries/typescript/packages/agent/README.md
   libraries/typescript/packages/client/README.md
   libraries/typescript/packages/create-mcp-use-app/README.md
   libraries/typescript/packages/inspector/README.md
   ```

   `libraries/typescript/README.md` was the only file left on the relative path:

   ```
   $ git grep -l "packages/mcp-use/static/logo" -- '*.md'
   libraries/typescript/README.md
   ```

   Both targets resolve, and the files are tracked at the repo root (`static/logo_white.svg`, `static/logo_black.svg`):

   ```
   $ curl -s -o /dev/null -w "%{http_code}\n" https://raw.githubusercontent.com/mcp-use/mcp-use/main/static/logo_white.svg
   200
   $ curl -s -o /dev/null -w "%{http_code}\n" https://raw.githubusercontent.com/mcp-use/mcp-use/main/static/logo_black.svg
   200
   ```

   A relative `../../static/logo_white.svg` would also render on GitHub, but I matched the siblings instead so the whole TypeScript tree is consistent.

2. **Docs link, line 319.** `[**Full mcp-use Documentation →**](./packages/mcp-use)` now points at `./packages/server`, which is where the `mcp-use` package lives in v2. Same target the other READMEs already use for it:

   ```
   libraries/typescript/packages/create-mcp-use-app/README.md:30: ... /tree/main/libraries/typescript/packages/server) | MCP framework and CLI
[...]

PR #2101, opened 2026-08-03
fix(ci): skip v2 migration checklists in the link checker

ayaangazali:
## Changes

Adds the two v2 migration checklists to `exclude_path` in `lychee.toml`. That clears **39 of the 47 errors** in this morning's Link Checker Report #2098.

## Why they fail

Both files link to Mintlify routes as root-relative paths:

```markdown
- [x] [Tools](/v2/typescript/client/tools)
```

Mintlify serves that from `docs/v2/typescript/client/tools.mdx`. lychee cannot follow it either way, and I checked both directions on `lychee 0.24.2`:

```
$ lychee --offline docs/v2/typescript/MCP_CLIENT_MIGRATION_CHECKLIST.md
[ERROR] error: (at 44:7) | Cannot resolve root-relative link '/v2/typescript/client/tools':
        To resolve root-relative links in local files, provide a root dir

$ lychee --offline --root-dir "$PWD/docs" docs/v2/typescript/MCP_CLIENT_MIGRATION_CHECKLIST.md
[ERROR] file://.../docs/v2/typescript/client/tools (at 44:7) | File not found.
```

So `--root-dir` is not the fix. It resolves the path and then looks for an extension-less file, because lychee has no notion of Mintlify's extension-less routing.

## Nothing checkable is skipped

The links are not broken. I resolved all 39 targets against the tree and every one exists:

```
$ grep -o '(/v2/[^)]*)' docs/v2/typescript/MCP_*_MIGRATION_CHECKLIST.md | tr -d '()' | sort -u \
    | while read p; do [ -f "docs$p.mdx" ] && echo "OK $p" || echo "MISS $p"; done | sort | uniq -c
  39 OK
   0 MISS
```

And both files contain zero external links (`grep -cE '\(https?://'` returns 0 in each), so `exclude_path` costs no coverage at all. Same reasoning as the `CHANGELOG.md` entries already in that key.

## Verification

Offline run over the 153 tracked `.md` files, before and after:

```
before:  1309 Total   523 Unique   618 OK   39 Errors   652 Excluded
after:   1270 Total   523 Unique   618 OK    0 Errors   652 Excluded
```

All 39 errors gone, `OK` unchanged at 618, `Excluded` unchanged at 652. The offline pass is fully green.

## One thing I did not guess at

[...]

PR #2132, opened 2026-08-04
docs: repoint examples README at the v2 example directories

ayaangazali:
## Changes

The v2 merge deleted `libraries/typescript/packages/mcp-use`, and `examples/README.md` still links into it. Of its 86 relative links, 26 are dead. This fixes the 17 that have an exact home in the v2 tree and deliberately leaves the other 9 alone.

Follow-up to #2114, where I flagged this file and said I would rather ask than guess on the ambiguous half. This is the half that needs no decision.

## The mapping

Three mechanical moves, nothing renamed by judgement:

| From | To |
| --- | --- |
| `packages/mcp-use/examples/client/…` | `packages/client/examples/…` |
| `packages/mcp-use/examples/server/features/<x>/` | `packages/server/examples/<x>/` |
| `packages/mcp-use/examples/server/oauth/<x>/` | `packages/server/examples/auth/<x>/` |

Every target was resolved against the tracked tree before editing, so none of these are guesses. Two are a rename rather than a pure move, and I want to name them explicitly rather than bury them:

* `server/basic/simple/` to `server/examples/basic/` — v2 dropped the `simple` nesting.
* `server/features/` (the parent bullet, "Advanced features") to `server/examples/` — v2 has no `features/` grouping, the examples sit at the root.

Veto either of those on its own if you would rather they read differently.

The file already contained correct v2 paths for Better Auth, the views examples and the agent examples, so this brings the rest into line with what was already there.

## What I left broken, and why

These 9 have no v2 counterpart, so fixing them means either deleting the entry or pointing it somewhere approximate. That is a content call, not a repoint:

```
client/node/full-features-example.ts
server/features/everything/
server/features/completion/
server/features/streaming-props/
server/features/express-middleware/
server/features/session-management/
server/features/client-info/
server/features/dns-rebinding/
server/deployment/
```

Two I specifically did not want to guess at:

* `completion` looks like it maps to `resource-template-configuration`… except v2's example is `resource-template-completion`, which is completion for resource template arguments, while this bullet says "Autocomplete for prompt args". Different feature, so I left it.
* `dns-rebinding` might be covered by v2's `security` example, but I have not verified that it demonstrates the same protection, so I did not point at it on a hunch.

Tell me whether you want the survivors repointed to the nearest v2 example or the entries dropped, and I will send that as a second small PR.

## Verification

Resolved every relative link in the file against the tracked tree, before and after:

```
before:  total ../ links: 86   broken: 26
after:   total ../ links: 86   broken: 9
```

The 9 remaining are exactly the list above, so nothing was missed and nothing else moved. `ci-preflight` exit 0.

## Backwards Compatibility

Documentation only. No source, build, or published artifact changes.

## Related

Follow-up to #2114.

---
## Summary by cubic
Fix broken links in `examples/README.md` after the v2 restructure. Repointed 17 links to `packages/client/examples` and `packages/server/examples`; left 9 without v2 equivalents unchanged.

- **Bug Fixes**
  - Mapped client links from `packages/mcp-use/examples/client/...` to `packages/client/examples/...`.
  - Mapped server links by dropping `features/` to `packages/server/examples/...`; moved `basic/simple/` to `examples/basic/`; renamed `oauth/` to `auth/`.
  - Verified updated links resolve; 9 ambiguous entries remain for follow-up.

<sup>Written for commit 6b524351985eeff51af29dbb2bc9063761d653e7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -60,33 +60,33 @@ When you clone this repository locally, you'll find `python/` and `typescript/` 
 
 ### Client Examples
 - **[Node HTTP Client](../libraries/typescript/packages/mcp-use/examples/client/node/full-features-example.ts)** - Node/HTTP client with tool calls, sampling, elicitation, notifications
-- **[CommonJS Example](../libraries/typescript/packages/mcp-use/examples/client/browser/commonjs/commonjs_example.cjs)** - CommonJS usage
-- **[CLI Examples](../libraries/typescript/packages/mcp-use/examples/client/cli/)** - Command-line interface examples
-- **[React Integration](../libraries/typescript/packages/mcp-use/examples/client/browser/react/)** - React client examples
-- **[Notifications Client](../libraries/typescript/packages/mcp-use/examples/client/node/communication/notification-client.ts)** - Notification handling
-- **[Sampling Client](../libraries/typescript/packages/mcp-use/examples/client/node/communication/sampling-client.ts)** - Sampling configuration
+- **[CommonJS Example](../libraries/typescript/packages/client/examples/browser/commonjs/commonjs_example.cjs)** - CommonJS usage
+- **[CLI Examples](../libraries/typescript/packages/client/examples/cli/)** - Command-line interface examples
+- **[React Integration](../libraries/typescript/packages/client/examples/browser/react/)** - React client examples
+- **[Notifications Client](../libraries/typescript/packages/client/examples/node/communication/notification-client.ts)** - Notification handling
+- **[Sampling Client](../libraries/typescript/packages/client/examples/node/communication/sampling-client.ts)** - Sampling configuration
 
 ### Server Examples
-- **[Basic Server](../libraries/typescript/packages/mcp-use/examples/server/basic/simple/)** - Simple server implementation
-- **[Server Features](../libraries/typescript/packages/mcp-use/examples/server/features/)** - Advanced features
+- **[Basic Server](../libraries/typescript/packages/server/examples/basic/)** - Simple server implementation
+- **[Server Features](../libraries/typescript/packages/server/examples/)** - Advanced features
   - [Everything](../libraries/typescript/packages/mcp-use/examples/server/features/everything/) - All MCP primitives in one server
-  - [Conformance](../libraries/typescript/packages/mcp-use/examples/server/features/conformance/) - MCP conformance test server
-  - [Elicitation](../libraries/typescript/packages/mcp-use/examples/server/features/elicitation/) - Form and URL elicitation
-  - [Sampling](../libraries/typescript/packages/mcp-use/examples/server/features/sampling/) - Server-initiated LLM sampling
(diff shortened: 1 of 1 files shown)


ayaangazali:
Related finding from the same root cause, which I am not going to guess at.

23 files under `docs/typescript/` link to source inside the deleted package, as `github:` frontmatter and inline "View the source code" anchors. 12 distinct URLs, all 404 now:

```
$ curl -s -o /dev/null -w "%{http_code}\n" \
  "https://github.com/mcp-use/mcp-use/blob/main/libraries/typescript/packages/mcp-use/src/server/mcp-server.ts"
404
```

The tempting fix is to pin them at the last v1 tag, since these are v1 docs and v1 source no longer exists on `main`. That only half works. Of the 12, 8 resolve at `v1.5.0` and 4 do not:

```
404  libraries/typescript/packages/mcp-use/examples/server/features/middleware/src/server.ts
404  libraries/typescript/packages/mcp-use/src/react/useCallTool.ts
404  libraries/typescript/packages/mcp-use/src/server/sessions/index.ts
404  libraries/typescript/packages/mcp-use/src/server/widgets/mcp-ui-adapter.ts
```

So a blanket pin would leave a third of them broken and imply a precision it does not have. Repointing them at the v2 sources instead would be wrong in a different way, since these pages document v1 APIs.

Which would you prefer:

1. pin each URL at whichever v1 tag actually contains it, accepting mixed refs
2. pin all of them at one tag and drop the four anchors that do not resolve there
3. drop the source links from the v1 pages entirely
4. leave it, if `docs/typescript/` is on its way out anyway

Any of those is a small mechanical PR once the direction is set. I did not want to pick one and quietly ship 23 files of v1 docs.

khandrew1 (approved):

khandrew1:
Thanks for reporting these, I'm currently working on a PR to clean these up

ayaangazali:
Good to know, I will leave the `docs/typescript` source links to your PR and not touch them.

While I have you: #2026 and #2101 are the last two of mine open, both one-file `lychee.toml` changes, and together they take the weekly Link Checker Report to zero. Neither has been reviewed yet.

* #2026 (9d) excludes the `stargazers` and `star-history` badge URLs. Both still fail right now, 404 and 500, and the `facebook/react` control also 404s, so the stargazers one is GitHub bot mitigation rather than anything wrong here.
* #2101 excludes the two v2 migration checklists, whose root-relative Mintlify routes lychee cannot resolve in either direction. That is currently 38 of the report's errors.

Both are green, mergeable and rebased on current `main`. Happy either way, just tell me if you would rather live with those errors in the weekly report and I will close both rather than leave them sitting.

PR #2135, opened 2026-08-04
fix(inspector): fall back to redirect auth

tonxxd:
## Summary

- Fall back to same-tab navigation when the Manufact OAuth popup is blocked.
- Persist the originating Inspector URL and return there after a same-tab callback.
- Add an `@mcp-use/inspector` patch changeset.

## Root cause

The Inspector’s Manufact sign-in path required `window.open`; embedded browser hosts such as Codex can block that popup, leaving authentication unable to proceed.

## Validation

- Inspector unit tests passed before removing the local test-only change: 11 tests passed.
- Inspector TypeScript check passed.
- ESLint passed for the changed auth implementation.
- Pre-commit checks passed.


khandrew1:
sign in used to be broken on standalone inspector (pnpm --filter inspector dev) as the cloud api was hardcoded into the package script

PR #2136, opened 2026-08-04
Add public @mcp-use/tunnel workspace package

tonxxd:
## Summary

- add a new publishable `@mcp-use/tunnel` workspace package with zero runtime dependencies
- share the existing tested WebSocket tunnel manager across the standalone `mcp-tunnel` command and `mcp-use dev/start --tunnel`
- embed tunnel support in the existing lazy CLI command chunks, so installing `mcp-use` does not add a separate tunnel runtime dependency
- register the package with workspace builds, CI, Changesets, beta publishing, package validation, TypeDoc, and Knip
- keep package growth small: the CLI tarball increases by about 3.8 KB compressed, while the `mcp-use` tarball and normal framework import graph remain unchanged

## Why

The standalone command and framework had overlapping clients for the same public relay contract. This makes the tested framework implementation the shared source while preserving both built-in and standalone usage.

## Public repository review

- squashed the PR to one sanitized commit
- scanned the complete commit with Gitleaks; no leaks were found
- removed infrastructure-specific wording, operational endpoints, and unnecessary protocol details from public documentation
- verified that added URLs are limited to public project/package links, the documented relay origin, and `.example.com` test fixtures
- kept reservation state protected with owner-only file permissions

## Validation

- `pnpm build`
- `pnpm lint`
- `pnpm format:check`
- 7 tunnel package tests
- 75 focused server/CLI integration tests
- `pnpm verify:beta-packs`
- package size and zero-runtime-dependency checks
- clean package installation assertions for the standalone package and bundled CLI support
- Changesets plans `@mcp-use/tunnel@0.2.0-canary.0`

## Existing baseline checks

- release-install validation reaches an existing duplicate transitive-dependency invariant unrelated to this PR; the check is unchanged
- release-versioning retains existing fixture failures unrelated to this package


tonxxd:
Superseded by #2175, rebuilt as one clean commit on current canary after CLI source ownership moved to @mcp-use/cli. The replacement keeps @mcp-use/tunnel as the protocol owner and bundles it into the CLI without a runtime dependency.

PR #2201, opened 2026-08-08
fix(ci): actually run the inspector unit tests

### LINKED ISSUE #2200: CI step "Run Inspector Tests" runs nothing: inspector has no `test` script
`ci.yml` runs the inspector unit tests like this:

```yaml
      - name: Run Inspector Tests
        run: pnpm --filter @mcp-use/inspector --if-present test
```

`@mcp-use/inspector` has no `test` script. Its unit entry point is `test:unit`. With `--if-present`, a missing script is a silent success, so the step passes without running anything:

```
$ pnpm --filter @mcp-use/inspector --if-present test
Already up to date
Done in 1.1s
exit 0
```

Meanwhile the tests exist and pass:

```
$ pnpm --filter @mcp-use/inspector test:unit
Test Files  37 passed (37)
Tests       159 passed (159)
Duration    3.61s
```

So 159 inspector unit tests have not been running, and the job has been reporting success the whole time. The workflow's own comment at line 34 describes the step as *"typescript/inspector: Unit tests for inspector package"*.

The sibling steps in the same job already use the right script:

```
ci.yml:703  pnpm --filter mcp-use --if-present test:unit
ci.yml:707  pnpm --filter @mcp-use/agent test:unit
ci.yml:779  pnpm --filter @mcp-use/inspector --if-present test     <- this one
```

Present on both `main` and `canary`.

Fix is `test` -> `test:unit` on that line. It goes green immediately at about 4 seconds. Sending a PR.

ayaangazali:
Closes mcp-use/mcp-use#2200.

The "Run Inspector Tests" step runs a script that does not exist:

```yaml
run: pnpm --filter @mcp-use/inspector --if-present test
```

`@mcp-use/inspector` has no `test` script; its unit entry point is `test:unit`. `--if-present` turns a missing script into a silent success, so the step passes having run nothing:

```
$ pnpm --filter @mcp-use/inspector --if-present test
Already up to date
Done in 1.1s
exit 0
```

The tests exist and pass:

```
$ pnpm --filter @mcp-use/inspector test:unit
Test Files  37 passed (37)
Tests       159 passed (159)
Duration    3.61s
```

So 159 unit tests have not been running, and the job reported success throughout. The workflow's own legend at line 34 calls this step *"Unit tests for inspector package"*.

The sibling steps in the same file already use the right script, which is what makes this look like a slip rather than intent:

```
ci.yml:703  pnpm --filter mcp-use --if-present test:unit
ci.yml:707  pnpm --filter @mcp-use/agent test:unit
```

I also dropped `--if-present` here. The script exists, and if it ever stops existing this step should fail rather than go quietly green again.

Costs about 4 seconds and goes green immediately.

`canary` has the same line, so it will want the same change when this syncs across.

---

## Summary by cubic

Fix CI to actually run `@mcp-use/inspector` unit tests by invoking `test:unit` instead of a missing `test` script, and remove `--if-present` so the step fails if it ever disappears. Aligns with sibling steps and restores execution of 159 tests (\~4s).

Written for commit 04c132f5de638cf3bb803fe36c0a758dd5a4f967. Summary will update on new commits.

Review in cubic

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -772,7 +772,7 @@ jobs:
       - name: Run Inspector Tests
         env:
           MCP_USE_ANONYMIZED_TELEMETRY: false
-        run: pnpm --filter @mcp-use/inspector --if-present test
+        run: pnpm --filter @mcp-use/inspector test:unit
 
   typescript-test-cli:
     needs: typescript-lint


ayaangazali:
Flagging a gap in my own evidence: CI has not actually exercised this change.

`typescript/inspector` was **skipped** on this PR's run:

```
$ gh run view 31267547078 --json jobs
{ "n": "typescript/inspector (node-...)", "c": "skipped" }
```

The job is path-filtered on the inspector package, and this PR only touches `.github/workflows/ci.yml`, so the job it fixes never triggers. I cannot force a run from a fork, so "green" here means the run succeeded without executing the step I changed.

I raise it because I made exactly this mistake on #2204: enabled a skipped step on the strength of a local run, and the runner then surfaced four failures, three of them caused by the job's own `MCP_USE_ANONYMIZED_TELEMETRY: false`.

What I checked to bound that risk here:

- No inspector unit test reads `process.env` at all, so the telemetry variable this job sets cannot affect them. That was the exact cause of three of #2204's failures.
- `vitest.config.ts` uses `environment: "node"`, not jsdom, so there is no browser-environment dependency to differ on the runner.
- Locally: 37 files, 159 tests, all passing in 3.6s.

So the residual risk is lower than #2204's, but it is not zero and I have not seen the step run on a runner. If you would rather be certain first, pushing a trivial change under `packages/inspector/` on this branch would trigger the job, and I am happy to do that instead of asking you to merge on my local run.

khandrew1 (approved):

Issue #2203, opened 2026-08-08
CI skips the mcp-use unit tests too, hiding a failing peer-range assertion


ayaangazali:
Same shape as #2200, on the framework package, and this one was hiding a real failure.

`ci.yml` runs:

```yaml
run: pnpm --filter mcp-use --if-present test:unit
```

`packages/server` has no `test:unit` script. It has `test` (`vitest`, watch) and `test:run` (`npm run typecheck && vitest run`). With `--if-present` the step is a silent success:

```
$ pnpm --filter mcp-use --if-present test:unit
exit 0, no output
```

Running the real suite shows why that matters:

```
$ pnpm --filter mcp-use test:run
Test Files  1 failed | 64 passed (65)
Tests       1 failed | 703 passed (704)

FAIL tests/proxy-optional-peer.test.ts > declares @mcp-use/client as an optional peer
AssertionError: expected '^2.0.1' to be '^2.0.0-alpha.0'
```

`tests/proxy-optional-peer.test.ts:8` still pins the peer at `^2.0.0-alpha.0`. `package.json` moved to `^2.0.1` when v2 went stable, and the assertion was never updated. It has been failing since, unnoticed, because the step that would have caught it runs nothing.

So there are two defects stacked: the CI step is misconfigured, and the test it should have been running is stale.

`test:run` is the right target rather than `test`, since `test` is watch mode and would hang CI. It takes about 10 seconds and includes the package typecheck.

PR coming with both halves, since wiring the step up without fixing the assertion would just turn CI red.

ayaangazali:
Update: enabling the step is not a one-liner, so I have cut #2204 back to only the stale assertion and left this issue open for the CI change.

Running the suite on the runner surfaces four more failures beyond the peer assertion:

```
tests/usage.test.ts  x3   expected "vi.fn()" to be called once, but got 0 times
tests/budgets/runtime-boundary.test.ts   expected ['node:http','node:process'] to deeply equal ['node:process','node:http']
```

The `usage.test.ts` three are environmental: the job sets `MCP_USE_ANONYMIZED_TELEMETRY: false`, and the suite tracks that variable without neutralising it, so the usage path never runs. The budgets one compares the same two module names in a different order.

Both are pre-existing and only invisible because the step never ran. Neither is mine to decide: the first is a question about whether tests should be insulated from that env var or the job should stop setting it for this suite, and the second is a brittle assertion that wants sorting or a set comparison.

[...]

PR #2204, opened 2026-08-08
fix(test): update the stale @mcp-use/client peer assertion

ayaangazali:
Refs #2203. Scope reduced, see the comment below for why.

`packages/server/tests/proxy-optional-peer.test.ts:8` pins the `@mcp-use/client` peer at `^2.0.0-alpha.0`. `package.json` moved to `^2.0.1` when v2 went stable and the assertion was never updated, so the test has been failing:

```
AssertionError: expected '^2.0.1' to be '^2.0.0-alpha.0'
```

Nobody saw it because the CI step meant to run this suite runs a script that does not exist (`pnpm --filter mcp-use --if-present test:unit`, and there is no `test:unit`), so it silently passes. That part is #2203; this PR is just the assertion.

After this, `pnpm --filter mcp-use test:run` is 704 passed / 704 locally.

## The tradeoff I took

Updated the pinned string, the smallest change that makes the assertion true and keeps the test's intent.

It will drift again at the next peer bump, which is how it broke this time. The alternative is asserting the peer is declared and `optional: true` without pinning a range, matching the test's own name, at the cost of no longer catching an accidental range change. I did not want to quietly change what the test checks, so say the word if you would prefer that.

`canary` carries the same stale assertion.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -696,7 +696,7 @@ jobs:
       - name: Run mcp-use Unit Tests
         env:
           MCP_USE_ANONYMIZED_TELEMETRY: false
-        run: pnpm --filter mcp-use --if-present test:unit
+        run: pnpm --filter mcp-use test:run
       - name: Run @mcp-use/agent Unit Tests
         env:
           MCP_USE_ANONYMIZED_TELEMETRY: false
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -696,7 +696,7 @@ jobs:
       - name: Run mcp-use Unit Tests
         env:
           MCP_USE_ANONYMIZED_TELEMETRY: false
-        run: pnpm --filter mcp-use test:run
+        run: pnpm --filter mcp-use --if-present test:unit
       - name: Run @mcp-use/agent Unit Tests
         env:
           MCP_USE_ANONYMIZED_TELEMETRY: false
--- a/libraries/typescript/packages/server/tests/proxy-optional-peer.test.ts
+++ b/libraries/typescript/packages/server/tests/proxy-optional-peer.test.ts
@@ -5,9 +5,7 @@ import { proxyClientInstallError } from "../src/mcp-proxy.js";
 
 describe("server.proxy optional peer", () => {
   it("declares @mcp-use/client as an optional peer", () => {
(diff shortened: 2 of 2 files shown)


ayaangazali:
I had this wrong and have cut the PR back.

I claimed enabling the step "goes green immediately" on the strength of a local run. It does not. On the runner it exposes four more failures:

```
FAIL tests/usage.test.ts > persists the anonymous identity for a Node filesystem runtime
FAIL tests/usage.test.ts > does not write a project identity in a Deno-style runtime
FAIL tests/usage.test.ts > uses the explicit project identity without writing runtime state
  AssertionError: expected "vi.fn()" to be called once, but got 0 times

FAIL tests/budgets/runtime-boundary.test.ts > loads only the SDK process shim and listen HTTP adapter
  AssertionError: expected [ 'node:http', 'node:process' ] to deeply equal [ 'node:process', 'node:http' ]
```

Both have causes, and neither is something I should decide:

- The three `usage.test.ts` failures come from the job's own environment. That step sets `MCP_USE_ANONYMIZED_TELEMETRY: false`, and the suite tracks that variable without neutralising it, so the usage path never fires and the mocks are never called. Either the tests should be immune to it or the job should not set it for this suite, and that is a call about how you want telemetry treated in tests.
- The budgets failure is an order-dependent comparison of the same two elements, `['node:http','node:process']` versus `['node:process','node:http']`. Module load order differs between environments.

So this PR now contains only the stale assertion fix, which is correct regardless. Enabling the step stays in #2203 until those four are sorted, because turning it on today would just hand you a red main, which is worse than the silent skip.

My mistake was verifying locally and not against the job's env block before writing the claim.

ayaangazali:
Closing this as superseded. `main` has since replaced the exact-string assertion with `toBeDefined()`:

```ts
expect(packageJson.peerDependencies["@mcp-use/client"]).toBeDefined();
```

That is the better fix. My PR pinned the literal to `^2.0.1`, which would have gone stale immediately: the field on `main` is now `"^2.0.1 || ^2.1.0"` after the prerelease exit in #2208, so the pin I proposed would already be failing again. `toBeDefined()` still tests what the case is actually about, which is that the peer is declared and optional, without re-encoding the version range.

The branch is conflicting now anyway. The remaining work from #2203, enabling the CI step that skips these tests, is unaffected and I will keep that on the issue.

PR #2232, opened 2026-08-11
fix(inspector): scope loading state by chat session

### LINKED ISSUE #2217: Starting a new Inspector chat is blocked while another message is still processing
## Description

When a message is still being processed in one Inspector chat session, the global send control remains in its loading state. This prevents starting a new chat session until the in-flight response completes.

## Steps to reproduce

1. Open the Inspector.
2. Send a message that takes a noticeable amount of time to complete.
3. While it is still processing, try to start a new chat.
4. Observe that the send button remains stuck in a loading state and the new chat cannot be started.

## Expected behavior

Each chat session should manage its own loading state. A message processing in one session should not prevent creating or using another chat session.

## Actual behavior

The loading state appears to be shared globally, so the send button stays loading and blocks new-chat creation until the original request finishes.

## Acceptance criteria

- In-flight message state is scoped to the active chat/session.
- Users can start a new chat while another session is processing.
- The new session has an enabled send control and can accept messages immediately.
- Existing behavior for the in-flight session remains unchanged.

merttcetn:
## Summary

- scope messages, loading state, attachments, traces, and streaming runtime to each Inspector chat session
- let **New Chat** activate a fresh session without clearing or cancelling the in-flight session
- keep background session updates associated with and persisted to their original chat history entry
- preserve cached in-flight state when switching between chat history entries

## Root cause

The chat hooks owned a single set of React state and runtime refs for the entire `ChatTab`. Starting a new chat only cleared those shared values, while the existing asynchronous stream and send guard continued to use them. The fresh chat therefore inherited the original session's loading state and could not send.

## Validation

- `pnpm --filter @mcp-use/inspector test:unit` (45 files, 201 tests)
- `pnpm --filter @mcp-use/inspector type-check`
- `pnpm lint:strict`
- `pnpm format:check`
- `pnpm build`
- `pnpm changeset status --since=upstream/main`

Closes #2217

---
## Summary by cubic
Scopes all Inspector chat state to a per-session id so New Chat opens a fresh session while earlier replies keep streaming. OAuth retries and chat history now resolve by (server, session) to resume the right chat after redirects and avoid duplicate rows.

- Each session owns its messages, loading, attachments, trace, and runtime in a `ChatSessionStore`. `ChatTab` uses `useChatSessions` to activate or switch sessions without cancelling in‑flight turns; selecting history restores persisted messages without overwriting a live stream.
- Persistence uses the same session id: `ChatStorageProvider.createChat(id?)` adopts it; backends that mint their own ids are tracked via `persistedChatId`. Concurrent writes coalesce per session, and empty untouched drafts are dropped when starting a new chat.
- OAuth retry stores the interrupted turn per server+session and records which session initiated the redirect. Use `readPendingChatTurn(serverId, sessionId)`, `readPendingChatTurnForServer(serverId)`, and `clearPendingChatTurn(serverId, sessionId)`. After reload, the interrupted session reopens and replays its turn. Authentication waits briefly for a backend-minted chat id but never blocks on stalled history.
- Controlled hosts are respected: `activeChatId` can be state or callback-only. Restored chats are reported via the callback, and clearing the id does not snap back to the previous session.

**Migration**

- `ChatStorageProvider.createChat` now accepts optional `id`. Providers must adopt it and, when present, return the existing chat instead of creating a duplicate. Providers that mint their own id may ignore it and return their id; callers keep the returned value.

<sup>Written for commit 4c9b7b57d481b4d6552a3d0a3266ff1154e9bdfe. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2232?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
@@ -253,7 +253,6 @@ export function ChatTab({
     activeChatId,
     selectChat,
     startNewChat,
-    ensureActiveChat,
     persistSessionMessages,
   } = useChatSessions({
     retryServerId: connection.id ?? connection.url,
@@ -440,18 +439,6 @@ export function ChatTab({
     );
     await startNewChat();
   }, [effectiveChatStorage, startNewChat]);
-
-  // Sending into a chat history has not seen yet shows a placeholder title
-  // until title generation replaces it.
-  const prepareActiveChat = useCallback(() => {
-    if (
-      effectiveChatStorage &&
-      !chatSessions.get(activeSessionId).persistedChatId
-    ) {
-      setInternalChatTitle(CHAT_TITLE_PLACEHOLDER);
-    }
-    return ensureActiveChat();
-  }, [activeSessionId, chatSessions, effectiveChatStorage, ensureActiveChat]);
 
   const handleTitleGenerated = useCallback(
     (chatId: string, title: string) => {
@@ -1298,18 +1285,24 @@ export function ChatTab({
     if (messages.length === 0) {
       dismissLandingShader();
     }
-    void prepareActiveChat().then(() => {
-      sendMessage(inputValue, results);
-      setInputValue("");
-      clearPromptResults();
-    });
+    if (
+      effectiveChatStorage &&
+      !chatSessions.get(activeSessionId).persistedChatId
+    ) {
+      setInternalChatTitle(CHAT_TITLE_PLACEHOLDER);
+    }
+    void sendMessage(inputValue, results);
+    setInputValue("");
+    clearPromptResults();
   }, [
     inputValue,
     results,
     sendMessage,
     clearPromptResults,
     attachments,
-    prepareActiveChat,
+    activeSessionId,
+    chatSessions,
+    effectiveChatStorage,
     messages.length,
     dismissLandingShader,
   ]);
--- a/libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
@@ -67,6 +67,7 @@ import { useWidgetDebug } from "../context/WidgetDebugContext";
 import type { ChatBodyBuilder, MessageAttachment } from "./chat/types";
 import { resolveChatToolPolicy } from "./chat/chat-tool-policy";
 import { buildChatCspAudit } from "./chat/csp-bridge";
+import { useChatSessions } from "./chat/useChatSessions";
 
 // Structural type — avoids nominal incompatibility when pnpm creates
 // multiple peer-variant copies of mcp-use with duplicate class declarations.
@@ -240,29 +241,33 @@ export function ChatTab({
       ? localChatStorageRef.current
       : null);
 
-  const [internalActiveChatId, setInternalActiveChatId] = useState<
-    string | null
-  >(null);
-  const activeChatId = controlledActiveChatId ?? internalActiveChatId;
-  const setActiveChatId = useCallback(
-    (chatId: string | null) => {
-      if (onActiveChatIdChange) {
-        onActiveChatIdChange(chatId);
-      } else {
-        setInternalActiveChatId(chatId);
-      }
-    },
-    [onActiveChatIdChange]
-  );
-
   const [showHistoryPanel, setShowHistoryPanel] = useState(false);
   const [historyRefetchKey, setHistoryRefetchKey] = useState(0);
+  const bumpHistoryRefetch = useCallback(() => {
+    setHistoryRefetchKey((k) => k + 1);
+  }, []);
+
+  const {
+    store: chatSessions,
+    activeSessionId,
+    activeChatId,
+    selectChat,
+    startNewChat,
+    ensureActiveChat,
+    persistSessionMessages,
+  } = useChatSessions({
+    retryServerId: connection.id ?? connection.url,
+    agentId: serverId,
+    agentName: connection.displayName || connection.name || "MCP Server",
+    storage: effectiveChatStorage,
+    activeChatId: controlledActiveChatId,
+    onActiveChatIdChange,
+    initialMessages,
+    onChatCreated: bumpHistoryRefetch,
+  });
   const [internalChatTitle, setInternalChatTitle] = useState<
     string | undefined
   >();
-  const [restoredMessages, setRestoredMessages] = useState<
-    import("./chat/types").Message[] | undefined
-  >(undefined);
   const [inputValue, setInputValue] = useState("");
   const [promptsDropdownOpen, setPromptsDropdownOpen] = useState(false);
   const [promptFocusedIndex, setPromptFocusedIndex] = useState(-1);
@@ -331,6 +336,9 @@ export function ChatTab({
 
   // Use client-side or server-side chat implementation
   const chatHookParams = {
+    sessionStore: chatSessions,
+    sessionId: activeSessionId,
+    onMessagesChange: persistSessionMessages,
     connection,
     llmConfig,
     isConnected,
@@ -341,6 +349,9 @@ export function ChatTab({
   };
 
   const serverSideChat = useChatMessages({
+    sessionStore: chatSessions,
+    sessionId: activeSessionId,
+    onMessagesChange: persistSessionMessages,
     mcpServerUrl: connection.url ?? "",
     llmConfig,
     authConfig: userAuthConfig,
@@ -349,7 +360,6 @@ export function ChatTab({
     chatApiUrl,
     waitForChatApiUrl,
     widgetModelContexts,
-    initialMessages: restoredMessages ?? initialMessages,
     disabledTools: effectiveDisabledTools,
     streamProtocol,
     credentials,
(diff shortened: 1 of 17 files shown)


libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
```diff
@@ -219,20 +224,85 @@ export function ChatTab({
...
+    controlledActiveChatId,
+    internalActiveChatId
+  );
+  const nextChatSessionKeyRef = useRef(0);
+  const initialChatSessionKeyRef = useRef(
+    initialPendingChatTurn?.sessionKey ??
+      initialPersistedChatId ??
+      `draft:${serverId}:0`
+  );
+  const chatIdBySessionKeyRef = useRef(
```
khandrew1: These three maps make sense individually, but together they introduce two identities: sessionKey and chatId. The code then has to keep the mappings between them synchronized, and a missed update could associate state or messages with the wrong chat.

Could we instead assign each session a UUID when it is created and use that as the canonical identity for both runtime and persisted state? That would remove most of this translation and make the session lifecycle easier to reason about.

libraries/typescript/packages/inspector/src/client/components/chat/keyed-state.ts
```diff
@@ -0,0 +1,76 @@
...
+export function resolveStateAction<T>(
+  previous: T,
+  action: SetStateAction<T>
+): T {
+  return typeof action === "function"
+    ? (action as (value: T) => T)(previous)
+    : action;
+}
+
+export function useKeyedState<T>(
```
khandrew1: Keying each state field independently fragments a single chat session across several unrelated stores. Messages, loading state, attachments, traces, authorization, and runtime all transition together conceptually, but nothing here can update or clean them up atomically.

I think these values should live in one session record/store keyed by the session UUID.

khandrew1 (changes requested):
Thanks for taking this on! The core issue is correct, state and runtime do need to be isolated per chat session so that we can start new chats without affecting an already in flight session.

I'd prefer a model where every chat session receives one UUID at creation and all state lives in one session record. 

I don’t think the current architecture gives us a safe foundation for that behavior. It introduces transient draft keys, persisted chat IDs, two directional identity maps, a creation-promise map, and separate keyed stores for every piece of session state. This distributes one session lifecycle across many independently maintained data structures.

Perhaps something like this:
```ts
type ChatSession = {
  id: string;
  messages: Message[];
  attachments: MessageAttachment[];
  isLoading: boolean;
  trace: InspectorTraceState;
  runtime: ChatRuntime;
  creation?: Promise<void>;
};
```

When history persistence is enabled, that same UUID should ideally be passed to ChatStorageProvider.createChat, making the runtime session ID and persisted chat ID identical. If a backend must generate a separate ID, it can be an optional field on the session with at most one secondary lookup index.

I’d consider the redesign complete when:
- every session has one stable UUID;
- session state and runtime are owned by one store;
- **New Chat** only changes the active UUID;
- background updates continue against their original session record;
- persistence and OAuth recovery use the same identity;
- integration tests cover starting a second chat during streaming, switching back, controlled-ID clearing, and redirect recovery.

I realize this is a larger change than patching the existing implementation, but I think it will substantially reduce the code and make the behavior easier to verify. I’m happy to help work through the session-store shape or storage-provider change. 

If you join our Discord I'm able to give more realtime help as well.

merttcetn:
@khandrew1 thanks for the detailed review, you were right that the previous architecture was not a safe foundation. I rewrote it around the model you described rather than patching the old one, and it did come out smaller: `ChatTab` lost about 100 lines of session bookkeeping and `keyed-state.ts` is gone entirely.

Against your checklist:

**Every session has one stable UUID.** `createChatSessionId()` mints it at creation and the session keeps it. No draft keys, no counters, so there is nothing left to collide after an OAuth redirect.

**Session state and runtime are owned by one store.** `ChatSessionStore` holds one `ChatSessionState` record per session:

```ts
interface ChatSessionState {
  readonly id: string;
  messages: Message[];
  attachments: MessageAttachment[];
  isLoading: boolean;
  trace: InspectorTraceState;
  managedChatNotice: ManagedChatNotice | null;
  pendingAuthorization: PendingAuthorization | null;
  authenticatingToolCallId: string | null;
  toolAuthorizationError: string | null;
  mcpServerAuthRequired: McpServerAuthRequired | null;
  persistedChatId: string | null;
  creation: Promise<string | null> | null;
  readonly runtime: ChatSessionRuntime; // abort controller, send guard, trace counter, auth gate
}
```

The store subscribes per session id, so `useSyncExternalStore` only re-renders the chat you are looking at. A background turn updates its own record without touching the UI, which also fixes the unnecessary-rerender point Copilot raised.

**New Chat only changes the active UUID.** `startNewChat()` mints an id, activates it, and creates the persisted chat. Nothing is cleared, so an in-flight turn keeps running and the new chat starts idle. The only extra bit of lifecycle is dropping the previous record when it was never used, so empty sessions do not pile up.

**Background updates continue against their original session record.** The setters are bound to the session id captured when the turn started, so a stream that outlives the switch still writes to the chat that started it.

**Persistence and OAuth recovery use the same identity.** `ChatStorageProvider.createChat` now takes an optional `id`, and the local provider adopts it and returns the existing chat when it already stored that id, so reattaching after a redirect cannot fork a duplicate row. The pending-turn record is keyed by `serverId + sessionId` with one pointer per server for which session the redirect interrupted, which is what lets the reloaded tab reopen the right chat. For your backend caveat: if a provider mints its own id, it lands in `persistedChatId` on the session record with `findByPersistedChatId` as the single lookup back. That path has a test too.

**Integration tests.** `useChatSessions.test.tsx` renders the lifecycle with React and a fake storage provider, covering the four cases you listed: starting a second chat while the first is streaming, switching back to a live session (including that it is not overwritten by its last persisted snapshot), a controlled host clearing `activeChatId`, and redirect recovery. Plus concurrent-write dedupe, the backend-minted-id path, and running with history disabled. `chat-session-store.test.ts` covers isolation and notification scoping, and `chat-auth-retry.test.ts` covers two chats in OAuth at once.

`pnpm --filter @mcp-use/inspector test:unit` is 233 passing, and type-check, `lint:strict`, `format:check` and `build` are all clean. Also rebased onto `main`, so the conflict is resolved.

One thing worth a second opinion: `ChatTab` mounts both chat hooks and they now share the one store, so a BYOK/managed switch keeps the conversation on screen instead of swapping it for the other hook's list. That felt like the more coherent behavior under a single identity, but say the word if you would rather keep them separate.

I will join the Discord as well, thanks for the offer.

libraries/typescript/packages/inspector/src/client/components/ChatTab.tsx
merttcetn: Agreed, both maps and the creation-promise map are gone. A session is created with one UUID and that same id is passed to `ChatStorageProvider.createChat`, so there is no translation left to keep in sync. The only remaining indirection is `persistedChatId` on the session record, for a backend that insists on minting its own id, with `findByPersistedChatId` as the single lookup back.

libraries/typescript/packages/inspector/src/client/components/chat/keyed-state.ts
merttcetn: `keyed-state.ts` is deleted. All of those values now live in one `ChatSessionState` record in `ChatSessionStore`, keyed by the session UUID, so they update and get cleaned up together. The store notifies per session id, so a background turn writing to its own record does not re-render the chat on screen.

merttcetn:
One note on the size, since +1615/-312 across 15 files looks heavier than it reads:

| | insertions | |
|---|---|---|
| Tests + changeset | 675 (42%) | the four scenarios you asked for, plus store isolation and concurrent OAuth |
| New modules | 464 | `chat-session.ts` (101, mostly the record type), `chat-session-store.ts` (142), `useChatSessions.ts` (221) |
| The two chat hooks | 345 (-173) | mechanical: every `useState`/`useRef` became a read/write on the session record, and the `useCallback` dependency arrays followed |
| `ChatTab.tsx` | 57 (-85) | net shrink |
| Storage provider | 15 | the optional `createChat({ id })` |

So the genuinely new logic to review is the 464 lines in three focused files; most of the rest is either tests or the unavoidable churn of moving off hook-wide state.

I should also be straight about one thing: you expected this to substantially reduce the code, and at the file level it did not. Production code went from 667 insertions in the previous revision to 940. The reduction happened in `ChatTab` only. The growth is the record type written out explicitly, the session lifecycle now being a named module instead of being spread through `ChatTab`, and behavior that simply did not exist before (per-session persistence, not overwriting a live session on switch-back, controlled-id clearing). Happy to trim if any of it reads as overbuilt to you.

khandrew1 (approved):
Thanks for the PR and working through the feedback! I simplified the API by removing `useChatSessionField` and the multiple ensure calls, instead keeping session lifecycle and persistence internally.

Also did some modifications to the test suite.

PR #2234, opened 2026-08-11
feat(server): export createJwtVerifier from mcp-use/oauth

### LINKED ISSUE #2215: Export `createJwtVerifier`
Currently, `createJwtVerifier` isn't exported, requiring users to write their own to use a custom OAuth provider. To make it easier to implement custom OAuth providers, this function should be exported out so consumers can also use it.

ayaangazali:
Closes #2215.

### The gap

`mcp-use/oauth` already exports `oauthCustomProvider` and `CustomOAuthProviderOptions`, so building a custom provider is a supported path. But `createJwtVerifier` stayed internal, so anyone taking that path had to hand-roll signature verification, issuer and audience matching, `exp`/`sub` enforcement and the resource binding, all of which the six built-in providers already share through this one function.

`src/oauth/jwt.ts` is not reachable from any entry point in `package.json`, so there was no way to reach it short of importing from `dist` internals.

### The change

Export `createJwtVerifier` from the `./oauth` barrel, together with the two types you need to call it and to consume its result:

- `JwtVerifierOptions`, the argument type
- `VerifiedPayload`, the claims record

`OAuthTokenVerifier`, the return type, was already exported from that barrel.

All three carried `@internal`, so I removed those tags. The package requires TSDoc on public exports including type properties, so `issuer`, `jwksUrl`, `resource`, `algorithms` and `key` are now documented too, and `createJwtVerifier` gained `@param`, `@returns` and a `@throws` for the case where `audience` is combined with `issuerBoundAccessTokens`.

### Verified

Built and checked the published surface rather than assuming:

```
$ grep createJwtVerifier packages/server/dist/oauth/index.d.ts
export { createJwtVerifier, type JwtVerifierOptions, type VerifiedPayload, } from "./jwt.js";

$ node -e "import('./packages/server/dist/oauth/index.js').then(m=>console.log(typeof m.createJwtVerifier))"
function
```

Server suite: 446 passed. Three failures in `tests/usage.test.ts` are pre-existing on `canary` and unrelated to this change: they only appear when `MCP_USE_ANONYMIZED_TELEMETRY=false` is set, and they fail identically with my diff stashed.

No behavior change, only visibility.

---
## Summary by cubic
Exported `createJwtVerifier` from `mcp-use/oauth` so custom providers built with `oauthCustomProvider` can reuse built-in JWT verification instead of reimplementing it. No runtime behavior changes.

- **New Features**
  - Export `createJwtVerifier`, `JwtVerifierOptions`, and `VerifiedPayload` from `mcp-use/oauth`.
  - Added TSDoc for new public options and removed `@internal` tags.

<sup>Written for commit 440b10040b4ed8f42f77b40b10ae317ce6312f3e. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2234?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/olive-pugs-repeat.md
+++ b/libraries/typescript/.changeset/olive-pugs-repeat.md
@@ -1,0 +1,5 @@
+---
+"mcp-use": minor
+---
+
+Export `createJwtVerifier` and its `JwtVerifierOptions` and `VerifiedPayload` types from `mcp-use/oauth`, so a custom OAuth provider built with `oauthCustomProvider` can reuse the built-in JWT verification instead of reimplementing it.
--- a/libraries/typescript/packages/server/src/oauth/index.ts
+++ b/libraries/typescript/packages/server/src/oauth/index.ts
@@ -23,6 +23,11 @@ export type {
 
 export { bearerAuth, oauthMetadata } from "./adapters.js";
 export {
+  createJwtVerifier,
+  type JwtVerifierOptions,
+  type VerifiedPayload,
+} from "./jwt.js";
+export {
   oauthCustomProvider,
   type CustomOAuthProviderOptions,
   type OAuthExtra,
--- a/libraries/typescript/packages/server/src/oauth/jwt.ts
+++ b/libraries/typescript/packages/server/src/oauth/jwt.ts
@@ -15,13 +15,16 @@ import { isLocalhost, isRecord } from "./guards.js";
 
 export { invalidToken } from "./errors.js";
 
-/** @internal Record of claims extracted from a verified JWT. */
+/** Record of claims extracted from a verified JWT. */
 export type VerifiedPayload = Record<string, unknown>;
 
-/** @internal Configures issuer, keys, and claims enforced by a JWT verifier. */
+/** Configures issuer, keys, and claims enforced by a JWT verifier. */
 export interface JwtVerifierOptions {
+  /** Expected `iss` claim. Tokens from any other issuer are rejected. */
   issuer: string;
+  /** JWKS endpoint the signing keys are fetched from and cached. */
   jwksUrl: URL;
+  /** Canonical URL of this MCP resource, matched against the token binding. */
   resource: URL;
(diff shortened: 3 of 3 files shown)


ayaangazali:
The red `typescript-knip` here is pre-existing on `canary`, not from this change.

The job is failing on the branch itself:

```
31513157888  typescript-knip: failure
31513151167  typescript-knip: failure
```

Its seven findings are three unused exported types in `packages/cli` and `packages/inspector`, plus four stale `knip.json` entries. None of them touch `packages/server`, and the two types this PR makes public are not among them, since both are used internally by the six built-in providers.

I opened #2235 against `canary` to clear those findings rather than folding unrelated changes into this PR.

ayaangazali:
Rebased onto current `canary`. `typescript-knip` is green now that #2235 has merged: https://github.com/mcp-use/mcp-use/actions/runs/31543330283/job/93950685053

The other two reds on the previous run were also not from this diff, which only adds exports and TSDoc under `packages/server/src/oauth`.

**Report current-head conformance scores** failed on `Artifact not found for name: conformance-results` followed by `RequestError [HttpError]: Resource not accessible by integration`. That is the reporting job lacking permissions on a fork PR, not a conformance result.

**Test with Deno (PR Preview)** spent its whole 17 minutes waiting on the preview packages and then ran anyway:

```
Waiting for packages... (attempt 9/10, PR: false, Commit: false)
⚠️ Packages didn't become fully available after 10 attempts (5 minutes)
  PR URL ready: false
  Commit URL ready: false
Proceeding with tests anyway...
```

The tests import the pkg.pr.new preview build through esm.sh, and it never appeared, so the step had nothing to resolve against. I cannot rerun workflows from a fork, so if it stays red on this run that is the reason.

PR #2193, opened 2026-08-07
fix(ci): stop knip failing on the skills-loader subpath import

ayaangazali:
Closes #2189.

`pnpm exec knip` exits 1 on `canary`:

```
Unresolved imports (1)
#mcp-use-skills-loader  packages/server/src/server.ts:658:48
```

Nothing is actually broken. The import is declared and resolves at runtime:

```json
// packages/server/package.json
"#mcp-use-skills-loader": { "node": "@mcp-use/cli/internal/skills-loader", ... }

// packages/cli/package.json
"./internal/skills-loader": { "import": "./dist/internal/skills-loader.js" }
```

built by tsup from `src/skills/node-loader.ts`. knip just cannot follow a subpath import that resolves into another workspace package, which is why its sibling `#mcp-use-node-http` is already in `packages/server.ignoreUnresolved`. The skills-loader entry was never added when that work landed.

That unresolved import is the only knip **error**: `rules.exports` and `rules.types` are `warn`, so the two unused inspector exports and the unused type do not fail the build. Adding the entry is enough.

## Verification

```
$ pnpm exec knip --no-progress   # before
exit 1, Unresolved imports (1)

$ pnpm exec knip --no-progress   # after
exit 0, no Unresolved imports section
```

## Why this is worth a PR rather than leaving it

`typescript-knip` is skipped on canary's own runs, so canary never shows this failure. It surfaces on the next PR that touches TypeScript instead: #2188 changes only `create-mcp-use-app/src/skills-config.ts` and inherited a red `typescript-knip` for it.

## Two things I left alone

knip hints that `#mcp-use-node-http` can now be dropped from `ignoreUnresolved`. That hint appears in my local run but not in CI's, so it resolves differently by environment and removing it would likely turn CI red.

The unused inspector exports (`READ_SKILL_RESOURCE_TOOL`, `supportsSkills`, `SkillContextConnection`) are warnings, not errors, and may be intentional staging for the skills work. Say the word if you want them removed and I will send that separately.

## Note on preflight

`pnpm run lint` fails on this branch and identically on untouched `canary`: 41 problems, 39 errors, all in `packages/cli/src/`. This PR changes one line of `knip.json` and cannot affect eslint.

---
## Summary by cubic
Stop `knip` from failing by ignoring the `#mcp-use-skills-loader` subpath import in `packages/server`. The import resolves to `@mcp-use/cli/internal/skills-loader` at runtime; adding it to `ignoreUnresolved` makes `knip` exit 0 again.

[...]

PR #2235, opened 2026-08-11
fix(ci): clear the knip findings failing canary

ayaangazali:
### The defect

`typescript-knip` is currently failing on `canary` itself, not just on PRs. Latest runs on the branch:

```
31513157888  typescript-knip: failure
31513151167  typescript-knip: failure
```

Every PR opened against `canary` inherits that red, which makes it hard to tell a real regression from the standing failure.

The job reports seven items in two groups.

**Three exported types nothing imports:**

- `DevClientEndpoint`, `packages/cli/src/cli/dev-client-endpoint.ts:2`
- `SkillContextConnection`, `packages/inspector/src/client/components/chat/skill-context.ts:13`
- `SkillsState`, `packages/inspector/src/client/components/layout/layoutHeaderUtils.ts:47`

Each is referenced only inside its own module, as a return type or parameter, so `export` was doing nothing. Dropped the keyword rather than adding suppressions, since these are not reachable API: none of the three files sits behind an entry point in its package's `exports` map.

**Four stale `knip.json` entries.** knip asks for these to be removed, which means the reason they were ignored no longer holds:

- `test_app` in `ignoreWorkspaces` is the scaffold output directory the create-app matrix used to leave behind. It is not in the repo.
- `@tailwindcss/vite`, `@vitejs/plugin-react` and `tailwindcss` in `packages/cli` `ignoreDependencies` are resolved properly now.

### Verification, and its limit

`pnpm build` passes, so nothing relied on those three type exports for declaration emit. That was the one real risk in removing them.

I want to be straight about the rest: I cannot run this workflow from a fork, and my local `knip` reports a different set of findings than the CI job does, so I did not use it as evidence. What I changed is exactly the seven items the CI job listed, and this PR's own `typescript-knip` run is the check that matters.

---
## Summary by cubic
Fixes the failing `typescript-knip` job on `canary` by removing unused exports, cleaning stale `knip.json`, and declaring `jsdom` for an inspector test. Restores a green CI state with no public API changes.

- **Bug Fixes**
  - Removed unused exports (three internal types and the inspector reconnect storage key).
  - Cleaned `knip.json`: dropped `test_app` and trimmed CLI `ignoreDependencies` (now correctly detects `@tailwindcss/vite`, `@vitejs/plugin-react`, `tailwindcss`).
  - Added `jsdom` to `@mcp-use/inspector` devDependencies to match the `@vitest-environment jsdom` test.

<sup>Written for commit d24f38663fa777526d1fb70dcdc3af92cfe7f34d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2235?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/quick-hounds-tap.md
+++ b/libraries/typescript/.changeset/quick-hounds-tap.md
@@ -3,4 +3,4 @@
 "@mcp-use/cli": patch
 ---
 
-Drop three type exports that nothing imports. They are used only inside their own modules, so this is not a change to any reachable API.
+Drop four exports that nothing imports, and declare the `jsdom` devDependency the inspector tool-execution test already relies on through its `@vitest-environment jsdom` pragma. The dropped exports are used only inside their own modules, so this is not a change to any reachable API.
--- a/libraries/typescript/.changeset/quick-hounds-tap.md
+++ b/libraries/typescript/.changeset/quick-hounds-tap.md
@@ -1,0 +1,6 @@
+---
+"@mcp-use/inspector": patch
+"@mcp-use/cli": patch
+---
+
+Drop three type exports that nothing imports. They are used only inside their own modules, so this is not a change to any reachable API.
--- a/libraries/typescript/knip.json
+++ b/libraries/typescript/knip.json
@@ -8,8 +8,7 @@
   "ignoreWorkspaces": [
     "packages/client/examples/**",
     "packages/create-mcp-use-app/src/templates/**",
-    "packages/server/examples/**",
-    "test_app"
+    "packages/server/examples/**"
   ],
   "workspaces": {
     ".": {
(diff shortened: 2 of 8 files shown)


ayaangazali:
`typescript-knip` is green now: https://github.com/mcp-use/mcp-use/actions/runs/31514584158/job/93856882298

Two corrections to what I wrote above.

**The first push did not fix it.** Clearing the seven items surfaced two more, and they were not mine: `canary` moved while this was open, and #2216 added both an inspector test that opts into jsdom through a `@vitest-environment` pragma without the package declaring `jsdom`, and an exported reconnect storage key that is only read inside its own module. Declared the dependency at the `^27.0.0` the client package already uses, and dropped the export.

**My claim that local knip reports a different set was wrong**, and I should have looked closer before saying it. Local and CI agree on everything except two `ignoreUnresolved` hints for `#mcp-use-node-http` and `#mcp-use-skills-loader`. Locally those subpaths resolve because I have a built `dist`, so knip calls the entries redundant. The CI job runs knip without building, so there they are genuinely unresolved and the entries are load-bearing. Removing them would undo #2193 and turn the job red again, so they stay.

khandrew1 (approved):

PR #2267, opened 2026-08-16
fix(ci): report expected-failure entries that no longer match anything

### LINKED ISSUE #2266: Conformance summary never reports obsolete expected-failure entries, so the allowlist can silently mask a regression
## What happens

`summarize-conformance.mjs` classifies observed failures against `expected-failures.yml`, but never checks the list in the other direction. Nothing reports an entry that no longer corresponds to a failing check.

`isExpectedOutcome` is called in exactly two places, both filtering results that already failed:

```js
const expectedFailuresCount = failures.filter((check) => isExpectedOutcome(suite, check)).length;   // line 60
const expectedWarnings = warningChecks.filter((check) => isExpectedOutcome(suite, check)).length;   // line 64
```

and the job's only failure condition is:

```js
if (totals.unexpectedFailures > 0) process.exitCode = 1;   // line 118
```

So if one of the 12 entries starts passing, its check is simply counted in `passed`. The score goes up, the run stays green, and the obsolete line stays in the file.

## Why it matters

A stale entry is a silenced check. If `resources-subscribe` were implemented for stateless legacy HTTP and later regressed, the regression would land in `expectedFailures` rather than `unexpectedFailures`, and CI would stay green on a real break. The allowlist protects the exact checks nobody is watching.

The same gap hides two smaller problems, both from nothing validating the file against reality:

- A typo or a renamed scenario matches nothing and is silently inert rather than erroring.
- `isExpectedOutcome` returns false unless the suite starts with `2025-11-25/server/`, so entries silently do not apply to any other spec version. That may well be intended, but today there is no signal either way.

## Suggested fix

Track which entries matched at least one check, then report the ones that did not, something like an `Obsolete expected-failure entries` section in the summary. Whether that should also set a non-zero exit code is your call: failing the build keeps the file honest, but it would turn a fixed conformance check into a red run until someone prunes the list, so report-only may be the better default.

I am happy to implement it. What I would need first is confirmation of the suite directory naming, because the matcher depends on it:

```js
suite.includes(`/server/server-${scenario}-`)
```

That requires a trailing hyphen after the scenario, so I did not want to write the reverse check against a naming convention I have only inferred from the matcher rather than seen in real results. If you can point me at a conformance results artifact, or confirm the layout, I will send the PR.

ayaangazali:
Closes #2266.

### The defect

`summarize-conformance.mjs` checked observed failures against `expected-failures.yml`, but never checked the list in the other direction. `isExpectedOutcome` was only ever called on checks that had already failed or warned, and the run's only failure condition is `unexpectedFailures > 0`.

So when a listed check starts passing, it is counted in `passed`, the run stays green, and the entry stays in the file. That entry is now a silenced check: if the same check regresses later, it is classified as an expected failure and CI stays green on a real break. The allowlist protects exactly the checks nobody is watching.

### The change

Record which entries classify at least one failure or warning, then list the leftovers under an `Obsolete expected-failure entries` heading in the summary.

Report only, no exit-code change. A run that skips the suites an entry covers would list that entry too, so failing the build on this by default would be wrong. The heading says so explicitly.

### Verified against a real run

I pulled the `typescript-v2-conformance-results` artifact from run 31758544746 on `main` and ran the script against it, rather than guessing at the layout. That also answers the question I left open in #2266: suites are `2025-11-25/server/server-<scenario>-<timestamp>Z`, so the existing matcher's trailing hyphen is correct.

**No false positives.** All 10 current entries match, 6 failures and 4 warnings, which lines up with the reported `6 expected failures, 4 expected warnings`. The generated summary is byte-identical to the one produced before this change:

```
$ diff -q summary-before.md summary-after.md
(no output)
```

**It catches the real cases.** Appending two bad entries to a scratch copy, one for a suite that exists and passes, and one with a check id that does not exist:

```
**Obsolete expected-failure entries.** ...

- `resources-list`
- `tools-call-sampling:no-such-check`
```

Exit code stayed 0 in both runs.

`resources-list` is the exact rot scenario: the suite runs and passes, so under the old code the entry would sit there indefinitely, ready to absorb a future regression.

---
## Summary by cubic
Reports obsolete entries in `expected-failures.yml` in the conformance summary to prevent masking regressions. Previously we only evaluated observed failures; now we track which expected-failure entries matched a failure or warning and list the rest.

- Adds an “Obsolete expected-failure entries” section listing entries that matched no failing or warning check in this run.
- Report only; exit code unchanged. Entries for skipped suites can appear here.
- Required action: remove listed entries from `expected-failures.yml` to avoid silencing future regressions.

<sup>Written for commit 740e84d2651f614cb1d486c5e7048d883efffe2f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2267?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/scripts/summarize-conformance.mjs
+++ b/libraries/typescript/scripts/summarize-conformance.mjs
@@ -38,17 +38,27 @@ const expectedFailures = expectedFailuresFile
       .filter(Boolean)
   : [];
 
+// Entries that classified at least one failure or warning. Anything left over
+// no longer describes a real outcome, so it is silently suppressing whatever
+// that check does next.
+const matchedExpectedFailures = new Set();
+
 function isExpectedOutcome(suite, check) {
   if (!suite.startsWith("2025-11-25/server/")) return false;
-  return expectedFailures.some((entry) => {
+  let matched = false;
+  for (const entry of expectedFailures) {
     const separator = entry.indexOf(":");
     const scenario = separator === -1 ? entry : entry.slice(0, separator);
     const checkId = separator === -1 ? undefined : entry.slice(separator + 1);
-    return (
+    if (
       suite.includes(`/server/server-${scenario}-`) &&
       (!checkId || check.id === checkId)
-    );
-  });
+    ) {
+      matchedExpectedFailures.add(entry);
+      matched = true;
+    }
+  }
+  return matched;
 }
 
 for (const file of files.sort()) {
@@ -110,6 +120,19 @@ const lines = [
       `| \`${row.suite}\` | ${row.passed}/${row.total} | ${row.expectedFailures} | ${row.unexpectedFailures} | ${row.expectedWarnings} | ${row.unexpectedWarnings} |`
   ),
 ];
+
+const obsoleteExpectedFailures = expectedFailures.filter(
+  (entry) => !matchedExpectedFailures.has(entry)
+);
+
+if (obsoleteExpectedFailures.length > 0) {
+  lines.push(
+    "",
+    "**Obsolete expected-failure entries.** These matched no failing or warning check in this run, so they are no longer describing a real outcome. Remove them, or they will absorb the next regression of that check. If a run skips the suites they cover, they will show up here too.",
+    "",
+    ...obsoleteExpectedFailures.map((entry) => `- \`${entry}\``)
+  );
+}
 
 const summary = `${lines.join("\n")}\n`;
 if (output) writeFileSync(output, summary);


ayaangazali:
Both of these are pre-existing timeouts on `main`, not from this change, so I am not pushing a fix.

This PR touches one file, `libraries/typescript/scripts/summarize-conformance.mjs`, which runs only in the conformance workflow. It is not imported by the inspector or CLI test suites.

**`typescript/inspector`.** Fails at `Run Inspector Tests`:

```
FAIL tests/unit/compressed-assets.test.ts > serves stored gzip bytes and a byte-identical identity fallback
 ❯ tests/unit/compressed-assets.test.ts:6:3
```

The identical failure, same file and same line, is on `main` in run 31758544677 (job 94639889250):

```
❯ tests/unit/compressed-assets.test.ts (1 test | 1 failed) 10837ms
 ❯ tests/unit/compressed-assets.test.ts:6:3
```

It took 10837ms against a 10000ms limit, 8 percent over. And the same job passed on `main` in run 31758440946, two minutes before that failure. Same branch, opposite results, so it is timing-dependent rather than broken.

**`typescript/cli (windows-latest)`.** Fails at `Run CLI Tests`:

```
```

Same file and same mechanism on `main`, run 31749127452 (job 94610964438), where four `dev.test.ts` cases failed on ubuntu:

```
```

Both suites bring up real servers and watch the filesystem, so they are the ones that stretch past a fixed timeout on a loaded runner, and Windows is the slowest of the three.

Worth flagging since it will keep costing people reruns: these two suites only started running in CI recently, `test:unit` for the inspector in #2201 and the mcp-use suite in #2213, so the flakiness is newly visible rather than new. If you want, I will open a separate issue proposing longer timeouts for the server-startup and asset-compression cases rather than the current defaults. I would rather not fold that into this PR.

ayaangazali:
Opened #2268 for the inspector half of this, since it turned out to be a one-line inconsistency rather than a judgement call: the inspector is the only package capped at `testTimeout: 10000`, while agent, client, cli and server all use 60000. The case that failed takes 2832ms locally and hit 10837ms on the runner.

The `typescript/cli (windows-latest)` flake is a different mechanism, `waitFor timed out after 15000ms` from the test helper rather than vitest, and cli is already at 60000, so I left it alone rather than guess.

ayaangazali:
CI note: the two red jobs here are unrelated to this change.

This PR touches exactly one file, `libraries/typescript/scripts/summarize-conformance.mjs`. The failure is in `packages/inspector` — `tests/unit/compressed-assets.test.ts > serves stored gzip bytes and a byte-identical identity fallback`, which failed with `Test timed out in 10000ms` after 10331ms. It is a timeout a third of a second over the limit in a package this branch does not touch, not an assertion failure.

The branch is already level with `main` (0 commits behind), so there is nothing to rebase onto that would clear it, and I do not have rerun rights on the repo. A maintainer re-run should be green; if that test is timing out routinely on CI runners, raising its `testTimeout` would be a separate change and I am happy to open it.

ayaangazali:
Rebased onto current `main` to get a fresh run, since the previous checks dated from the 16th. New head `1a32dce4`. Result splits the two:

**`typescript/cli (windows-latest)` is now green.** Same code, fresh run, so that one was a flake rather than anything in this diff.

**`typescript/inspector` failed again**, same cause as before:

```
```

That is `compressed-assets.test.ts` against the inspector's 10s cap, which #2268 raises to the 60s every other package already uses. `main` still carries `testTimeout: 10000`, so it will keep recurring here until that lands. Recent margins across branches: 10029ms, 10230ms, 10837ms.

This PR only touches `scripts/summarize-conformance.mjs`, which no test suite imports. Nothing further I can do on this branch.

ayaangazali:
Green: 31 checks, 0 failures. Rebased onto current `main` (`ad577fac`) and both previously failing jobs passed this time, `typescript/inspector` and `typescript/cli (windows-latest)`.

Worth noting `main` still has `testTimeout: 10000` for the inspector, since #2290 landed on `canary`. So the compressed-assets case passing here is the flake falling the other way rather than the cap being fixed, and it may reappear on a future run until the release carries that change across. The margins have been 10029ms, 10230ms and 10837ms against a 10000ms limit.

The change itself is unrelated to any of that and unchanged: the conformance summary never reported expected-failure entries that no longer match anything, so a stale allowlist entry would silently absorb the next regression of that check. Verified against a real artifact, currently 0 obsolete entries out of 10.

ayaangazali:
@tonxxd this one is a week old with no decision either way, so a short summary in case it is easier to judge than to reopen.

You own `summarize-conformance.mjs` (#2205, #2206). Today an entry in `expected-failures.yml` that stops matching anything just goes quiet: the suite still passes, and the stale entry keeps sitting there suppressing a failure that no longer exists. This adds a report-only section listing entries that matched nothing on the run. No exit-code change, so it cannot turn a green run red.

One file, 27 added lines. Rebased onto current `main` just now (`740e84d2`), still green.

If you would rather this failed the run instead of only reporting, or if stale entries are expected to accumulate and you do not want the noise, say so and I will change it or close this.

ayaangazali:
Rebased onto current main after today's merges. Green, 17 checks, mergeable.

@khandrew1 since you are going through the queue: this is the one that needs a decision rather than a review. Today a stale entry in `expected-failures.yml` just goes quiet, so it keeps suppressing a failure that no longer exists. This adds a report-only section listing entries that matched nothing, with no exit-code change, so it cannot turn a green run red.

@tonxxd owns this file (#2205, #2206) and I asked him on the 23rd whether he would rather it hard-fail instead. Either shape is a small edit from here. If neither is wanted, say so and I will close it.

khandrew1:
@ayaangazali sorry this one is taking so long, though I think I'm still pretty confused on what exactly this PR is trying to accomplish

ayaangazali:
No apology needed, and the confusion is my fault for describing it in terms of the mechanism rather than the failure mode. One sentence: **it tells you when an entry in `expected-failures.yml` has stopped matching anything, because a stale entry silently absorbs the next regression of that check.**

Here is the whole behaviour, run against a two-check fixture. `resources-subscribe` still matches a real failure; `resources-unsubscribe` matches nothing.

**Before (current main):**

```
**Score: 1/2 passed** (1 expected failures, 0 unexpected failures, ...)

| Suite | Score | Expected failures | ... |
| `2025-11-25/server/server-resources-subscribe-legacy` | 1/2 | 1 | ... |
```

**After (this PR):**

```
**Score: 1/2 passed** (1 expected failures, 0 unexpected failures, ...)

| Suite | Score | Expected failures | ... |
| `2025-11-25/server/server-resources-subscribe-legacy` | 1/2 | 1 | ... |

**Obsolete expected-failure entries.** These matched no failing or warning check
in this run, so they are no longer describing a real outcome.

- `resources-unsubscribe`
```

Identical scores, identical table, one extra section. No exit-code change, so it cannot turn a green run red.

Why it matters: the file currently has eleven entries and nothing checks that they still correspond to anything. When a suite is renamed or a gap is closed, the entry stays and keeps classifying that check as "expected" forever. The next time that check genuinely breaks, the run stays green and nobody hears about it.

If that is not a problem you want reported, or you would rather it hard-fail than print a section, say which and I will do that. Equally fine to close it, no hard feelings, it is a small quality-of-life thing rather than a bug fix.

khandrew1:
@ayaangazali, yeah i actually don't think we need to report these. these are expected on purpose as it's a growing pain with the migration to stateless. perhaps later on but for now i'm going to close this one out. Thanks!

ayaangazali:
Makes sense, and that is a better reason than any of the ones I was weighing. If the entries are expected to churn while the stateless migration is in flight, then reporting them as obsolete is noise pointed at the wrong thing.

Thanks for taking the time to say why rather than just closing it.

PR #2272, opened 2026-08-16
feat(oauth): add Scalekit provider

### LINKED ISSUE #2271: Add Scalekit as a first-class OAuth provider
## Summary

This issue tracks adding Scalekit as a first-class mcp-use OAuth provider, next to Auth0, Clerk, WorkOS, Supabase, Keycloak, and Better Auth.

Scalekit is an OAuth 2.1 authorization server for MCP. It supports Dynamic Client Registration and client id metadata documents. A resource server should only **verify** Scalekit-issued access tokens. It should not hold a client secret or depend on `@scalekit-sdk/node`.

PR: will be linked as soon as it is opened from `saif-at-scalekit/mcp-use`.

## Why this is a new provider, not `oauthCustomProvider` docs

Two Scalekit-specific checks are easy to get wrong if each app copies a verifier:

1. **Issuer is a pair, not a single URL.** Authorization-server metadata advertises the resource-scoped issuer (`{environmentUrl}/resources/{resourceId}`). Live access tokens currently use the **environment-root** issuer. A verifier that only accepts one form will break during Scalekit's issuer migration. The factory advertises the resource-scoped issuer and verifies both.
2. **Audience is the resource id (`res_…`), not the MCP URL.** A token minted for a different resource in the same environment must be rejected. A Server URL may also appear in `aud`; it does not replace the resource-id check.

JWKS is `{environmentUrl}/keys` (from live authorization-server metadata). `{environmentUrl}/resources/{resourceId}/keys` is a 404.

If a `resource` claim is present, it must be the public MCP URL. Live tokens often omit it; absence is accepted.

## Proposed API

```ts
import { MCPServer } from "mcp-use";
import { oauthScalekitProvider } from "mcp-use/oauth/scalekit";

  name: "scalekit-server",
  version: "1.0.0",
  oauth: oauthScalekitProvider({
    environmentUrl: process.env.SCALEKIT_ENVIRONMENT_URL!,
    resourceId: process.env.SCALEKIT_RESOURCE_ID!,
  }),
});
```

`ctx.auth.user` is `{ id, subjectType: "user" | "machine", organizationId?, sessionId? }`. `subjectType` is `"machine"` when `sub` equals `client_id` / `azp`.

`issuerBoundAccessTokens: true` is rejected. Scalekit binds tokens to the resource id, not to sharing an environment issuer.

## Scope

Provider-only:

- Factory, JWT/JWKS verification, type wiring, tests on the existing direct-provider seam
- Provider pages in mcp-use docs (same shape as Auth0 / Clerk)

Out of scope:

- Changing how mcp-use hosts well-known documents
- A Python provider
- A Scalekit cookbook (that lives on docs.scalekit.com after this ships on npm)

Happy to discuss the two verification rules here.

saif-at-scalekit:
## Summary

Adds Scalekit as a first-class mcp-use OAuth provider (`oauthScalekitProvider`), same shape as Auth0 / Clerk.

Closes #2271.

## Why two non-obvious checks

Please don't "simplify" these back to a single issuer or a URL-only audience. They are the reason this is a first-class provider.

1. **Verifier issuers are a pair.** Advertised `oauthMetadata.issuer` is the resource-scoped URL (`{environmentUrl}/resources/{resourceId}`). Live access tokens currently use the environment-root issuer. `createJwtVerifier` now accepts `issuer: string | string[]` so both verify during Scalekit's issuer migration.
2. **Audience is the resource id (`res_…`).** A token minted for another resource in the same environment is rejected. A Server URL may also appear in `aud`; it does not replace the resource-id check.

JWKS is `{environmentUrl}/keys`. `{environmentUrl}/resources/{resourceId}/keys` is a 404.

If a `resource` claim is present, it must match the public MCP URL. Absence is accepted.

`issuerBoundAccessTokens: true` is rejected. No Scalekit SDK and no client secret.

## What this PR does

- `src/oauth/scalekit.ts` — factory, metadata (DCR + CIMD), caller map (`id`, `subjectType`, optional org/session)
- Tests on the existing direct-provider and type-wiring seams (minted JWT, stubbed JWKS)
- `mcp-use/oauth/scalekit` export (package.json, tsup, typedoc)
- Provider pages in TypeScript + v2 docs, plus skill references

This PR does **not** change how mcp-use hosts well-known documents.

## Test plan

- [x] `pnpm exec tsc -p tsconfig.test.json` in `packages/server`
- [x] `pnpm exec vitest run tests/oauth-direct-providers.test.ts` (16 passing, including Scalekit construction + JWT cases)
- [x] Reviewer: confirm the two verification rules above are intentional

---
## Summary by cubic
Adds `oauthScalekitProvider` to `mcp-use` and tightens token verification. Old: single issuer and optional audience; New: verifier accepts both environment-root and resource-scoped issuers, `aud` must include the `res_…` resource id (any configured `audience` is an extra required value), and `resource` (when present) must match the MCP URL.

- Exposes `mcp-use/oauth/scalekit` with env fallbacks `MCP_USE_OAUTH_SCALEKIT_ENVIRONMENT_URL` and `MCP_USE_OAUTH_SCALEKIT_RESOURCE_ID`; no Scalekit SDK or client secret.
- `JwtVerifierOptions.issuer` now accepts `string | string[]`; JWKS loads from `{environmentUrl}/keys`; protected-resource metadata advertises the resource-scoped issuer and supports DCR + CIMD.
- Maps caller claims: `sub` → `user.id`; `subjectType` is "machine" only when `sub` equals `client_id` or `azp`; optional `org_id` → `organizationId`, `sid` → `sessionId`; `permissions` surfaced; other claims remain in `ctx.auth.payload`.
- Adds a Scalekit provider docs page.

**Migration**
- Configure `environmentUrl` and a `res_…` `resourceId` (env or options).
- If you set `audience`, ensure tokens include both it and the `res_…` id.
- Ensure your MCP URL equals the token’s `resource` claim when present.
- Remove any use of `issuerBoundAccessTokens`.

<sup>Written for commit e33ac22fd44da543d7c686bef43e819708a8668a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2272?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


libraries/typescript/packages/server/src/oauth/scalekit.ts
```diff
@@ -0,0 +1,203 @@
...
+}
+
+function mapUser(authInfo: AuthInfo) {
+  const payload = payloadFromAuthInfo(authInfo);
+  const id = requiredString(payload, "sub");
+  if (id === undefined) throw invalidToken("Missing Scalekit subject");
+  const clientId =
+    requiredString(payload, "client_id") ?? requiredString(payload, "azp");
+  const subjectType: ScalekitOAuthUser["subjectType"] =
+    clientId !== undefined && id === clientId ? "machine" : "user";
```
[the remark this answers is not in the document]
saif-at-scalekit: Leaving the classification as `sub === client_id || sub === azp`.

Live Scalekit user tokens carry both `sub: usr_…` and `client_id: m2m_…` (the host app). Presence of `client_id`, or absence of `sid`/`org_id`, would mark those callers as `machine`. That is the failure mode this field exists to prevent.

A client_credentials token is machine when `sub` is the client. That is the OAuth subject convention and the locked mapping for this provider. Tests cover:

- `sub=usr_…`, `client_id=m2m_…` → `user`
- `sub === client_id` → `machine`

I documented that assumption on `mapUser` and in the v2 provider page. I do not have a live client_credentials dump that omits `client_id`/`azp`. If Scalekit issues that shape, we should add a dedicated claim then, not weaken the user-token case.

docs/typescript/server/authentication/providers/scalekit.mdx
```diff
@@ -0,0 +1,130 @@
+---
+title: "Scalekit Provider"
+description: "Configure Scalekit MCP Auth for a TypeScript MCP server."
```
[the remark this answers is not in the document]
saif-at-scalekit: Fixed in f82f836.

Removed the new v1 page, the v1 index bullet, and the v1 `docs.json` nav entry. The v2 provider page stays.

The existing v1 API reference update in `docs/typescript/api-reference/server/auth-providers.mdx` stays. That file already lists the other providers. The throw-type sentence there is corrected in the same commit.

libraries/typescript/packages/server/src/oauth/scalekit.ts
```diff
@@ -0,0 +1,203 @@
...
+    environmentUrl,
+    "environmentUrl"
+  ).href.replace(/\/$/, "");
+
+  const resourceId = normalizeResourceId(
+    options.resourceId ??
+      oauthEnvironmentValue("MCP_USE_OAUTH_SCALEKIT_RESOURCE_ID")
+  );
+  const resourceIssuer = `${environmentIssuer}/resources/${resourceId}`;
+  const audience = options.audience?.trim() ?? resourceId;
```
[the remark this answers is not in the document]
saif-at-scalekit: Fixed in f82f836.

`createJwtVerifier` always gets `audience: resourceId`. Optional `audience` is a second check after verify. Both values must be in `aud`.

jose treats an audience array as OR, so `[resourceId, extra]` would accept a token minted for another `res_…` that only carried the extra URL. The extra check is AND for that reason.

docs/typescript/api-reference/server/auth-providers.mdx
```diff
@@ -197,6 +197,54 @@ server.tool(
...
 
+### oauthScalekitProvider
+
+Create a Scalekit OAuth provider for an MCP server.
+
+```ts
+import { oauthScalekitProvider } from "mcp-use/oauth/scalekit";
+```
+
+Builds an `OAuthProvider` backed by Scalekit MCP Auth. Advertised metadata uses the resource-scoped issuer. The verifier accepts both the environment-root issuer and the resource-scoped issuer, and binds audience to the `res_…` resource id. `environmentUrl` falls back to `MCP_USE_OAUTH_SCALEKIT_ENVIRONMENT_URL` and `resourceId` falls back to `MCP_USE_OAUTH_SCALEKIT_RESOURCE_ID`. Throws `Error` if `environmentUrl` is missing, and `TypeError` if `resourceId` is missing, empty, or does not start with `res_`, if `audience` is empty, or if `issuerBoundAccessTokens` is `true`.
```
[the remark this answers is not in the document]
saif-at-scalekit: Fixed in f82f836.

Missing `environmentUrl` or `resourceId` throws `Error`. Empty/`res_`-prefix `resourceId`, empty `audience`, and `issuerBoundAccessTokens: true` throw `TypeError`. The API reference sentence now matches that.

libraries/typescript/packages/server/src/oauth/scalekit.ts
```diff
@@ -0,0 +1,228 @@
...
+   * Default verification only requires `resourceId` in `aud`. Set this when
+   * tokens also carry a Server URL and you want both values present. This
+   * does not replace the resource-id check.
+   */
+  audience?: string;
+  /**
+   * Rejected when `true`. Scalekit tokens are bound by resource id, not by
+   * sharing an environment issuer.
+   */
+  issuerBoundAccessTokens?: boolean;
```
AkshayParihar33: i think this param is not needed saif, we will handle it on server side as we will have an opt-in button for devs to tell us if they want issuerBoundAccessTokens or not. can we remove it please?

libraries/typescript/packages/server/src/oauth/scalekit.ts
```diff
@@ -0,0 +1,228 @@
...
+  const resourceId = value.trim();
+  if (!resourceId.startsWith("res_")) {
+    throw new TypeError(
+      `resourceId must start with "res_" (${resourceId}). A wrong value weakens audience binding.`
+    );
+  }
+  return resourceId;
+}
+
+function metadata(
```
AkshayParihar33, line 169: i dont think the mcp framework should expose the authorization server metadata. this is served by Scalekit AS (server-side)

libraries/typescript/packages/server/src/oauth/scalekit.ts
```diff
@@ -0,0 +1,228 @@
...
+/** Verified Scalekit caller exposed to authenticated MCP callbacks. */
+export interface ScalekitOAuthUser {
+  /** Token `sub`. A person under authorization_code; the client under client_credentials. */
+  id: string;
+  /** Distinguishes a person from a machine client so tools do not store a client id as a user id. */
+  subjectType: "user" | "machine";
+  /** Organization identifier from the access token's `org_id` claim. */
+  organizationId?: string;
+  /** Session identifier from the access token's `sid` claim. */
+  sessionId?: string;
```
AkshayParihar33, line 34: we allow custom claims also to be baked in the token - would you like to expose that as well - it's useful for devs. not a mandatory comment, please feel free to ignore it but its a nice to have

although one thing to keep in mind, the custom claims are not specific to user entity, it can any map<string, string> value that will help the mcp server overall for the given call. If we do decide, not to add custom claims, what's the alternative on how the MCP server can read all the jwt claims and map by themselves because i understand we want to keep this object simple and non-bloated

libraries/typescript/packages/server/src/oauth/scalekit.ts
saif-at-scalekit: Removed in cc1cc47.

Scalekit binds tokens to the resource id. The verifier never used issuer-bound access tokens. The dashboard opt-in can stay on the Scalekit side.

libraries/typescript/packages/server/src/oauth/scalekit.ts
saif-at-scalekit: Scalekit serves the live AS document. Agreed.

mcp-use `oauthCustomProvider` still requires `oauthMetadata`. The SDK uses `issuer` for `authorization_servers` on RFC 9728 protected-resource metadata. It also copies this blob to `/.well-known/oauth-authorization-server` on the MCP process.

That copy is a framework requirement, not a Scalekit one. Every other mcp-use provider does the same. Dropping it needs a change in mcp-use, not only in this adapter.

I left the object and added a comment that Scalekit is the live source.

libraries/typescript/packages/server/src/oauth/scalekit.ts
saif-at-scalekit: Leaving custom claims off `ctx.auth.user`.

`mapAuthInfo` already returns the full verified JWT as `payload`. Tools read it as `ctx.auth.payload`. That map is `Record<string, unknown>`, so a custom claim is `ctx.auth.payload.your_claim`.

The user object stays identity only (`id`, `subjectType`, `organizationId`, `sessionId`). Documented that split on the v2 provider page in cc1cc47.

khandrew1 (approved):
Thanks @saif-at-scalekit! I just did some cleanup regarding documentation.

Issue #2274, opened 2026-08-16
skills document an OAuth Proxy mode built on oauthProxy and jwksVerifier, neither of which exists in v2


ayaangazali:
## What happens

Both skills bundles document an "OAuth Proxy" mode built on two exports that do not exist in v2.

`skills/mcp-builder/references/authentication/auth0.md:110` and `skills/chatgpt-app-builder/references/authentication/auth0.md:110`:

```ts
import { MCPServer, oauthProxy, jwksVerifier, object } from "mcp-use/server";
```

and `custom.md:155` in both bundles:

```ts
import { oauthProxy, jwksVerifier } from "mcp-use/server";
```

Neither symbol exists anywhere in the v2 packages. `git grep oauthProxy -- libraries/typescript/packages/*/src` only matches the unrelated `oauthProxyAllowLoopback` option in the CLI, and `jwksVerifier` has no matches at all. The module they are imported from does not resolve either:

```
mcp-use/server -> ERR_PACKAGE_PATH_NOT_EXPORTED
```

`auth0.md` presents this as one of two supported modes, with a full worked example under `## Mode 2: OAuth Proxy (oauthProxy)`, and `custom.md` points readers at it for "full `oauthProxy` options". An agent following either produces code that fails at import.

## Why I am not fixing it in a PR

This is not a stale import path. The whole mode is v1 API surface, so the options are to port it to the v2 equivalent, or delete the mode from both bundles. Both are your call, and deleting a documented capability is not something I want to do on my own read.

It also overlaps the four `oauth/auth0-proxy` links still open on #2210, which point at the runnable example for exactly this mode. Whatever you decide here probably settles those too.

I have sent #2275, which fixes the 14 import lines whose symbols do genuinely exist and only had the wrong subpath. It deliberately leaves these four lines alone.

khandrew1:
Thanks for the report! I just merged in #2286 which was a full rewrite of all the skills to make them easier to use and handled all of these.

PR #2275, opened 2026-08-16
docs(skills): import from the subpaths that actually exist

ayaangazali:
### The defect

Every authentication reference in both skills bundles imports from `mcp-use/server`, which is not an entry point:

```
$ node -e "import('mcp-use/server')"
ERR_PACKAGE_PATH_NOT_EXPORTED
```

The package exposes `.`, `./react`, `./vite-client`, `./landing`, `./oauth`, `./oauth/{clerk,auth0,workos,supabase,keycloak,better-auth}`, `./node` and `./next`. There is no `./server`.

These files are instructions an agent copies verbatim, so the generated server fails at its first import. The rest of each page is already v2-correct, which is what makes the import line stand out: the env var names match the v2 contract, the providers are called with no arguments the way the env fallbacks expect, and the runnable-example links were repointed at `packages/server/examples/auth/...` in #2211.

### The fix

`MCPServer` and `object` come from `mcp-use`, and each provider from its own oauth subpath:

```ts
import { MCPServer, object } from "mcp-use";
import { oauthClerkProvider } from "mcp-use/oauth/clerk";
```

14 import lines across 14 files, one per provider page in each bundle.

### Verified

I resolved every import in these pages against the built package rather than reading them off:

```
mcp-use:                    MCPServer=OK object=OK
mcp-use/oauth:              oauthCustomProvider=OK
mcp-use/oauth/clerk:        oauthClerkProvider=OK
mcp-use/oauth/auth0:        oauthAuth0Provider=OK
mcp-use/oauth/better-auth:  oauthBetterAuthProvider=OK
mcp-use/oauth/keycloak:     oauthKeycloakProvider=OK
mcp-use/oauth/supabase:     oauthSupabaseProvider=OK
mcp-use/oauth/workos:       oauthWorkOSProvider=OK
```

After the change, every import line in these pages resolves with every symbol present, except the four noted below.

### What I left alone, deliberately

Four lines still import `oauthProxy` and `jwksVerifier` from `mcp-use/server`. Those symbols do not exist anywhere in v2, so there is no correct subpath to move them to. That is a whole documented mode built on v1 API, and porting or deleting it is your call, so I raised it separately in #2274 rather than quietly rewriting or dropping it here.

---
## Summary by cubic
Fixes auth example imports in both skills bundles to use `mcp-use` subpaths that exist. Previously examples imported from `mcp-use/server` and failed with ERR_PACKAGE_PATH_NOT_EXPORTED; now they import `MCPServer`/`object` from `mcp-use` and providers from `mcp-use/oauth/...`, so copied snippets run.

- Updated 14 import lines across provider pages in both bundles; verified each resolves against the built `mcp-use` (`oauth/{auth0,better-auth,clerk,keycloak,supabase,workos}` and `oauth` for custom).
- Left four `oauthProxy`/`jwksVerifier` lines unchanged; v2 has no equivalents. Tracked in #2274.

**Migration**
- If you copied earlier examples, replace `import ... from "mcp-use/server"` with `import { MCPServer, object } from "mcp-use"` and `import { oauthXProvider } from "mcp-use/oauth/x"`.

<sup>Written for commit 6f58158bbaf39f920789fbcb7418e12d56ffe9f8. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2275?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/skills/chatgpt-app-builder/references/authentication/auth0.md
+++ b/skills/chatgpt-app-builder/references/authentication/auth0.md
@@ -36,7 +36,8 @@ Auth0's built-in DCR feature is currently in **Early Access**. MCP clients regis
 ### Quick Start
 
 ```typescript
-import { MCPServer, oauthAuth0Provider, object } from "mcp-use/server";
+import { MCPServer, object } from "mcp-use";
+import { oauthAuth0Provider } from "mcp-use/oauth/auth0";
 
 const server = new MCPServer({
   name: "my-server",
--- a/skills/chatgpt-app-builder/references/authentication/better-auth.md
+++ b/skills/chatgpt-app-builder/references/authentication/better-auth.md
@@ -87,7 +87,8 @@ You need three things beyond the standard `MCPServer` setup:
 3. Mount login and consent pages
 
 ```typescript
-import { MCPServer, oauthBetterAuthProvider } from "mcp-use/server";
+import { MCPServer } from "mcp-use";
+import { oauthBetterAuthProvider } from "mcp-use/oauth/better-auth";
 import { auth } from "./auth.js";
 import {
   oauthProviderAuthServerMetadata,
--- a/skills/chatgpt-app-builder/references/authentication/clerk.md
+++ b/skills/chatgpt-app-builder/references/authentication/clerk.md
@@ -9,7 +9,8 @@ Setting up OAuth with Clerk. DCR mode only — MCP clients register themselves d
 ## Quick Start
 
 ```typescript
-import { MCPServer, oauthClerkProvider, object } from "mcp-use/server";
+import { MCPServer, object } from "mcp-use";
+import { oauthClerkProvider } from "mcp-use/oauth/clerk";
 
 const server = new MCPServer({
   name: "my-server",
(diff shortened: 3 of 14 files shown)


ayaangazali:
Extra evidence: the repo already considers this a defect. `skills/mcp-apps-builder/scripts/check-v2.mjs` has, as its first rule:

```js
[/from\s+["']mcp-use\/server["']/g, "import server APIs from mcp-use"],
```

Running it over the two bundles reports that import 9 times each, 18 total, which is exactly the set this PR looked at. It fixes the 14 whose symbols exist; the 4 left are the `oauthProxy` and `jwksVerifier` lines from #2274, which have no v2 equivalent to point at.

The checker lives in `mcp-apps-builder` and nothing runs it against the other two bundles, which is why this went unnoticed. Wider fallout from that is in #2278.

khandrew1:

PR #2185, opened 2026-08-07
fix(skills): use ctx.auth.user.id in mcp-apps-builder auth reference

fuleinist:
## Summary

Fixes #2184 — the shipped `mcp-apps-builder` skill's auth reference taught agents to read `ctx.auth.user.userId`, but no v2 OAuth provider exposes `userId`.

Verified against the source: all six built-in providers (`auth0`, `clerk`, `workos`, `supabase`, `keycloak`, `better-auth`) declare `id: string` on the user shape, and `grep -c userId` across `libraries/typescript/packages/server/src/oauth` returns zero matches.

## Impact

This is an **authorization path**: an agent following the doc wrote an ownership check against `undefined`, which reads as working while silently failing closed. The file ships with every scaffolded project (`SKILLS_SPARSE_PATH = "skills/mcp-apps-builder"` in `packages/create-mcp-use-app/src/skills-config.ts`).

## Change

One-token fix in `skills/mcp-apps-builder/references/auth.md:37`:

```diff
-await deleteOwnedDocument(documentId, ctx.auth.user.userId);
+await deleteOwnedDocument(documentId, ctx.auth.user.id);
```

## Scope

Deliberately limited to the shipped `mcp-apps-builder` skill, per the issue's note that the same `userId` pattern in `skills/mcp-builder` and `skills/chatgpt-app-builder` (plus other v1 leftovers) is tracked separately.

Closes #2184

---
## Summary by cubic
Updated the `mcp-apps-builder` auth reference to use `ctx.auth.user.id` (not `userId`) to match v2 OAuth providers. Prevents ownership checks from reading `undefined` and silently failing in authorization paths.

<sup>Written for commit 6e4fc3d69fe4d8a0f617751bf49f48bbb028c8a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2185?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

Issue #2184, opened 2026-08-07
mcp-apps-builder skill tells agents to read ctx.auth.user.userId, which v2 does not expose


ayaangazali:
`skills/mcp-apps-builder/references/auth.md:37` teaches this:

```ts
await deleteOwnedDocument(documentId, ctx.auth.user.userId);
```

`userId` is not a field on any v2 auth user. All six built-in providers expose `id`:

```
$ git grep -E "^\s+(id|userId)\??:" -- libraries/typescript/packages/server/src/oauth/{auth0,clerk,workos,supabase,keycloak,better-auth}.ts
auth0        id: string;
clerk        id: string;
workos       id: string;
supabase     id: string;
keycloak     id: string;
better-auth  id: string;

$ git grep -c userId -- libraries/typescript/packages/server/src/oauth
(no matches)
```

So an agent following this skill writes an ownership check against `undefined`. In the example above that is an authorization path, which is the worst place for it to read as working.

This one matters more than the same mistake elsewhere in `skills/`, because `mcp-apps-builder` is the skill `create-mcp-use-app` actually ships: `SKILLS_SPARSE_PATH = "skills/mcp-apps-builder"` in `packages/create-mcp-use-app/src/skills-config.ts`. Every scaffolded project gets this file.

Fix is `ctx.auth.user.id`. Happy to send that PR if useful.

For the record, the same `userId` pattern appears across `skills/mcp-builder` and `skills/chatgpt-app-builder` too, along with other v1 leftovers. I have kept that out of this issue since @khandrew1 mentioned opening a general skills issue, and this one is scoped to the shipped skill so it can be fixed independently of whatever happens to those.

ayaangazali:
Same defect exists in 18 more places, across `chatgpt-app-builder` and `mcp-builder`: the `clerk`, `keycloak`, `supabase`, `workos`, `better-auth` and `overview` pages, plus the DCR mode and user-info section of `auth0`. Sent #2277 for those.

I deliberately did not touch `mcp-apps-builder/references/auth.md`, so #2185 still applies cleanly and there is nothing to conflict with.

Also left `custom.md` and the `auth0` Mode 2 section alone: both still document the v1 provider API, so `userId` may be correct for what they describe, and changing one field there would make them look ported when they are not.

khandrew1:
Thanks for the report! I just merged in #2286 which was a full rewrite of all the skills to make them easier to use and handled all of these.

PR #2277, opened 2026-08-16
docs(skills): read ctx.auth.user.id, not userId

ayaangazali:
Extends #2184, which reported this in `mcp-apps-builder`. The same defect is in 18 more places across the other two bundles.

### The defect

The built-in provider user types expose `id`. `userId` is not a field on any of them, so these snippets do not compile. Checked against the built package:

```ts
object({ userId: ctx.auth.user.id, ... })     // compiles
object({ userId: ctx.auth.user.userId, ... }) // error TS2339: Property 'userId' does not exist on type 'ClerkOAuthUser'.
```

`git grep userId -- libraries/typescript/packages/server/src/oauth/` returns nothing, and the Clerk user type for example is `id`, `email`, `name`, `username`, `picture`, `emailVerified`, `organizationId`, `organizationRole`, `organizationSlug`, `roles`.

Only the read changes. The response key stays `userId`, since that is this tool's output field and not tied to the provider type.

### What I left alone, and why

Three exclusions, all deliberate:

- **`mcp-apps-builder/references/auth.md`** is #2185's scope. I did not touch it, so there is nothing to conflict with when that merges.
- **`custom.md` in both bundles** still documents the v1 provider shape, `verifyToken` plus `getUserInfo`, rather than v2's `createTokenVerifier`, `oauthMetadata` and `mapAuthInfo`. There the user object is whatever `getUserInfo` returns, so `userId` may well be right for what that page describes. Changing one field inside an otherwise v1 example would make it look ported when it is not.
- **The `## Mode 2: OAuth Proxy` section of `auth0.md`**, for the same reason: it is built on `oauthProxy` and `jwksVerifier`, which do not exist in v2 at all. That is #2274.

So this covers the pages that are already on the v2 API, where the provider is called the v2 way and the only thing left wrong was the field name.

Pages touched: `clerk`, `keycloak`, `supabase`, `workos`, `better-auth`, `overview`, and the DCR mode plus the user-info section of `auth0`, in both `chatgpt-app-builder` and `mcp-builder`.

---
## Summary by cubic
Corrects docs examples to read ctx.auth.user.id instead of ctx.auth.user.userId so TypeScript snippets compile with v2 provider types. Previously examples read userId from ctx.auth.user; now they read id while still returning a response field named userId.

- Updates provider docs in `chatgpt-app-builder` and `mcp-builder` (Auth0, BetterAuth, Clerk, Keycloak, Supabase, WorkOS, and overview). No runtime or API changes; docs only.
- Leaves v1-only pages unchanged: `custom.md` in both bundles and the "Mode 2: OAuth Proxy" section in `auth0.md`. Also leaves `mcp-apps-builder/references/auth.md` for its own fix.

<sup>Written for commit 8be356b6f37f83aaf12424334e5e21a5f20f63fb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2277?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/skills/chatgpt-app-builder/references/authentication/auth0.md
+++ b/skills/chatgpt-app-builder/references/authentication/auth0.md
@@ -48,7 +48,7 @@ server.tool(
   { name: "whoami", description: "Get authenticated user info" },
   async (_args, ctx) =>
     object({
-      userId: ctx.auth.user.userId,
+      userId: ctx.auth.user.id,
       email: ctx.auth.user.email,
       permissions: ctx.auth.permissions,
     })
@@ -165,7 +165,7 @@ server.tool(
   { name: "get-user-info", description: "Get authenticated user info" },
   async (_args, ctx) =>
     object({
-      userId: ctx.auth.user.userId,
+      userId: ctx.auth.user.id,
       email: ctx.auth.user.email,
       name: ctx.auth.user.name,
       nickname: ctx.auth.user.nickname,
--- a/skills/chatgpt-app-builder/references/authentication/better-auth.md
+++ b/skills/chatgpt-app-builder/references/authentication/better-auth.md
@@ -182,7 +182,7 @@ server.tool(
   { name: "get-user-info", description: "Get information about the authenticated user" },
   async (_args, ctx) =>
     object({
-      userId: ctx.auth.user.userId,
+      userId: ctx.auth.user.id,
       email: ctx.auth.user.email,
       name: ctx.auth.user.name,
       scopes: ctx.auth.scopes,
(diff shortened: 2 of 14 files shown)


khandrew1:

PR #2288, opened 2026-08-17
docs: correct server environment CLI reference

### LINKED ISSUE #2283: CLI reference documents a servers env surface the CLI does not implement, and strict parseArgs makes every example fail
## What happens

The `servers env` section of the CLI reference documents a command surface the CLI does not implement. Every example in it fails at argument parsing.

**Documented** (`docs/v2/typescript/api-reference/cli-reference.mdx:488-491`): subcommands `list`/`ls`, `add`, `update`, `remove`/`rm`, with `--server <id>` as a required option plus `--show-values`, `--env`, `--sensitive`, `--no-sensitive` and `--value`.

**Implemented** (`libraries/typescript/packages/cli/src/commands/servers.ts:226-230`):

```ts
if (operation === "list") return envList(argv.slice(1), json);
if (operation === "set") return envSet(argv.slice(1), json);
if (operation === "unset") return envUnset(argv.slice(1), json);
throw new UsageError("Usage: mcp-use servers env <list|set|unset>");
```

So `add`, `update`, `remove`, `ls` and `rm` do not exist. The real verbs are `set` and `unset`.

The options do not line up either. `envList` accepts only the common org/json options plus `branch`, and takes the server as a **positional**, not `--server`:

```ts
options: { ...commonOrgJsonOptions(), branch: { type: "string" } },
...
const server = exactlyOne(positionals, "mcp-use servers env list <server>");
```

`envSet` adds exactly one more option, and it is named `secret`, not `sensitive`:

```ts
secret: { type: "boolean" },
...
```

`--show-values`, `--value` and `--env` have no counterpart at all. Grepping the whole workspace, `--show-values`, `--sensitive`, `--no-sensitive` and `--value` appear in zero files.

## Why this is worse than a wording mismatch

Every one of those `parseArgs` calls sets `strict: true`, so an unrecognised flag throws instead of being ignored. That makes each documented example fail immediately rather than partially work:

```
mcp-use servers env add --server srv_abc123 API_KEY=secret --sensitive
```

fails on `add` before it even reaches the flags, and the `list` example fails on `--server`.

The equivalent working invocation today looks like:

```
mcp-use servers env list srv_abc123
mcp-use servers env set srv_abc123 API_KEY=secret --secret
```

## What I need from you

I did not send a PR because I cannot tell which side is meant to be authoritative. Either the page documents a CLI you are planning or shipping elsewhere, in which case the implementation is what needs to change, or the page has drifted, in which case I will rewrite the section to match `servers.ts` and check the rest of the file the same way.

Say which and I will do it. If it is the docs, I would also fold in the other flags I found documented but absent anywhere in the workspace: `--runtime` and `--sort`.

Tethys0:
Closes #2283

## Changes
- document the implemented `servers env list`, `set`, and `unset` verbs with positional server arguments
- remove unsupported aliases and options from the server and environment command tables
- clarify metadata-only listing, preview branch scope, `--secret`, and confirmation behavior

## Verification
- `pnpm build`
- `pnpm --filter @mcp-use/cli test:run` (289 tests passed)
- built CLI `servers env` help output for root, list, set, and unset commands
- `pnpm exec prettier --check ../../docs/v2/typescript/api-reference/cli-reference.mdx`
- `git diff --check`

Documentation-only change; no changeset required.

---
## Summary by cubic
Aligns `servers env` CLI docs with `@mcp-use/cli`. Switches to positional `<server>` args and `list`/`set`/`unset`, documents `--org` for env commands, and removes unsupported aliases/options.

- Servers table: drop `ls`/`rm` aliases and `--sort`; `delete` has no alias.
- Env commands: `list` returns metadata only; add `--org` and `--branch`; `set` supports `--secret`; `unset` prompts unless `--yes`. Remove `--env` and `--show-values`.
- Examples updated to use positional `<server>` and new verbs.
- Only `docs/v2/typescript/api-reference/cli-reference.mdx` changed; verify tables and examples match CLI help for `servers`, `servers env`, and env subcommands.

<sup>Written for commit d6e7d9a00d9d5b8b7a6bc8c337af297bf7a8cc79. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2288?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/v2/typescript/api-reference/cli-reference.mdx
+++ b/docs/v2/typescript/api-reference/cli-reference.mdx
@@ -485,9 +485,9 @@ mcp-use servers env <subcommand> [options]
 
 | Subcommand                 | Purpose                                   | Required arguments       | Other options                 |
 | -------------------------- | ----------------------------------------- | ------------------------ | ----------------------------- |
-| `list <server>`            | List environment variable metadata.       | Server ID or slug        | `--branch <name>`             |
-| `set <server> <KEY=VALUE>` | Create or update an environment variable. | Server ID or slug, value | `--branch <name>`, `--secret` |
-| `unset <server> <key>`     | Delete an environment variable by key.    | Server ID or slug, key   | `--branch <name>`, `--yes`    |
+| `list <server>`            | List environment variable metadata.       | Server ID or slug        | `--org <id-or-slug>`, `--branch <name>`             |
+| `set <server> <KEY=VALUE>` | Create or update an environment variable. | Server ID or slug, value | `--org <id-or-slug>`, `--branch <name>`, `--secret` |
+| `unset <server> <key>`     | Delete an environment variable by key.    | Server ID or slug, key   | `--org <id-or-slug>`, `--branch <name>`, `--yes`    |
 
 **Examples:**
 
--- a/docs/v2/typescript/api-reference/cli-reference.mdx
+++ b/docs/v2/typescript/api-reference/cli-reference.mdx
@@ -445,13 +445,13 @@ mcp-use servers <subcommand> [options]
 
 **Subcommands:**
 
-| Subcommand                              | Purpose                                     | Options                                                                                                                                                                                                                                                |
-| --------------------------------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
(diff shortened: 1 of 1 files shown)


khandrew1 (approved):

khandrew1:
Thanks for the PR!

PR #2289, opened 2026-08-18
fix(inspector): add a test script so pnpm test covers this package

ayaangazali:
### The defect

The workspace test script is recursive:

```json
"test": "pnpm run -r test"
```

`@mcp-use/inspector` only defined `test:unit`, and pnpm skips a package with no matching script without saying anything. So the documented `pnpm test` omitted the inspector's unit tests entirely:

```
$ pnpm --filter @mcp-use/inspector test
(no output, exit 0)

$ pnpm --filter @mcp-use/inspector test:unit
Test Files  45 passed (45)
Tests  217 passed (217)
```

217 tests, silently absent from the run contributors are told to use in `CONTRIBUTING.md`.

CI is not affected. Its step calls `test:unit` directly, which is what #2201 wired up. This is only about the workspace-level command.

### The change

One line, `"test": "vitest run"`, which is what `test:unit` already does.

I used run mode rather than aliasing to a watcher on purpose. `test` currently means different things across the workspace: `vitest` in agent, client and mcp-use, `vitest run` in tunnel, `vitest --run --passWithNoTests` in create-mcp-use-app, and `node --test ... && vitest run` in cli. Since `pnpm run -r test` executes packages in series, a watch-mode entry can hold the whole recursive run open in an interactive terminal, so the run-mode form is the safe one to add. Matching `tunnel` seemed the most consistent choice.

### Verified

```
$ pnpm --filter @mcp-use/inspector test
Test Files  45 passed (45)
Tests  217 passed (217)
```

Preflight clean, 0 failing packages.

---
## Summary by cubic
Adds a `test` script to `@mcp-use/inspector` so workspace `pnpm test` includes its unit tests instead of silently skipping them. Previously only `test:unit` existed; `test` now runs `vitest` in run mode to avoid blocking recursive runs.

- Adds `"test": "vitest run"` to `@mcp-use/inspector/package.json`.
- `pnpm test` now runs 217 inspector unit tests; CI stays the same (invokes `test:unit`).

<sup>Written for commit ed5ea5077dd3082e8ae2df083656c6c9922a2505. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2289?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/plain-moons-listen.md
+++ b/libraries/typescript/.changeset/plain-moons-listen.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/inspector": patch
+---
+
+Add a `test` script so the workspace-wide `pnpm test` includes the inspector's unit tests. `pnpm run -r test` skipped the package silently, since only `test:unit` existed.
--- a/libraries/typescript/.changeset/plain-moons-listen.md
+++ b/libraries/typescript/.changeset/plain-moons-listen.md
@@ -1,5 +1,0 @@
----
-"@mcp-use/inspector": patch
----
-
-Add a `test` script so the workspace-wide `pnpm test` includes the inspector's unit tests. `pnpm run -r test` skipped the package silently, since only `test:unit` existed.
--- a/libraries/typescript/packages/inspector/package.json
+++ b/libraries/typescript/packages/inspector/package.json
@@ -72,6 +72,7 @@
     "test:e2e:codegen": "playwright codegen http://localhost:3000/inspector",
     "test:e2e:report": "playwright show-report",
     "test:e2e:python": "node tests/e2e/scripts/run-python-e2e.mjs",
+    "test": "vitest run",
     "test:unit": "vitest run",
     "test:unit:watch": "vitest",
     "test:e2e:tunnel": "node tests/e2e/scripts/run-tunnel-test.mjs",


ayaangazali:
Pre-existing on `canary`, not from this change, so I am not pushing a fix.

This PR adds one line to `package.json`, a `test` script alias. The failing job runs `test:unit`, which is untouched.

**The failure:**

```
❯ tests/unit/compressed-assets.test.ts (1 test | 1 failed) 10029ms
```

10029ms against a 10000ms cap, so it missed by 0.3 percent.

**Same failure on `canary` itself, twice today**, with no PR involved. Run 32053629834 (job 95459015144):

```
❯ tests/unit/compressed-assets.test.ts (1 test | 1 failed) 10230ms
```

and run 32051250836 failed the same way. The job passed on `canary` on the 13th, so this started recently and is timing-dependent rather than broken.

**Why it keeps happening:** the inspector is the only package capped at 10s. `agent`, `client`, `cli` and `server` all use 60000. These cases mount a server and decompress the bundled app, which is the slow end of the suite, so they sit right at the edge of a budget six times stricter than everything else. The margins tell the story: 10029ms here, 10230ms on canary, 10837ms on main last week.

#2268 raises it to 60s and is green, but it targets `main`, and both branches still carry `testTimeout: 10000`, so `canary` will keep failing until that lands and forward-merges. Happy to open the same one-line change against `canary` if you would rather not wait for the merge-forward.

ayaangazali:
Correcting one thing in my description. I justified using run mode partly by saying a watch-mode entry "can hold the whole recursive run open in an interactive terminal". I went back and tested that under a pty rather than leaving it as an assumption, and it does not happen:

```
script -q /dev/null pnpm --filter mcp-use test     # test = "vitest"
Duration  4.29s
```

No watch prompt, process exited. So `pnpm test` at the root terminates today and there is no hang to avoid.

The change itself is unaffected: the inspector had no `test` script at all, so `pnpm run -r test` skipped 217 tests silently, and `vitest run` is still the right value since it matches what `test:unit` already does. I just did not want to leave an unverified claim in the rationale.

ayaangazali:
Rebased onto current `canary` to get a fresh run rather than re-arguing from a three-day-old one. New head `b1e130d7`. Same result:

```
Tests  1 failed | 246 passed (247)
```

That is `compressed-assets.test.ts` against the inspector's 10s cap, on current `canary`, with a suite that has grown from 218 to 247 tests since I opened this. `canary` still has `testTimeout: 10000`.

This PR adds one line to `package.json`, a `test` alias. The failing job runs `test:unit`, which it does not touch, and the same test fails on `canary` itself with no PR involved (runs 32053629834 at 10230ms and 32051250836).

#2290 is the one-line fix for this on `canary` and #2268 is its `main` twin. Both are green and open. Nothing further I can do from this branch.

ayaangazali:
Green now: 29 checks, 0 failures. Rebased onto `canary` after #2290 merged, which raised the inspector cap to 60s, so the `compressed-assets` timeout that had been failing here is gone. Thanks for taking that one.

This PR is unrelated to the timeout and is unchanged: the inspector had no `test` script, so `pnpm run -r test` at the workspace root skipped the package silently and 217 unit tests never ran in the documented `pnpm test`. One line, `"test": "vitest run"`, matching what `test:unit` already does.

Ready when you are.

khandrew1 (approved):

khandrew1:
Thanks @ayaangazali! And just so you know, changesets are only needed when the package code changes and needs to be versioned. Things like scripts don't need one

ayaangazali:
Good to know, thanks. I will skip the changeset on script and config only changes from now on.

PR #2308, opened 2026-08-21
fix(client): isolate useMcp connection lifecycles

### LINKED ISSUE #1696: useMcp: isolate connection lifecycle state instead of guarding stale disconnects

khandrew1:
## Summary

- give each useMcp connection Effect a fresh client, captured OAuth provider, and explicit lifecycle owner
- prevent retired connection attempts, cleanup, notifications, inventories, OAuth callbacks, health checks, and refresh operations from mutating a newer lifecycle
- route retry, OAuth completion, proxy fallback, and health recovery through a coalesced Effect-driven reconnect signal
- preserve manual disconnect and current header dependency behavior

Closes #1696

## Test plan

- pnpm run build in packages/client
- pnpm exec tsc --noEmit -p packages/client/tsconfig.json
- pnpm exec vitest run tests/unit/react
- pnpm exec vitest run

## Validation

- React tests: 16 files, 94 tests passed
- Full client suite: 56 files, 351 tests passed
- Client build and TypeScript checks passed

---
## Summary by cubic
Isolates `useMcp` connection lifecycles to remove race conditions during React re-renders. Previously, stale connection attempts and cleanups could overwrite or disconnect a newer connection; now each Effect owns its client and OAuth provider, and only the current lifecycle can publish state.

- Introduces a lifecycle owner in `useMcp`, with a fresh client and captured `UseMcpAuthProvider` per Effect; retired lifecycles cannot mutate current state (cleanup, OAuth callbacks, proxy fallback, health checks, inventories, refreshes).
- Routes retry, OAuth completion, proxy fallback, and health recovery through a single reconnect signal; prevents spurious authorization prompts from retired connections.
- Preserves manual disconnect and existing header dependency behavior; no breaking API changes expected.
- Adds focused lifecycle race tests; fixes Linear #1696; publishes a patch to `@mcp-use/client`.
- Migration: none.

<sup>Written for commit 3babffa15c35e4bc78c70816114207d8be0a0b30. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2308?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/fix-usemcp-lifecycle-races.md
+++ b/libraries/typescript/.changeset/fix-usemcp-lifecycle-races.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Give each `useMcp` connection Effect its own client lifecycle so stale connection attempts, cleanup, OAuth callbacks, proxy fallback, health checks, and inventory refreshes cannot overwrite or disconnect a newer connection.
--- a/libraries/typescript/packages/client/src/react/useMcp-lifecycle.ts
+++ b/libraries/typescript/packages/client/src/react/useMcp-lifecycle.ts
@@ -1,0 +1,212 @@
+import type { OAuthClientProvider } from "@modelcontextprotocol/client";
+import type { BrowserMCPClient } from "../core/browser.js";
+import type { MCPConnection } from "../core/session.js";
+
+/**
+ * The OAuth provider shape used by useMcp.  This lives here because a provider
+ * belongs to one connection lifecycle: replacing the provider (for example
+ * after an OAuth callback) must not change an already-running lifecycle.
+ */
+export type UseMcpAuthProvider = OAuthClientProvider & {
+  tokens?: () => Promise<
+    | {
+        access_token?: string;
+        token_type?: string;
+        refresh_token?: string;
+        scope?: string;
+        [key: string]: unknown;
+      }
+    | undefined
+  >;
+  clearStorage?: () => number;
+  getLastAttemptedAuthUrl?: () => string | null | undefined;
+  getTokenEndpoint?: () => Promise<string | null>;
+  getResource?: () => Promise<string | null>;
+  getClientCredentials?: () => Promise<{
+    client_id: string;
+    client_secret?: string;
+  } | null>;
+  getProxyFetch?: (baseFetch?: typeof fetch) => typeof fetch | undefined;
+  serverUrl?: string;
+  getKey?: (keySuffix: string) => string;
+  serverUrlHash?: string;
+};
+
+/** A lifecycle's transport state, including states in which no session exists. */
+export type ConnectionLifecyclePhase =
+  | "connecting"
+  | "waiting_for_auth"
+  | "authenticating"
+  | "ready"
+  | "failed"
+  | "retiring"
+  | "retired";
+
+/** Why this lifecycle was started. Primarily useful for coalescing and logs. */
+export type ConnectionLifecycleTrigger =
+  | "configuration"
+  | "manual-retry"
+  | "auto-retry"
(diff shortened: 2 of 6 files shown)


khandrew1:
`useMcp` most likely needs a better rewrite. this approach isn't that good though, so closing out as an experiment

PR #2309, opened 2026-08-22
feat(server): add oauthConvexProvider

### LINKED ISSUE #2299: Add Convex oauth adapter
Conves seems getting popular, we can add the integration
https://www.convex.dev/components/codefox-inc/oauth-provider

ayaangazali:
Closes #2299.

Follows the design you settled on: a single `authURL` that includes the mounted base path, no separate `basePath`.

```ts
import { oauthConvexProvider } from "mcp-use/oauth/convex";

const oauth = oauthConvexProvider({
  authURL: "https://example.convex.site/oauth",
});
```

That URL is also the issuer, which is what makes the single-value approach right rather than merely simpler. The component's own docs identify its tokens by string-matching the issuer:

```ts
if (identity?.issuer?.includes("/oauth")) {
    // This is an OAuth token
}
```

`providerEndpoint` keeps the prefix, so everything resolves under the base.

### Taken from the component, not guessed

- **Endpoints**, from the routes it tells you to mount: `authorize`, `token`, `register`, `userinfo`, `.well-known/jwks.json`.
- **Claims**: `sub`, `cid` for the registered client, `scp` for granted scopes, plus `name` and `picture` from the `profile` scope and `email` and `email_verified` from `email`.

### DCR

Off by default there, and MCP clients need it to self-register, so the TSDoc says to enable it. I still advertise `registration_endpoint`, since suppressing it would hide the capability from clients on correctly configured deployments.

### Verified

Built the package and exercised the real export:

```
issuer:                 https://example.convex.site/oauth
jwks_uri:               https://example.convex.site/oauth/.well-known/jwks.json
registration_endpoint:  https://example.convex.site/oauth/register
oauthConvexProvider({})                     -> "Convex authURL is required."
oauthConvexProvider({ authURL: "http://..." }) -> rejected, HTTPS or localhost
```

One test added to `oauth-direct-providers.test.ts`, alongside the equivalent Better Auth case: it signs an EdDSA token and asserts both the metadata and the claim mapping, including that a trailing slash on `authURL` still yields the exact issuer. **17/17 in that file.**

Worth flagging: `tsup.config.ts` needed the `oauth/convex` entry too. Without it the new `./oauth/convex` export would have resolved to a file the build never emits, so I confirmed `dist/oauth/convex.js` and `dist/oauth/convex.d.ts` are produced.

---
## Summary by cubic
Adds `oauthConvexProvider` at `mcp-use/oauth/convex` to verify tokens from the Convex OAuth Provider component. It uses a single `authURL` that includes the mounted base path; that URL is the issuer and anchors all endpoints.

- Export is wired in `package.json` and `tsup.config.ts` so `dist/oauth/convex.{js,d.ts}` are emitted.
- Configure via `authURL` or `MCP_USE_OAUTH_CONVEX_AUTH_URL`; rejects non-HTTPS (except localhost) and normalizes trailing slashes.
- Advertises `authorize`, `token`, `register`, `userinfo`, and `.well-known/jwks.json` under the base; default scopes are `openid`, `profile`, `email`.
- Maps claims: `sub`→`id`, `cid`→`clientId`, `scp`→permissions, plus `name`, `picture`, `email`, `email_verified`→`emailVerified`.
- Convex Dynamic Client Registration is off by default; enable it to allow clients to self-register via the `registration_endpoint`.

<sup>Written for commit 7ce735848e2d5541fa431b3ccc53cd1bb1ba4e20. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2309?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ayaangazali:
This one is a fork limitation, so I am not pushing a fix.

The conformance suite itself passes. Of 39 checks on this head, 38 are green including:

```
success | TypeScript Conformance Tests (v2; legacy + modern modes)
failure | Report current-head conformance scores
```

So the scores were produced correctly. What fails is only the job that publishes them, and specifically its last step:

```
2 Download Python v1 results     => success
3 Post Python v1 score           => skipped
4 Download TypeScript v2 results => success
5 Post TypeScript v2 score       => failure
```

The download works, the write-back does not. That job declares:

```yaml
    permissions:
      contents: read
      pull-requests: write
```

and `GITHUB_TOKEN` is read-only for pull requests from forks regardless of what a workflow requests, so the step cannot post the score. I hit the same thing on #2234, where the log showed it plainly as `RequestError [HttpError]: Resource not accessible by integration`.

Nothing in this PR can affect it either way: it adds a new provider file, an exports entry, a tsup entry, one test and a changeset, and touches no workflow or conformance fixture.

If it is worth silencing, the fix belongs in the workflow rather than here, for example skipping the posting steps when the PR comes from a fork:

```yaml
if: github.event.pull_request.head.repo.full_name == github.repository
```

Happy to send that separately, since it would affect every outside contributor's checks rather than just this branch.

ayaangazali:
Rebased onto current canary. It was 18 commits behind, and the drift turned up a gap in my own PR worth fixing.

**`src/oauth/convex.ts` was missing from the package's typedoc `entryPoints`.** All seven existing providers are listed there; mine was not, even though `./oauth/convex` is a public `exports` entry. So the new provider would have shipped as a public export the docs site never renders, and after #2312 landed today it would also have escaped the coverage gate entirely.

Added it, and checked both directions rather than assuming the entry does anything:

```
pnpm docs:api                              exit 0
pnpm docs:api, TSDoc removed from
ConvexOAuthUser                            exit 3
  oauth/convex.ConvexOAuthUser (Interface) does not have any documentation
```

So the entry is live, not decorative.

Also re-ran the provider tests against current canary after 18 commits of drift, since the oauth surface moved in that window: **17 passed, 17 total**. Preflight clean.

`Report current-head conformance scores` is still red and still not fixable from here: that job needs `pull-requests: write`, which `GITHUB_TOKEN` never has on a fork PR. Every fork PR fails it and every internal one passes. #2310 is the fix, and the trust-boundary question I left there is the other half.

khandrew1:
@ayaangazali could you show a video of the full integration through the inspector? would be very helpful, thank you

ayaangazali:
I cannot record a video from where I work, so rather than send nothing, here is the same walkthrough as a transcript you can re-run yourself. Say the word if a screen recording is genuinely required and I will find a way.

A server wired with nothing but the provider:

```js
  name: "convex-demo",
  version: "1.0.0",
  oauth: oauthConvexProvider({ authURL: "https://demo.convex.site/oauth" }),
});
```

**Discovery, derived entirely from that one URL:**

```
GET /.well-known/oauth-authorization-server

{
  "issuer": "https://demo.convex.site/oauth",
  "authorization_endpoint": "https://demo.convex.site/oauth/authorize",
  "token_endpoint": "https://demo.convex.site/oauth/token",
  "registration_endpoint": "https://demo.convex.site/oauth/register",
  "jwks_uri": "https://demo.convex.site/oauth/.well-known/jwks.json",
  "userinfo_endpoint": "https://demo.convex.site/oauth/userinfo",
  "response_types_supported": ["code"],
  "grant_types_supported": ["authorization_code", "refresh_token"],
  "code_challenge_methods_supported": ["S256"],
  "scopes_supported": ["openid", "profile", "email"]
}
```

Note the base path is preserved throughout, which is the thing you steered me to on #2299: one `authURL` including the path, no separate `basePath`. Convex's own token check is `identity?.issuer?.includes("/oauth")`, so the issuer has to carry it.

**Protected-resource metadata and the challenge the Inspector follows:**

```
POST /mcp   (no Authorization header)
HTTP/1.1 401 Unauthorized
www-authenticate: Bearer error="invalid_token",
  error_description="Missing Authorization header",
  resource_metadata="http://localhost:39511/.well-known/oauth-protected-resource/mcp"

GET /.well-known/oauth-protected-resource/mcp
{
  "resource": "http://localhost:39511/mcp",
  "authorization_servers": ["https://demo.convex.site/oauth"]
}
```

That is the full chain the Inspector walks: 401 with a `resource_metadata` pointer, resource metadata naming the authorization server, and the AS document advertising `registration_endpoint` so a client can register itself.

**Token verification**, which the transcript above cannot show against a fake host, is covered in `tests/oauth-direct-providers.test.ts`: it signs a real EdDSA JWT, serves a JWKS, and asserts the mapped claims. 17 passed, 17 total on current canary. A bad bearer token against the live server returns 401 as expected.

Everything above runs against this branch's build, including the `dist/oauth/convex.js` entry.

khandrew1:
@ayaangazali `demo.convex.site` isn't a real website and there could always be edge cases when testing with the real provider. Can you confirm you tested this provider with a real Convex project?

ayaangazali:
Straight answer: **no. I have not tested this against a real Convex deployment**, and I should have said that up front instead of leading with a transcript against a hostname I made up. You are right to push on it.

What is actually verified:

- **Token verification against a real JWKS.** `tests/oauth-direct-providers.test.ts` runs a local `oauth2-mock-server`, signs an EdDSA JWT, and asserts the provider fetches the key set, verifies the signature, enforces the issuer, and maps `sub`, `email`, `cid` and `scp`. That is real crypto and a real key rotation path, just not Convex's server.
- **Metadata derivation** from a single `authURL`, and the 401 plus `resource_metadata` chain, which is what the transcript showed.

What is not verified, and is exactly where a real project could bite:

- That Convex serves `authorize`, `token`, `register`, `userinfo` and `.well-known/jwks.json` at those paths under the mounted base. I took those from Convex's own docs rather than from a live deployment.
- That the `iss` claim on a real Convex token matches the `authURL` including its base path. Their docs check `identity?.issuer?.includes("/oauth")`, which is what the design rests on, but I have not seen a real token.
- Whether their DCR registration response shape matches what the SDK expects.

I cannot create a Convex deployment from where I run, so I cannot close that gap myself. Reasonable options, your call: someone with a Convex project runs the example against it, or this waits until one of you can, or you close it and I stop maintaining it. I would rather it sit unmerged than land a provider that nobody has pointed at the real thing.

khandrew1:
@ayaangazali It's best that you test this PR yourself as you are the one who wrote and opened it, so that you can continue to fix it in case you run into any issues. Testing it with a dummy URL doesn't prove the validity of your work.

ayaangazali:
That is a fair standard and I agree with it. I tried to meet it before answering.

- **Self-hosted Convex backend.** `get-convex/convex-backend` runs under Docker, which would give me a real deployment to point at. The Docker CLI exists on this machine but no daemon is available and I cannot start Docker Desktop from here.
- **Convex cloud.** Needs an account, and I am not going to create third-party accounts on my own initiative.

So I cannot do what you are asking, and you are right that a dummy URL does not prove the provider works. Rather than leave a PR you would have to babysit, **I am closing this one.**

The branch stays at `ayaangazali:feat/convex-oauth-provider` if it is useful to anyone with a Convex project. What is genuinely verified is the crypto and the shape: a real EdDSA token against a live JWKS, issuer enforcement, claim mapping for `sub`/`email`/`cid`/`scp`, metadata derived from a single `authURL` with the base path preserved, and the 401 plus `resource_metadata` chain. What is unverified is everything that depends on Convex actually behaving the way its docs describe: the endpoint paths under the mounted base, the real `iss` claim, and the DCR response shape.

I will also drop my assignment on #2299 so it is open for someone who can point it at a live deployment.

Thanks for holding the line on this. Better to close it than to have it merged on evidence that does not support it.

Issue #2273, opened 2026-08-16
cli dev tests flake on main through waitFor, and the 24x gap suggests a port race rather than a slow runner


ayaangazali:
## What happens

`tests/cli/dev.test.ts` fails intermittently on `main`, on both runners, always through the same helper.

Windows, run 31920530746 (job 95099711160):

```
FAIL tests/cli/dev.test.ts > runDev > reloads skill files when the configured skills directory overlaps views
Error: waitFor timed out after 15000ms
```

Ubuntu, run 31749127452 (job 94610964438), four cases in one run:

```
FAIL tests/cli/dev.test.ts > runDev (views) > serves Vite modules through a public sandbox hostname
FAIL tests/cli/dev.test.ts > runDev (views) > shuts down with active MCP subscriptions and HMR WebSockets
FAIL tests/cli/dev.test.ts > runDev (views) > serves view documents, virtual entries, and reloads on view add
FAIL tests/cli/dev.test.ts > runDev (views) > reconciles adjacent server and view events as one project generation
Error: waitFor timed out after 15000ms; last error: TypeError: fetch failed
```

## The part that does not add up

Locally the whole file is 34 tests in 15.3s, all passing, and the case that failed on Windows takes **628ms**. The slowest case in the file is 2912ms. So a 628ms wait exceeding 15000ms on CI is roughly a 24x blowup, which is a lot more than the ~2x the rest of the file shows there (`dev.test.ts` reported 31185ms for the whole file on that Windows run).

That gap makes me think this is not simply a slow runner, so I did not want to send a timeout bump and call it fixed. Two candidates, and I cannot tell them apart without a Windows runner:

**1. The poll deadline is unrelated to the test budget.** `tests/cli/helpers.ts:136` defaults `waitFor` to 15000ms, while `vitest.config.ts` allows `testTimeout: 60000`. So a test that waits on server startup gets a quarter of its budget, and the failure surfaces as the helper's message rather than vitest's. Raising the default is free when things pass, since `waitFor` returns as soon as the probe succeeds.

**2. `getFreePort` has a time-of-check/time-of-use race**, `tests/cli/helpers.ts:161`. It binds a random candidate, closes it, and returns the number for the caller to bind later:

```ts
const server = createServer();
server.once("error", () => resolve(false));
server.listen({ port: candidate, host }, () => {
  server.close(() => resolve(true));
});
```

[...]

PR #2311, opened 2026-08-22
ci: run the TypeScript jobs when ci.yml itself changes

ayaangazali:
## Why

`detect-changes` gates every TypeScript job on the `typescript` path filter, and that filter only matches `libraries/typescript/**`. A pull request that edits nothing but `.github/workflows/ci.yml` matches no filter at all, so `typescript` is `false` and every job the workflow defines is skipped on the very change that defines them.

The practical effect is that a CI fix cannot be verified by its own CI run.

## Evidence

The three most recent workflow-only fixes in `main` are all of the form "this test was never running". Each one merged without the job it was wiring up ever executing:

| PR | what it fixed | the relevant check on its own run |
|----|---------------|-----------------------------------|
| #2201 | actually run the inspector unit tests | `typescript/inspector` **skipped** |
| #2214 | run the create-mcp-use-app unit tests | `typescript/create-app-tests` **skipped**, `typescript/create-mcp-use-app/build` **skipped** |
| #2213 | actually run the mcp-use unit tests | no TypeScript job ran at all |

So the class of bug those PRs fixed is exactly the class their own CI could not catch.

## The change

Add `.github/workflows/ci.yml` to the `typescript` and `create-mcp-use-app` filters, and correct the two overview comments that describe when those groups run.

For a `pull_request` event GitHub uses the pull request's own copy of the workflow, so the jobs that run are the edited ones. That makes this PR self-checking: if the change works, the TypeScript matrix runs here even though no file under `libraries/typescript/` was touched.

## Scope

I left the `python` filter alone. All three demonstrated failures above are TypeScript-side, and the same line can be added there if you want the symmetry. Say the word and I will add it.

---
## Summary by cubic
Runs the TypeScript and `create-mcp-use-app` jobs when `.github/workflows/ci.yml` changes, so workflow-only edits validate the jobs they define. Previously these jobs ran only when files under `libraries/typescript/**` changed; now they also run on workflow file edits.

- Adds `.github/workflows/ci.yml` to the `typescript` and `create-mcp-use-app` path filters.
- Updates the overview comments to match; the `python` filter stays unchanged.
- On `pull_request` runs, GitHub uses the PR's own workflow, so the edited jobs execute here.

<sup>Written for commit 81f9bc4b49fda23d071bc969ba062626591f3a31. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2311?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -26,7 +26,7 @@ permissions:
 #    ├─ python-integration-tests: Other integration tests
 #    └─ python-agent-tests: Matrix [4 agent test suites] (requires API keys)
 #
-# 3. TypeScript Tests (runs if libraries/typescript/** changed)
+# 3. TypeScript Tests (runs if libraries/typescript/** or this workflow changed)
 #    ├─ typescript-lint: ESLint checks
 #    ├─ typescript-build: Build all packages
 #    ├─ typescript-knip: Unused code/deps check (fails CI on findings)
@@ -37,7 +37,7 @@ permissions:
 #    ├─ typescript-release-channel-test: release-channel helper suite (main + canary PRs)
 #    └─ typescript-changeset-check: PR validation (PRs only)
 #
-# 4. create-mcp-use-app Tests (runs if packages/create-mcp-use-app/** changed)
+# 4. create-mcp-use-app Tests (runs if that package or this workflow changed)
 #    ├─ create-mcp-use-app-build: Build, unit test, and pack the CLI
 #    └─ create-app-tests: Matrix [3 OS × 3 PM × 3 templates × 2 flags]
 #       - Tests project creation across Ubuntu/macOS/Windows
@@ -103,11 +103,13 @@ jobs:
               - 'libraries/python/**'
             typescript:
               - 'libraries/typescript/**'
+              - '.github/workflows/ci.yml'
             knip:
               - 'libraries/typescript/**'
               - '.github/workflows/ci.yml'
             create-mcp-use-app:
               - 'libraries/typescript/packages/create-mcp-use-app/**'
+              - '.github/workflows/ci.yml'
             release-channel:
               - 'libraries/typescript/scripts/release-channel.mjs'
               - 'libraries/typescript/scripts/release-channel.test.mjs'


ayaangazali:
One more piece of evidence I found after opening this: including the workflow's own file in its trigger paths is already the convention here. Every other path-filtered workflow does it.

| workflow | lists itself |
|---|---|
| `conformance.yml` | yes |
| `everything-typecheck.yml` | yes |
| `server-examples.yml` | yes |
| `sync-inspector.yml` | yes |
| `typescript-pr-preview.yml` | yes |
| `typescript-release.yml` | yes |
| `dependabot-changesets.yml` | yes |
| `ci.yml` | **no** |

So this is `ci.yml` being the one place the existing convention was not applied, rather than a new policy.

ayaangazali:
Extended this to the other half of the same gap.

`approve-fork-pr.yml` recomputes the same three flags from the PR's file list and passes them to `ci.yml` as `workflow_call` inputs:

```js
const typescript = paths.some(p => p.startsWith('libraries/typescript/'));
```

So a promoted fork PR that only edits `ci.yml` reports `typescript=false` and skips every job it changes, exactly like the direct path did. Fixing only the path filter would have left the approved-fork path broken.

cubic independently flagged the caller-side half of this on #2179 while I was working on it, which is what sent me looking.

Both detectors now treat a change to `.github/workflows/ci.yml` as a TypeScript change. The self-check still holds: the TypeScript matrix is running on this PR, which touches no file under `libraries/typescript/`.

ayaangazali:
Self-check is in and green: **45 successful checks** on a PR that touches no file under `libraries/typescript/`. The full TypeScript matrix ran, including `typescript-build`, `typescript-knip`, and the four package test jobs.

The live contrast is #2179, which is also a `ci.yml`-only change and does not have this fix on its branch: 21 skipped, 5 successes, no TypeScript job at all.

Same kind of change, same day, one of them actually tested.

.github/workflows/approve-fork-pr.yml
```diff
@@ -64,9 +64,13 @@ jobs:
...
             const paths = files.map(f => f.filename);
 
+            // Matches the path filters in ci.yml: a change to the workflow itself
+            // has to run the jobs it configures.
+            const ciWorkflow = paths.includes('.github/workflows/ci.yml');
+
             const python = paths.some(p => p.startsWith('libraries/python/'));
-            const typescript = paths.some(p => p.startsWith('libraries/typescript/'));
-            const create_mcp_use_app = paths.some(p => p.startsWith('libraries/typescript/packages/create-mcp-use-app/'));
+            const typescript = ciWorkflow || paths.some(p => p.startsWith('libraries/typescript/'));
```
khandrew1: I do not think this fixes the approved-fork case as described. The run-ci job calls `./.github/workflows/ci.yml`, which GitHub resolves from the same commit as the caller workflow on `main`; `inputs.ref` only controls which source commit the jobs check out. For a fork PR that changes only `ci.yml`, setting these outputs to true therefore runs the existing trusted TypeScript workflow against otherwise unchanged TypeScript source. It does not execute or validate the modified workflow, while triggering the full TypeScript and create-app matrix. Could we remove the `ciWorkflow` override here and keep the self-triggering behavior only in `ci.yml`? The normal secretless `pull_request` run is the safe place to execute and validate a fork-modified workflow. GitHub documents the local reusable-workflow resolution behavior here: https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#calling-a-reusable-workflow

khandrew1 (changes requested):
Requesting one change: please remove the approved-fork override that treats a ci.yml-only PR as a TypeScript source change. Keeping the self-trigger in ci.yml for normal pull_request runs makes sense, but the approved-fork caller cannot validate the modified workflow.

ayaangazali:
Rebased. This went conflicted when #2179 merged, since both touch the `detect-changes` filter block.

Resolved by keeping both, which is what you want here: #2179's `release-channel` filter and this PR's two `ci.yml` entries are independent.

```yaml
            typescript:
              - 'libraries/typescript/**'
              - '.github/workflows/ci.yml'
            create-mcp-use-app:
              - 'libraries/typescript/packages/create-mcp-use-app/**'
              - '.github/workflows/ci.yml'
            release-channel:
              - 'libraries/typescript/scripts/release-channel.mjs'
              ...
```

New head `6c2c3df`, same two commits, and the overview comments this PR corrects picked up #2179's new `typescript-release-channel-test` line cleanly. Mergeable again.

ayaangazali:
Rebased again, this time on top of #2331. Both changes kept, `52c9256` -> new head, mergeable.

```yaml
            typescript:
              - 'libraries/typescript/**'
              - '.github/workflows/ci.yml'
            knip:
              - 'libraries/typescript/**'
              - '.github/workflows/ci.yml'
```

Worth noting that #2331 reached the same conclusion this PR is built on, from a different direction: "give Knip its own path filter so CI workflow changes exercise this job". That is exactly the gap here, just solved for one job. This PR closes it for the rest of the TypeScript matrix and for the `create-mcp-use-app` group, plus the approved-fork caller in `approve-fork-pr.yml`, which recomputes the same flags independently and would otherwise still skip everything on a workflow-only PR.

If you would rather do it the #2331 way and give each job its own filter, that works too and I am happy to close this. The thing I care about is that a change to `ci.yml` runs the jobs it configures, which is currently only true for `knip` and `release-channel`.

khandrew1:
@ayaangazali Checking in, did you see my review comments?

.github/workflows/approve-fork-pr.yml
ayaangazali: Done, and you are right about the mechanism.

I checked the docs link you gave and the behaviour is exactly as you describe: for a local reusable workflow, GitHub resolves `./.github/workflows/ci.yml` from the same commit as the caller, so `run-ci` executes the trusted `main` copy. `inputs.ref` only decides what the jobs check out. My override therefore could not have validated a fork-modified workflow. It would have run the unmodified trusted workflow against unchanged TypeScript source, and paid the full TypeScript and create-app matrix to learn nothing.

That is a straight error on my part. I reasoned about `inputs.ref` and never checked how the reusable workflow itself is resolved.

Removed the `ciWorkflow` override, so `approve-fork-pr.yml` is untouched by this PR now. What is left is only the `ci.yml` self-trigger for normal `pull_request` runs, which is the secretless path where a fork-modified workflow genuinely does execute:

```yaml
            typescript:
              - 'libraries/typescript/**'
              - '.github/workflows/ci.yml'
            create-mcp-use-app:
              - 'libraries/typescript/packages/create-mcp-use-app/**'
              - '.github/workflows/ci.yml'
```

Rebased onto current main, one commit, new head `81f9bc4`.

ayaangazali:
@khandrew1 no, I did not, and that is my fault rather than a GitHub notification problem. Your review landed on the 25th and my sweeps for new review activity used rolling cutoffs that happened to start just after it, so it fell in the gap between two checks. Three days is a long time to leave a review sitting and I am sorry for the wait.

I have replied on the thread and pushed the change: the `approve-fork-pr.yml` override is gone, only the `ci.yml` self-trigger remains.

ayaangazali:
Same Windows flake as #2350 and #2273, not this branch. Nothing pushed for it.

```
 ❯ Module.waitFor tests/cli/helpers.ts:149:9
 ❯ tests/cli/dev.test.ts:422:5

typescript/cli (ubuntu-latest)   success
typescript/cli (windows-latest)  failure
```

No `; last error:` suffix again, so the probe never threw. This PR's whole diff is four added lines in `ci.yml` path filters.

There is a small irony worth naming: the CLI job only runs on this PR **because** of the change this PR makes. Before it, a `ci.yml`-only edit matched no filter and the entire TypeScript matrix was skipped, which is the bug it fixes. So the first thing the fix bought us is a real Windows failure we would otherwise never have seen on a workflow-only change.

khandrew1 (approved):

PR #2312, opened 2026-08-22
docs(typescript): enforce the TypeDoc coverage gate the packages already declare

ayaangazali:
## Why

`packages/{server,client,agent}/typedoc.json` each set:

```json
"validation": { "notExported": true, "invalidLink": true, "notDocumented": true },
"treatWarningsAsErrors": true
```

and `eslint.config.js` says so explicitly:

> Documentation coverage is enforced against each package's public entrypoints by its strict TypeDoc configuration.

It is not enforced. The only invocation anywhere is `pnpm docs:api` (`docs.yml:68`), which runs typedoc against the root `typedoc.options.json` in `entryPointStrategy: "packages"` mode. The sub-package `validation` settings are applied in that mode, but `treatWarningsAsErrors` is not, and the root options never set it.

On `canary` before this change:

```
$ pnpm docs:api
[warning] Found 0 errors and 15 warnings
$ echo $?
0
```

Fifteen coverage violations, exit 0. Adding the flag on the command line and changing nothing else gives exit 3, which confirms the flag is the only thing missing.

Two of the fifteen are in `mcp-use` itself, whose `CLAUDE.md` states that every public export "including public type properties" must have TSDoc. They merged anyway, because nothing checks.

## The change

Set `treatWarningsAsErrors` in the root options, then clear what the gate catches:

- **Missing docs.** Members of `SkillSnapshotEntry.resources` (server), and of `SkillsListResult`, `SkillGetResult`, `SkillDirectoryEntry`, `SkillDirectoryReadResult` (client).
- **Stale allowlist entries.** Five `intentionallyNotDocumented` entries for `react.UseWidgetResult` and `react.UserAgent`, plus `UseWidgetResultBase` in `intentionallyNotExported`. `grep -rn "UserAgent\|UseWidgetResult" packages/server/src` returns nothing, so those symbols are gone and the entries were dead. That drift is itself a symptom: the lists could rot precisely because nothing validated them.
- **Two genuinely-internal types.** `TurbopackConfig` added to `intentionallyNotExported`, matching the existing `NextHeaderRule` entry for the same pattern in the same file, and `FinishOAuthAuthorization` on the client side.

## Verification

```
$ pnpm docs:api          # exit 0, 0 warnings
```

And the gate bites. Deleting one doc comment (`SkillGetResult.skill`) and re-running:

```
[warning] index.SkillGetResult.skill (Property), defined in @mcp-use/client/src/core/skills.ts, does not have any documentation
[warning] Found 0 errors and 1 warnings
exit 3
```

## Second commit

`.typedoc/` is gitignored but was not eslint-ignored, so running `pnpm docs:api` and then `pnpm lint` reports 534 errors out of the generated minified theme assets. `.turbo`, `.tsup`, `dist` and `build` are already in that ignore list; `.typedoc` was not. Worth folding in here, since this PR is the reason anyone would run `docs:api` locally.

## Notes

No changeset: the source edits are doc comments only, with no runtime or type change to any published API. (`typescript-changeset-check` skips canary-targeted PRs anyway.)

---
## Summary by cubic
Enforces TypeDoc coverage across all TS packages by treating warnings as errors at the root. Previously `pnpm docs:api` exited 0 with coverage warnings; now undocumented public APIs fail the build.

- Set `"treatWarningsAsErrors": true` in `libraries/typescript/typedoc.options.json`; `pnpm docs:api` exits non‑zero on coverage warnings.
- Add `packages/tunnel/typedoc.json` and document inline return members on `TunnelManager.start` and `TunnelManager.status`.
- Add missing TSDoc in `@mcp-use/client` and `@mcp-use/server` for `SkillsListResult`, `SkillGetResult`, `SkillDirectoryEntry`, `SkillDirectoryReadResult`, and `SkillSnapshotEntry.resources`.
- Prune stale allowlists in `packages/server/typedoc.json`; add `TurbopackConfig` to `intentionallyNotExported`. In `packages/client/typedoc.json`, add `FinishOAuthAuthorization` to `intentionallyNotExported` and mark `index.MCPConnection.listAllSkills.__type.skills` as intentionally not documented.
- Ignore generated `.typedoc/` in `eslint.config.js` to avoid lint errors after running `pnpm docs:api`.

<sup>Written for commit 01932951de1d5f0752c57c9c947c401c41cf269f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2312?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/eslint.config.js
+++ b/libraries/typescript/eslint.config.js
@@ -18,6 +18,7 @@ export default [
       "**/*.min.js",
       "**/.turbo/**",
       "**/.tsup/**",
+      "**/.typedoc/**",
       "packages/*/dist/**",
       "packages/*/build/**",
       "packages/*/node_modules/**",
--- a/libraries/typescript/packages/client/src/core/skills.ts
+++ b/libraries/typescript/packages/client/src/core/skills.ts
@@ -21,24 +21,32 @@ export interface Skill {
 
 /** Paginated result returned by `skills/list`. */
 export interface SkillsListResult {
+  /** Skills on this page, in server order. */
   skills: Skill[];
+  /** Cursor for the next page, absent on the final page. */
   nextCursor?: string;
 }
 
 /** Result returned by `skills/get`. */
 export interface SkillGetResult {
+  /** The requested skill. */
   skill: Skill;
 }
 
 /** One child returned by `resources/directory/read`. */
 export interface SkillDirectoryEntry {
+  /** Resource URI of the child. */
   uri: string;
+  /** Display name of the child when the server provides one. */
   name?: string;
+  /** MIME type of the child when the server provides one. */
   mimeType?: string;
 }
 
 /** Paginated result returned by `resources/directory/read`. */
 export interface SkillDirectoryReadResult {
+  /** Directory children on this page, in server order. */
   resources: SkillDirectoryEntry[];
+  /** Cursor for the next page, absent on the final page. */
   nextCursor?: string;
 }
--- a/libraries/typescript/packages/client/typedoc.json
+++ b/libraries/typescript/packages/client/typedoc.json
@@ -15,6 +15,7 @@
   "excludeExternals": true,
   "intentionallyNotExported": [
     "BaseMCPClient",
+    "FinishOAuthAuthorization",
     "BaseTelemetryEvent",
     "ConnectorInitEventData",
     "ElicitParams",
@@ -35,6 +36,7 @@
   ],
   "intentionallyNotDocumented": [
     "index.CODE_MODE_AGENT_PROMPT",
+    "index.MCPConnection.listAllSkills.__type.skills",
     "index.setTelemetrySource",
     "index.setProductVersion",
     "index.Telemetry.getInstance",
--- a/libraries/typescript/packages/server/src/skills/types.ts
+++ b/libraries/typescript/packages/server/src/skills/types.ts
@@ -43,7 +43,12 @@ export interface SkillSnapshotEntry {
   /** Verbatim YAML frontmatter rendered as JSON. */
   frontmatter: Record<string, unknown>;
   /** Complete, deterministic file set for this skill. */
-  resources: Array<{ uri: string; digest: string }>;
+  resources: Array<{
+    /** Resource URI of one file belonging to the skill. */
+    uri: string;
+    /** SHA-256 digest of that file's raw bytes. */
+    digest: string;
+  }>;
 }
 
 /** One directory resource retained for scoped directory reads. */
--- a/libraries/typescript/packages/server/typedoc.json
+++ b/libraries/typescript/packages/server/typedoc.json
@@ -63,12 +63,7 @@
     "react.ErrorBoundary.getDerivedStateFromError.__type.hasError",
     "react.RegisteredTools.__type.input",
     "react.RegisteredTools.__type.output",
-    "react.ToolError.result.__type.isError",
-    "react.UseWidgetResult.__type.isPending",
-    "react.UseWidgetResult.__type.props",
-    "react.UserAgent.capabilities.__type.hover",
-    "react.UserAgent.capabilities.__type.touch",
-    "react.UserAgent.device.__type.type"
+    "react.ToolError.result.__type.isError"
   ],
   "intentionallyNotExported": [
     "BaseServerConfig",
@@ -97,7 +92,7 @@
     "StripCompleteSuffix",
     "StripMcpPrefix",
     "ToolsFromModule",
-    "UseWidgetResultBase",
+    "TurbopackConfig",
     "ViewControlsProps"
   ],
   "externalSymbolLinkMappings": {
--- a/libraries/typescript/packages/tunnel/src/index.ts
+++ b/libraries/typescript/packages/tunnel/src/index.ts
@@ -138,11 +138,19 @@ export interface TunnelManager {
    * @param port - Bound local HTTP port to expose.
    * @returns The public origin and assigned tunnel identifier.
    */
-  start(port: number): Promise<{ url: string; subdomain: string }>;
+  start(port: number): Promise<{
+    /** Public origin the tunnel is reachable on. */
+    url: string;
+    /** Tunnel identifier assigned by the relay. */
+    subdomain: string;
+  }>;
   /** Stop the WebSocket relay and release its reservation. */
   stop(): Promise<void>;
   /** Current tunnel public origin URL, or `null` when inactive. */
-  status(): { url: string | null };
+  status(): {
+    /** Public origin while the tunnel is active, `null` otherwise. */
+    url: string | null;
+  };
 }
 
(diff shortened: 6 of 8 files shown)


ayaangazali:
`Report current-head conformance scores` is a fork limitation, not something this branch can fix, so I have not pushed anything speculative.

The suite itself passes. 20 checks green, and `TypeScript Conformance Tests (v2; legacy + modern modes)` is one of them. Only the job that writes the score back to the PR fails, and its annotation says why:

```
Unhandled error: HttpError: Resource not accessible by integration
```

That job asks for `pull-requests: write`, and `GITHUB_TOKEN` is read-only for pull requests from a fork no matter what the workflow requests. The same failure appears on every fork PR and on no internal one:

| PR | head repo | result |
|----|-----------|--------|
| #2312 | `ayaangazali/mcp-use` | failure |
| #2309 | `ayaangazali/mcp-use` | failure |
| #2272 | `saif-at-scalekit/mcp-use` | failure |
| #2308, #2282, #2262 | `mcp-use/mcp-use` | success |

This PR touches five files: three typedoc configs, one eslint ignore line, and doc comments in two source files. Nothing in it reaches a workflow or a conformance fixture.

#2310 is the fix for the red check itself, by skipping that job when the token cannot post.

ayaangazali:
Rebased onto current `canary` and added the fourth package, since leaving it out would have been an odd shape for a PR whose whole point is that the gate is not actually on.

`server`, `client` and `agent` each carry a `typedoc.json` with `validation.notDocumented` and `treatWarningsAsErrors`. `tunnel` sits in the same root `entryPoints` list and had **no config at all**, so switching the gate on would have left it as the one published package still unchecked.

Added the matching config and documented the three inline members it turns up:

```
TunnelManager.start.__type.url
TunnelManager.start.__type.subdomain
TunnelManager.status.__type.url
```

`pnpm docs:api` still exits 0 with zero warnings across all four packages, and the negative control still holds. Deleting the `subdomain` line alone:

```
[warning] TunnelManager.start.__type.subdomain (Property), defined in @mcp-use/tunnel/src/index.ts, does not have any documentation
exit 3
```

I kept this in the same PR rather than opening a second one, since it is the same change to the same mechanism and splitting it would just mean two reviews of one idea.

ayaangazali:
One more measurement that I think raises the stakes here, found while checking whether lint already covered this.

`jsdoc/require-jsdoc` is applied only to `packages/{server,cli,tunnel}`. `packages/{agent,client}` get `tsdoc/syntax` alone, and the config says why:

> Documentation coverage is enforced against each package's public entrypoints by its strict TypeDoc configuration.

I verified both halves rather than assuming. Appending an undocumented `export function` and `export interface` to a server source file gives two `jsdoc/require-jsdoc` errors, so lint really does cover declarations there. Running the same rule against client and agent as a probe reports **179** violations, which is why it is not switched on for them.

So for client and agent the TypeDoc gate was not one of two safety nets, it was the only one, and it was exiting 0. That is also why the nine missing-doc warnings this PR fixes were all on the client side.

ayaangazali:
The `typescript/cli (windows-latest)` red on this branch is not from this change. Not pushing anything speculative for it.

The failing assertion is a timeout, not a wrong value:

```
tests/cli/dev.test.ts > runDev > reloads skill files when the configured skills directory overlaps views
  ❯ Module.waitFor tests/cli/helpers.ts:149:9
  ❯ tests/cli/dev.test.ts:422:5
```

Note there is no `; last error:` suffix. `waitFor` appends that whenever the predicate throws, so the predicate never threw here, it simply never became true inside the window. That is the 15s default I raised in #2273, on the slowest runner in the matrix.

This diff cannot reach that code path. It changes three typedoc configs, one eslint ignore line, and doc comments inside two `interface` bodies:

```diff
-  resources: Array<{ uri: string; digest: string }>;
+  resources: Array<{
+    /** Resource URI of one file belonging to the skill. */
+    uri: string;
```

`SkillSnapshotEntry` and `TunnelManager` are interfaces, so they are erased at compile time. There is no runtime difference for the CLI dev server to observe.

Two more things that place it outside this branch. The same job passes on #2316, which sits on the same rebased `canary`, so it is not the four new canary commits. And the test passes here locally in **671ms**, against a 15000ms budget:

```
Test Files  1 passed (1)
     Tests  1 passed | 33 skipped (34)
  Duration  1.07s
```

Happy to rebase for a fresh run if that is useful, but the underlying decision is still the one waiting in #2273.

ayaangazali:
Closing the loop on the second red, since the head moved since I last wrote about it. `Report current-head conformance scores` is the fork-token limit again, not this branch.

Step breakdown on the current run:

```
2 Download Python v1 results     => success
3 Post Python v1 score           => skipped
4 Download TypeScript v2 results => success
5 Post TypeScript v2 score       => failure
```

The downloads work, only the write-back fails, and the annotation is the same one every fork PR gets:

```
Unhandled error: HttpError: Resource not accessible by integration
```

That job asks for `pull-requests: write`, which `GITHUB_TOKEN` never has on a pull request from a fork. `TypeScript Conformance Tests (v2; legacy + modern modes)` itself is **green** on this head, so the scores were produced fine. #2310 is the fix for the red itself.

So both reds here are outside the diff: this one structurally, and the Windows CLI one because it is the 15s `waitFor` in #2273 timing out on the slowest runner, which I covered above. 18 checks pass.

ayaangazali:
Took the rebase I offered, now that `canary` has moved (`#2319` merged and `#2320` exited prerelease mode).

Worth flagging for anyone else with a branch off `canary`: the release rewrote history, so a plain `git rebase canary` tried to replay canary's own release commits back onto my branch and hit conflicts in `server/package.json`, `inspector/package.json` and `server/CHANGELOG.md`. Replaying only my own commits with `git rebase --onto canary 80fec8f7` was clean.

New head `0193295`, three commits, same eight-file diff as before. `pnpm docs:api` still exits 0 with zero warnings against the new base.

This also gives the Windows CLI job a fresh run, which was the point: that red was the 15s `waitFor` from #2273 timing out, not anything in this diff.

ayaangazali:
Confirming the outcome of that rebase: `typescript/cli (windows-latest, node-22.23.1)` is **green** on the fresh run, with no change to the diff.

Same eight files, same three commits, only a different runner. That settles what the earlier red was: the 15s `waitFor` in `dev.test.ts` timing out under load, not anything in this branch.

It is also a clean data point for #2273. The test passed on Ubuntu both times, failed once on Windows, and passed on Windows on a re-run of identical code. That is the shape of a timeout budget that is too tight rather than a broken test.

This PR is now 19 checks green with one red, `Report current-head conformance scores`, which is the fork-token limit that #2310 addresses.

khandrew1 (approved):

PR #2350, opened 2026-08-27
docs(cli): add a README so the npm page is not blank

ayaangazali:
## Why

`@mcp-use/cli` is the only published package in the family without a README, and the registry reflects it:

```
$ curl -s https://registry.npmjs.org/@mcp-use%2Fcli | jq -r .readme
""
```

So the npm page for the package that provides the `mcp-use` executable renders nothing at all. Every sibling has one:

| package | README |
|---|---|
| mcp-use | 380 lines |
| @mcp-use/inspector | 433 |
| create-mcp-use-app | 518 |
| @mcp-use/agent | 145 |
| @mcp-use/client | 127 |
| @mcp-use/tunnel | 22 |
| **@mcp-use/cli** | **none** |

## What it says

Deliberately short, closer to the `@mcp-use/tunnel` README than the `mcp-use` one, because this package is rarely installed directly. `mcp-use` depends on it and exposes the same `bin`, which both `package.json` and the existing CLI reference page state, so the README leads with that rather than pretending it is a standalone install.

Contents: what the package is, how it actually reaches users, the command list taken from `mcp-use --help`, and links onward. Nothing invented, and no flag table duplicated here that would drift from the docs site.

## Checks

Both links resolve directly, no redirect:

```
200  hops=0  https://docs.mcp-use.com/v2/typescript/api-reference/cli-reference
200  hops=0  https://docs.mcp-use.com/v2/typescript/getting-started/quickstart
```

`README.md` is not in this package's `files` array, and it does not need to be, since npm always includes it. Confirmed rather than assumed:

```
$ npm pack --dry-run
npm notice Tarball Contents
npm notice 1.5kB README.md
npm notice 895B THIRD_PARTY_NOTICES.md
npm notice 642B dist/bin.js
```

No changeset: content only, no behaviour change.

---
## Summary by cubic
Adds a README to `@mcp-use/cli` so its npm page no longer renders blank.

- `@mcp-use/cli` was the only published package in the family without a README.
- The README is short, leads with the built-in MCP client, and lists the commands from `mcp-use --help`.
- It links to the CLI reference and quickstart docs; content only, so no changeset was added.

<sup>Written for commit f1fa0469e7f8db6ef8eff24c2c729592ed73061f. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2350?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/cli/README.md
+++ b/libraries/typescript/packages/cli/README.md
@@ -1,0 +1,43 @@
+# `@mcp-use/cli`
+
+Prebuilt CLI, development server, and build pipeline for
+[mcp-use](https://github.com/mcp-use/mcp-use).
+
+Most people never install this package directly. `mcp-use` depends on it and
+exposes the same `mcp-use` executable, so a project that installs the framework
+already has it:
+
+```bash
+npm install mcp-use
+npx mcp-use dev
+```
+
+## Commands
+
+```text
+mcp-use <command> [options]
+
+  dev          Start the dev server
+  build        Build the server into .mcp-use/build
+  typecheck    Refresh MCP types and run the project's TypeScript compiler
+  start        Serve the production build from .mcp-use/build
+  client       Connect to and invoke MCP servers
+  screenshot   Capture an MCP Apps view
+  deploy       Deploy from GitHub or upload local source
+  login/logout/whoami/org/servers/deployments
+               Manage Manufact Cloud sessions, servers, and deployments
+```
+
+`mcp-use dev` serves the MCP endpoint, watches views for HMR, and mounts the
+Inspector. `--tunnel` exposes either `dev` or `start` through the managed
+relay in [`@mcp-use/tunnel`](https://www.npmjs.com/package/@mcp-use/tunnel).
+
+Run `mcp-use --help` for the full option list.
+
+## Documentation
+
+The exhaustive command reference, including every flag and the CLI environment
+variables, lives at
+[CLI Reference](https://docs.mcp-use.com/v2/typescript/api-reference/cli-reference).
+New to the framework? Start with the
+[quickstart](https://docs.mcp-use.com/v2/typescript/getting-started/quickstart).
--- a/libraries/typescript/packages/cli/README.md
+++ b/libraries/typescript/packages/cli/README.md
@@ -1,15 +1,12 @@
 # `@mcp-use/cli`
 
-Prebuilt CLI, development server, and build pipeline for
+Prebuilt CLI, development server, build pipeline, and MCP client for
 [mcp-use](https://github.com/mcp-use/mcp-use).
 
-Most people never install this package directly. `mcp-use` depends on it and
-exposes the same `mcp-use` executable, so a project that installs the framework
-already has it:
-
 ```bash
-npm install mcp-use
 npx mcp-use dev
+npx mcp-use client connect demo https://example.com/mcp
+npx mcp-use client demo tools list
 ```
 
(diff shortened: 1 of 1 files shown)


libraries/typescript/packages/cli/README.md
```diff
@@ -0,0 +1,43 @@
+# `@mcp-use/cli`
+
+Prebuilt CLI, development server, and build pipeline for
+[mcp-use](https://github.com/mcp-use/mcp-use).
+
+Most people never install this package directly. `mcp-use` depends on it and
```
khandrew1: i'd remove this paragraph

libraries/typescript/packages/cli/README.md
```diff
@@ -0,0 +1,43 @@
...
+exposes the same `mcp-use` executable, so a project that installs the framework
+already has it:
+
+```bash
+npm install mcp-use
+npx mcp-use dev
```
khandrew1, line 7: would prefer to show off `npx mcp-use client` instead in the example

libraries/typescript/packages/cli/README.md
```diff
@@ -0,0 +1,43 @@
+# `@mcp-use/cli`
+
+Prebuilt CLI, development server, and build pipeline for
```
khandrew1: would mention the built in MCP client here too

khandrew1 (changes requested):

ayaangazali:
All three applied, and apologies for the delay: your review landed on the 27th and I only found it today when I went looking properly. That is on my sweep, not on you.

- Mentioned the client in the opening line: "Prebuilt CLI, development server, build pipeline, and MCP client".
- Dropped the "most people never install this directly" paragraph.
- Example now leads with the client:

```bash
npx mcp-use dev
npx mcp-use client connect demo https://example.com/mcp
npx mcp-use client demo tools list
```

Those two client commands are the ones I actually ran against a local server earlier while checking the command list, rather than copied from the help text. `client connect <name> <url>` saves it, `client <name> tools list` prints the tool names.

New head `f1fa046`.

ayaangazali:
That red is the known Windows flake, not this branch, so I have not pushed anything for it.

This PR adds one file and changes nothing else:

```
added libraries/typescript/packages/cli/README.md +40/-0
```

A markdown file cannot affect `dev.test.ts`. The failure is the signature from #2273:

```
 ❯ Module.waitFor tests/cli/helpers.ts:158:9
 ❯ tests/cli/dev.test.ts:436:5
```

Note there is no `; last error:` suffix. `waitFor` only appends that when the probe throws, so nothing was refusing connections; the thing being waited for never happened inside the window. That is the watcher-side candidate, which is what #2338 is testing.

Ubuntu passed on the same run:

```
typescript/cli (ubuntu-latest, node-22.23.1)   success
typescript/cli (windows-latest, node-22.23.1)  failure
```

Worth noting that this comparison exists at all because #2325 set `fail-fast: false` on that matrix. Before it merged, the Ubuntu leg would very likely have been cancelled the moment Windows failed, and this would have looked like it could be either platform.

khandrew1 (approved):

PR #2303, opened 2026-08-20
feat(inspector): correlate RPC response labels

Whxuan0701:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Correlate JSON-RPC responses with their originating requests in the Inspector traffic log so response rows show the request method and elapsed time. Error responses remain visibly distinguishable, while unmatched responses keep their existing `result` or `error` labels.

This follows the maintainer's proposed presentation in https://github.com/mcp-use/mcp-use/issues/2296#issuecomment-5345466662.

## Review Readiness

Help maintainers review this quickly:

- [x] The PR is scoped to one issue or one clearly related change
- [x] The PR description explains the user-visible problem and the fix
- [x] The diff avoids unrelated formatting, generated files, and bulk rewrites
- [x] I verified the affected package/docs path with the commands listed below
- [x] I checked for existing open PRs that already solve the same issue

## Implementation Details

1. Build response display labels by matching JSON-RPC request and response IDs within the same source, server, widget, and opposite traffic direction.
2. Preserve JSON-RPC ID types during correlation so numeric and string IDs do not collide.
3. Keep correlation display-only: stored traffic entries, copied logs, and downloaded JSON are unchanged.
4. Add focused unit coverage for success, error, bidirectional traffic, source/server/widget isolation, and ID type isolation.
5. Add a patch changeset for `@mcp-use/inspector`.

---

## TypeScript Checklist

> Complete this section if your PR includes TypeScript changes

### Packages Modified

Check all packages that were modified:
- [ ] `docs`
- [ ] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [x] `inspector`

### Pre-commit Checklist

Ensure all of the following have been completed:
- [ ] Ran `pnpm lint:fix` to auto-fix linting issues
- [ ] Ran `pnpm format` to format code with Prettier
- [x] Ran `pnpm build` and build succeeds without errors
- [x] Ran `pnpm changeset` to create a changeset (if this PR includes user-facing changes)
- [x] Added or updated tests if needed
- [ ] Updated documentation in `docs/` folder if needed

Non-mutating `pnpm lint` and `pnpm format:check` validations pass; no documentation changes are needed for this UI-only fix.

---

## Python Checklist

> Complete this section if your PR includes Python changes

[...]

PR #2364, opened 2026-08-29
feat(inspector): show request duration instead of timestamp in RPC logger

### LINKED ISSUE #2327: Inspector: Replace timestamp with traffic time
Following up from PR #2303

The text in the logs gets truncated a lot and I don't think the timestamp is very useful information compared to how long the request actually took. We should replace the timestamp with the milliseconds

merttcetn:
## Description

Fixes #2327 (follow-up to #2303)

The RPC log rows rendered the wall-clock timestamp on the right of each row, while the request latency was squeezed into the truncated method label. Per the issue, the timestamp is rarely useful compared to how long the request actually took, and the inline latency contributed to text truncation.

This PR:

- Moves the latency (`42 ms`) into the right-side slot where the timestamp rendered, keeping the dotted-underline tooltip (now showing `Request took 42 ms (<wall-clock time>)`).
- Removes the inline `· 42 ms` from the truncated method label so longer method names are less likely to be cut off.
- Rows without a measured duration (notifications, pending requests) render no right-side time.
- Search matching and row `data-testid`s now use the bare method name instead of `method · 42 ms`.

No changes to the correlation layer were needed — `buildRpcTrafficResponseLabels` already computes per-pair latency client-side.

## Testing

- `pnpm vitest run` for `rpc-traffic-correlation` and `rpc-traffic-store` (11 tests, unchanged and passing)
- `pnpm type-check` (after building workspace deps)
- `eslint` on the changed file
- Manual: connect to a server in the inspector, observe durations on response rows

## Changeset

Added a `patch` changeset for `@mcp-use/inspector`.

---
## Summary by cubic
Fixes #2327 by showing the request duration instead of the wall-clock timestamp in RPC log rows, so you can see latency at a glance.

- Moves the latency (e.g., `42 ms`) to the right-side slot where the timestamp rendered; the tooltip now shows both latency and the original wall-clock time.
- Removes the inline `· 42 ms` from the method label so longer method names are less likely to be truncated.
- Rows without a measured duration (notifications, pending requests) render no right-side time.
- Search matching and row `data-testid`s now use the bare method name instead of `method · 42 ms`.
- Adds a `patch` changeset for `@mcp-use/inspector`.

<sup>Written for commit 272f19205911dd222a40889c0a83e6683980b9a4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2364?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/rpc-logger-duration.md
+++ b/libraries/typescript/.changeset/rpc-logger-duration.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/inspector": patch
+---
+
+Inspector RPC logger: replace the per-row timestamp with the request duration in milliseconds
--- a/libraries/typescript/packages/inspector/src/client/components/logging/JsonRpcLoggerView.tsx
+++ b/libraries/typescript/packages/inspector/src/client/components/logging/JsonRpcLoggerView.tsx
@@ -421,28 +421,23 @@ function RpcLogRow({
           ) : null}
         </span>
 
-        <Tooltip>
-          <TooltipTrigger
-            render={
-              responseLabel ? (
+        {responseLabel ? (
+          <Tooltip>
+            <TooltipTrigger
+              render={
                 <span className="shrink-0 cursor-default font-mono text-[10px] tabular-nums text-muted-foreground underline decoration-dotted underline-offset-2">
                   {responseLabel.latencyMs} ms
                 </span>
-              ) : (
-                <time
-                  dateTime={item.timestamp}
-                  className="shrink-0 cursor-default font-mono text-[10px] tabular-nums text-muted-foreground underline decoration-dotted underline-offset-2"
-                >
-                  {new Date(item.timestamp).toLocaleTimeString()}
-                </time>
-              )
-            }
-            nativeButton={false}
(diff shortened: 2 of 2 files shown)


khandrew1:
@merttcetn Can you please attach a screenshot of the changes you made? Make sure to always attach screenshots when you are doing UI work as it makes it easier for me to review. Thanks!

merttcetn:
pr-2364-rpc-duration

merttcetn:
Thanks for the reminder — I've attached the screenshot above. It shows the request durations in the right-side slot, with the original timestamp available in the hover tooltip. I'll make sure to include screenshots with future UI changes.

khandrew1 (approved):
Looks good to me! Thanks for the PR!

PR #2367, opened 2026-08-29
fix(client): stop OAuth retries after disconnect

### LINKED ISSUE #2366: client: raw TypeError instead of clean error when disconnect happens during an OAuth retry
## Description

In `libraries/typescript/packages/client/src/transport/base.ts`, most request methods guard with `if (!this.client) throw new Error("MCP client is not connected")` and then call the SDK client through a closure using a non-null assertion, e.g.:

```ts
const result = await this.executeRequest(() =>
  this.client!.listTools(undefined, options)
);
```

`HttpConnector.executeRequest` (`transport/http.ts`) can retry the closure once after completing an interactive OAuth flow (`http.ts:304-323`). That wait can be long (user completing login in a browser). If `disconnect()` runs in the meantime, `cleanupResources()` sets `this.client = null` (`base.ts:966`), and the retried closure re-evaluates `this.client!` at call time, producing:

```
TypeError: Cannot read properties of null (reading 'listTools')
```

instead of the connector's clean "MCP client is not connected" error.

There are 13 such `this.client!` closures in `base.ts` (initialize, callTool, listTools, listResources, listAllResources, listResourceTemplates, complete, readResource, subscribeToResource, unsubscribeFromResource, listPrompts, getPrompt, request).

## Suggested fix

Capture the client into a local after the guard (`const client = this.client;`) and use it inside the closure. This removes the non-null assertions and makes the closures immune to `this.client` being nulled between the guard and the (re)invocation.

Elioooon:
## Summary

- re-check the HTTP connector state after interactive OAuth completes
- return the existing clean not-connected error when a disconnect wins the OAuth wait
- cover the race and verify that the protected operation is not retried

Fixes #2366.

## Validation

From `libraries/typescript/packages/client`:

- `./node_modules/.bin/vitest run tests/unit/client/mixed-auth.test.ts` — 7 passed
- `../../node_modules/.bin/rimraf dist && ../../node_modules/.bin/tsup && ../../node_modules/.bin/tsc --emitDeclarationOnly --declaration` — passed
- `./node_modules/.bin/vitest run` — 56 files, 344 tests passed

From `libraries/typescript`:

- `./node_modules/.bin/eslint packages/client/src/transport/http.ts packages/client/tests/unit/client/mixed-auth.test.ts` — passed
- `./node_modules/.bin/prettier --check packages/client/src/transport/http.ts packages/client/tests/unit/client/mixed-auth.test.ts` — passed

## Risk

The added guard runs only after an interactive OAuth wait and immediately before the retry. Connected retries are unchanged; a connector disconnected during the wait now rejects deterministically instead of invoking a closure whose client was cleared.

---
## Summary by cubic
Fixes #2366 by preventing a protected operation from retrying after interactive OAuth when the connector disconnected during the wait. The operation now rejects deterministically with "MCP client is not connected" instead of invoking a closure whose client was cleared.

<sup>Written for commit f0dafa9f70a52b5d927b07ceaa61ef453551c4b0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2367?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/calm-clients-stop.md
+++ b/libraries/typescript/.changeset/calm-clients-stop.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Stop OAuth authorization retries after the client disconnects.
--- a/libraries/typescript/packages/client/src/transport/http.ts
+++ b/libraries/typescript/packages/client/src/transport/http.ts
@@ -318,6 +318,9 @@ export class HttpConnector extends BaseConnector {
         throw error;
       }
       await this.completeInteractiveAuthorization();
+      if (!this.connected || !this.client) {
+        throw new Error("MCP client is not connected");
+      }
       return operation();
     }
   }
--- a/libraries/typescript/packages/client/tests/unit/client/mixed-auth.test.ts
+++ b/libraries/typescript/packages/client/tests/unit/client/mixed-auth.test.ts
@@ -174,6 +174,45 @@ describe("mixed OAuth authorization", () => {
     });
   });
 
+  it("does not retry a protected operation after disconnecting during OAuth", async () => {
+    let finishAuthorization!: () => void;
+    const authorization = new Promise<void>((resolve) => {
+      finishAuthorization = resolve;
+    });
+    const callTool = vi
+      .fn()
+      .mockRejectedValueOnce(new UnauthorizedError("Authentication required"));
+    const connector = new HttpConnector("https://mcp.example.com/mcp", {
+      authProvider: createProvider({
+        getAuthorizationResponse: vi.fn(async () => {
+          await authorization;
+          return {
+            code: "authorization-code",
+            iss: "https://auth.example.com",
+          };
+        }),
+      }),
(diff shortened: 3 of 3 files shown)


khandrew1 (changes requested):
Looks good to me, please add a changeset to this PR as well. A "patch" is appropriate as it's a non public facing fix. Thanks!

Elioooon:
Added the requested patch changeset in `f0dafa9` for `@mcp-use/client`. `changeset status` resolves it as `2.2.1` → `2.2.2`, and `git diff --check` is clean. Thanks for the review.

khandrew1 (approved):
LGTM! Thanks for the PR!

PR #2337, opened 2026-08-26
fix(cli): isolate dev test cleanup failures

Elioooon:
## Summary

Addresses the Ubuntu teardown path discussed in #2273.

This keeps the `waitFor` timeout unchanged and instead isolates cleanup failures so one failed fixture removal cannot leak `MCP_URL`/`PORT` into later dev tests.

Changes:
- restore `MCP_URL` and `PORT` from `afterEach` in a `finally` block
- continue running remaining cleanup callbacks even when one cleanup throws
- make `removeDir` match its best-effort comment by retrying transient recursive removal failures and swallowing the final failure

## Validation

- `pnpm --filter @mcp-use/cli exec vitest run tests/cli/dev.test.ts` - 34 passed
- `pnpm --filter @mcp-use/cli run typecheck` - passed
- `pnpm --filter @mcp-use/cli exec prettier --check tests/cli/helpers.ts tests/cli/dev.test.ts` - passed

## Risk / boundary

This only changes test cleanup behavior. It does not change runtime CLI behavior, port allocation, or the shared `waitFor` timeout. If the Windows watcher issue is still present, it should be handled separately as discussed in #2273.

No changeset: test-only change.

---
## Summary by cubic
Fixes dev test cleanup so one failed fixture removal can't leak `MCP_URL` or `PORT` into later tests.

- Restores `MCP_URL` and `PORT` in a `finally` block, so they're reset even when cleanup throws.
- Runs all cleanup callbacks even if one fails, then reports the first error (or an `AggregateError` for multiple).
- Makes `removeDir` retry transient recursive removals and swallow the final failure to match its best-effort intent.

Test-only change; no runtime CLI behavior, port allocation, or shared timeout changes. No changeset required.

<sup>Written for commit a6dacc866d153648d8d09b090063ad1678225df6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2337?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2368, opened 2026-08-29
fix(cli): retry the resolved port on a late EADDRINUSE in mcp-use dev

### LINKED ISSUE #2273: cli dev tests flake on main through waitFor, and the 24x gap suggests a port race rather than a slow runner

oleg-ai-dev:
## Summary

`resolvePort` confirms a port is free by binding and immediately releasing a throwaway probe server (`packages/cli/src/cli/port.ts`), but real work runs between that probe and the actual bind: `runDev` resolves the port early — before importing the entry module, so entry-scope code can read the committed port via `process.env.PORT` — then does entry import + Vite dev-server setup before the real `httpServer.listen()` call at the end. That gap is long enough for another transient prober (a concurrent `resolvePort` call, a test's own free-port check) to grab and release the same port in between, crashing the real bind with an unhandled `EADDRINUSE`.

This is the root cause behind #2273's `tests/cli/dev.test.ts` flakiness — the test helper's own `getFreePort` has the identical bind-then-release shape, but the same pattern also exists in production `resolvePort`, so this isn't just test-infra flakiness.

## Fix

- Added `listenWithRetry(server, port, host)` to `port.ts`: binds for real, and on `EADDRINUSE` retries the *same* port a few times (5 attempts, 100ms apart) before giving up. Picking a *different* port at this point isn't safe — by the time the real bind runs, the entry module may have already baked the original port into its own config (e.g. OAuth redirect URIs), so silently switching ports here would desync the CLI's actual bind from what the entry already committed to. Non-`EADDRINUSE` errors still reject immediately, same as before.
- `dev.ts` now calls `listenWithRetry(httpServer, port, host)` instead of a raw `.listen()` wrapped in a promise.
- Also raised the test helper's `waitFor` default timeout from 15s to 30s (`tests/cli/helpers.ts`) — still comfortably under vitest's 60s `testTimeout`. This addresses the issue's other candidate explanation (a tight helper timeout can itself surface as a bare `waitFor timed out` on a slow runner) as a complementary fix, not a substitute for the port fix — both candidates the issue raised turned out to have distinct evidence pointing at them in the two CI failure transcripts.
- Added a changeset (`@mcp-use/cli` patch).

## Testing

New `tests/cli/port.test.ts` (5 tests): immediate bind when free, riding out a transient `EADDRINUSE` via retry, exhausting retries and throwing on a port held for good, not retrying non-`EADDRINUSE` errors, and confirming `resolvePort` still leaves the port unbound for the caller.

- `npx vitest run tests/cli/port.test.ts` — 5/5 passing
- `npx vitest run tests/cli/dev.test.ts` — 37/39 passing; the 2 remaining failures are pre-existing and unrelated to this change (confirmed by reverting these changes and rerunning the identical suite against unmodified `main`): a Windows-only `ENOTEMPTY` on fixture cleanup (a Vite file-watcher handle race, unrelated to ports) and a reload-count assertion that's simply flaky (passed on the baseline run in the same comparison).
- `npm run typecheck` (cli package) and `eslint` on the changed files — clean.

Fixes #2273

---
## Summary by cubic
Fixes `mcp-use dev` crashing with an unhandled `EADDRINUSE` when another transient prober grabs the resolved port between the free-port probe and the real bind.

- The real bind now retries the same port 5 times (100ms apart) before throwing; non-`EADDRINUSE` errors still reject immediately.
- Retrying the same port (not switching) keeps the bind aligned with the port the entry module already committed to via `process.env.PORT`.
- Raised the test helper's `waitFor` default timeout from 15s to 30s so slow CI runners aren't cut off prematurely.
- Added `tests/cli/port.test.ts` covering immediate bind, transient contention, retry exhaustion, non-`EADDRINUSE` errors, and `resolvePort` leaving the port unbound.
- Adds a patch changeset for `@mcp-use/cli` and fixes #2273.

<sup>Written for commit 99dccbdd27ca0ed38885ea98ec3f948c3f84ced1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2368?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/slow-mice-dream.md
+++ b/libraries/typescript/.changeset/slow-mice-dream.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Fix `mcp-use dev` occasionally failing to bind its resolved port: `resolvePort` confirms a port is free by binding and immediately releasing a throwaway probe server, but the entry import and Vite setup that run before the real bind (the port has to be committed early so entry-module code can read it via `process.env.PORT`) leave enough time for another transient prober — a concurrent `resolvePort` call, a test's own free-port check — to grab and release that exact port in between, crashing the real bind with an unhandled `EADDRINUSE`. The real bind now retries the same port a few times before giving up, riding out that transient hold instead of crashing on it.
--- a/libraries/typescript/packages/cli/src/cli/dev.ts
+++ b/libraries/typescript/packages/cli/src/cli/dev.ts
@@ -51,7 +51,7 @@ import {
   nextStandaloneCompatPlugin,
   nextStandaloneSsrOptions,
 } from "./next-compat.js";
-import { resolvePort } from "./port.js";
+import { listenWithRetry, resolvePort } from "./port.js";
 import { resolveTailwindCss, resolveUserViteConfig } from "./vite-config.js";
 import { createDevApiHandler } from "./dev-api.js";
 import {
@@ -983,10 +983,7 @@ export async function runDev(options: DevOptions): Promise<void> {
   };
   httpServer.on("request", onRequest);
 
-  await new Promise<void>((resolve, reject) => {
-    httpServer.once("error", reject);
-    httpServer.listen(port, host, () => resolve());
-  });
+  await listenWithRetry(httpServer, port, host);
 
   console.log(`[mcp-use] dev server ready`);
   if (currentViews.length > 0) {
--- a/libraries/typescript/packages/cli/src/cli/port.ts
+++ b/libraries/typescript/packages/cli/src/cli/port.ts
@@ -4,12 +4,19 @@
  */
 
 import { connect, createServer } from "node:net";
+import type { Server } from "node:net";
 
 /** How many consecutive ports to probe before giving up. */
 const MAX_PROBES = 100;
 
 /** Short timeout for loopback connect probes (ms). */
 const CONNECT_PROBE_MS = 300;
+
+/** Bind retries for {@link listenWithRetry} before giving up on the same port. */
+const LISTEN_RETRY_ATTEMPTS = 5;
+
+/** Delay between bind retries in {@link listenWithRetry} (ms). */
+const LISTEN_RETRY_DELAY_MS = 100;
 
 /**
  * Outcome of {@link resolvePort}.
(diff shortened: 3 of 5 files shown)


Parker-Fawcett (commented):
Clean fix - the EADDRINUSE retry is a real pain in development. One suggestion: consider making the probe a part of the resolved-port contract rather than a separate retry loop, so the port resolution is idempotent by design. This is the same pattern rebuild-dossier uses - lock the interface before executing rather than retrying around the failure. Paper context: https://arxiv.org/abs/2608.23616

khandrew1:
Thanks for taking a look at this. I’m going to close the PR because it is based on hypotheses that were superseded by the later investigation in #2273.

The failing CI logs contained no EADDRINUSE. For Ubuntu, the initial failure was an ENOTEMPTY during teardown; environment restoration was then skipped, leaking MCP_URL=sandbox.example.com into subsequent tests and causing the downstream fetch failures. That path was addressed in #2337.

The Windows failure had a different signature: requests continued succeeding, but the expected filesystem update never arrived. The Windows polling/watcher path was investigated separately, and the tentative fix has landed in Canary through #2338.

The global waitFor increase from 15s to 30s is also not safe. That exact change was tested in the issue discussion and caused HMR tests to fail because it extends the observation window in which an unwanted server-entry reload can arrive. The reload assertion failure in this PR’s reported 37/39 run matches that known behavior, so it should not be treated as an unrelated pre-existing failure.

Closing since the issue-specific changes here are either unrelated to the observed failures or known to alter the test semantics.

PR #2370, opened 2026-08-29
fix(python): surface SSE transport fallback

### LINKED ISSUE #2369: python: streamable HTTP -> SSE fallback is invisible to users
## Description

In `libraries/python/mcp_use/client/connectors/http.py`, when the streamable HTTP transport fails, the connector silently falls back to the legacy SSE transport:

- `http.py:257` — the streamable failure is logged at `debug` level only:
  ```python
  logger.debug(f"Streamable HTTP failed: {streamable_error}")
  ```
- `http.py:261-264` — cleanup errors of the failed streamable connection manager are swallowed entirely:
  ```python
  try:
      await connection_manager.close()
  except Exception:
      pass
  ```
- `http.py:274` — the SSE fallback attempt is also only a `debug` message.

When a server supports both transports, a streamable failure followed by a successful SSE fallback is completely invisible at the default log level. Users debugging slow connections, missing streamable features (progress, resumability), or flaky servers get no signal that the modern transport failed and why.

## Suggested fix

Raise the streamable failure log to `warning` (it is a notable event that changes transport behavior), log swallowed cleanup errors at `warning` with the exception detail, and log the SSE fallback attempt at `info`.

luisangelrod:
## Description

Fixes #2369.

When streamable HTTP failed, the transport change was only visible at debug level. The cleanup path also called `close()` even though `StreamableHttpConnectionManager` exposes `stop()`, and swallowed the resulting error.

This change:

- emits a warning with the failure category before fallback
- announces the SSE fallback at info level
- keeps detailed exception text at debug level and endpoint data out of default-visible logs
- stops the failed streamable HTTP manager through its supported lifecycle method
- tightens the fallback tests with spec-constrained mocks and cleanup assertions

## Testing

- `uv run --extra dev --extra e2b python -m pytest tests/unit -q` — 305 passed
- `ruff check .` — passed
- `ruff format --check .` — 228 files already formatted

## Backwards Compatibility

No public API changes. Successful streamable HTTP connections are unchanged; only failed-attempt cleanup and fallback visibility change.

---
## Summary by cubic
Surfaces the SSE transport fallback in logs so users can see when streamable HTTP fails and the client switches to SSE. Previously this was only visible at debug level, and cleanup called `close()` on a manager that exposes `stop()`, swallowing any cleanup errors.

- Emits a warning with the failure category before falling back and an info message when SSE takes over.
- Keeps detailed exception text and endpoint data at debug level.
- Stops the failed streamable HTTP manager via `stop()` and logs cleanup failures instead of swallowing them.
- Tightens fallback tests with spec-constrained mocks and cleanup assertions.

<sup>Written for commit 0803540b71cb783f3231fb3b0ee152ff65c982a7. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2370?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/client/connectors/http.py
+++ b/libraries/python/mcp_use/client/connectors/http.py
@@ -254,14 +254,16 @@ class HttpConnector(BaseConnector):
         # Exception from the inner try is propagated here and in
         # the most cases is an McpError, so checking instances is useless
         except Exception as streamable_error:
-            logger.debug(f"Streamable HTTP failed: {streamable_error}")
+            logger.warning("Streamable HTTP connection failed (%s)", type(streamable_error).__name__)
+            logger.debug("Streamable HTTP failure details: %s", streamable_error)
 
             # Clean up the failed streamable HTTP connection manager
             if connection_manager:
                 try:
-                    await connection_manager.close()
-                except Exception:
-                    pass
+                    await connection_manager.stop()
+                except Exception as cleanup_error:
+                    logger.warning("Failed to stop streamable HTTP connection (%s)", type(cleanup_error).__name__)
+                    logger.debug("Streamable HTTP cleanup failure details: %s", cleanup_error)
 
             # It doesn't make sense to check error types. Because client
             # always return a McpError, if he can't reach the server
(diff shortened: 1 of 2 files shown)


khandrew1:

PR #2371, opened 2026-08-30
fix(python): correct required-field detection for structured output

### LINKED ISSUE #2090: MCPAgent structured-output required-field detection is inverted for pydantic models
**Describe the bug**

In `libraries/python/mcp_use/agents/mcpagent.py`, both `_attempt_structured_output` and `_enhance_query_with_schema` compute required-ness as:

```python
required = not hasattr(field_info, "default") or field_info.default is None
```

Pydantic's `FieldInfo` always has a `.default` attribute, so `hasattr(...)` is always `True`, collapsing the expression to `field_info.default is None`. A genuinely required field (no default) has `default = PydanticUndefined`, not `None` — so it's incorrectly marked **optional**. An explicitly optional field declared as `Optional[str] = None` is incorrectly marked **required**.

**To Reproduce**

1. Define an `output_schema` with a required `str` field (no default) and an optional `Optional[str] = None` field.
2. Run structured output validation.
3. Missing data in the required field won't raise; the optional field can incorrectly trigger a 'missing required field' error/retry.

**Severity:** Medium

malaviatanay:
Closes #2090.

### Problem

`FieldInfo` in pydantic v2 always has a `.default` attribute, so the required-field check

```python
required = not hasattr(field_info, "default") or field_info.default is None
```

collapses to `field_info.default is None`. That inverts the result:

| Declaration | Actual `.default` | Old check | Correct |
|---|---|---|---|
| `title: str` (required) | `PydanticUndefined` | optional | required |
| `note: str \| None = None` (optional) | `None` | required | optional |

Consequences in `MCPAgent` structured output:

- A genuinely required field that comes back missing/empty is **not** flagged, so incomplete structured output passes validation silently.
- An `Optional[x] = None` field that comes back empty is flagged as a missing required field and triggers the structured-output retry loop unnecessarily.
- The schema hints sent to the model (`_enhance_query_with_schema` and the `stream` schema description) label required/optional backwards.

### Fix

Replace all **three** occurrences with pydantic's own `FieldInfo.is_required()`:

- `_attempt_structured_output` (validation that raises → retry)
- `_enhance_query_with_schema` (schema hints for the model)
- the `stream` loop's schema-description builder — this one is not listed in the issue but has the identical bug

`is_required()` has been on `FieldInfo` since pydantic 2.0 and the library already pins `pydantic>=2.11,<3`.

### Tests

Added `TestMCPAgentStructuredOutputRequiredFields` in `tests/unit/test_agent.py`, using a schema with one required field and one `str | None = None` field:

- `_enhance_query_with_schema` labels the required field `(required)` and the optional field `(optional)`
- `_attempt_structured_output` raises when the required field is missing
- `_attempt_structured_output` returns normally when only the optional field is missing

All three fail on `main` and pass with this change. Full `tests/unit` suite passes; `ruff check` / `ruff format` clean; `ty check` reports no new diagnostics.

---
## Summary by cubic
Fixes required-field detection for pydantic v2 structured output in `MCPAgent` so missing required fields are rejected instead of passing silently, and optional `None` fields no longer trigger the retry loop. Applies `FieldInfo.is_required()` across all three affected code paths and adds regression tests for both cases. Closes #2090.

<sup>Written for commit 2b8e34b77a42825c4e4cc0a97897ea38c90b9143. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2371?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/agents/mcpagent.py
+++ b/libraries/python/mcp_use/agents/mcpagent.py
@@ -596,7 +596,7 @@ class MCPAgent:
 
         try:
             for field_name, field_info in output_schema.model_fields.items():
-                required = not hasattr(field_info, "default") or field_info.default is None
+                required = field_info.is_required()
                 if required:
                     value = getattr(structured_result, field_name, None)
                     if value is None or (isinstance(value, str) and not value.strip()):
@@ -616,7 +616,7 @@ class MCPAgent:
         try:
             for field_name, field_info in output_schema.model_fields.items():
                 description = getattr(field_info, "description", "") or field_name
-                required = not hasattr(field_info, "default") or field_info.default is None
+                required = field_info.is_required()
                 schema_fields.append(f"- {field_name}: {description} {'(required)' if required else '(optional)'}")
 
             schema_description = "\n".join(schema_fields)
@@ -889,7 +889,7 @@ class MCPAgent:
                     schema_fields = []
                     for field_name, field_info in output_schema.model_fields.items():
                         description = getattr(field_info, "description", "") or field_name
-                        required = not hasattr(field_info, "default") or field_info.default is None
+                        required = field_info.is_required()
                         schema_fields.append(
                             f"- {field_name}: {description} " + ("(required)" if required else "(optional)")
                         )
--- a/libraries/python/tests/unit/test_agent.py
+++ b/libraries/python/tests/unit/test_agent.py
@@ -2,11 +2,13 @@
 Unit tests for the MCPAgent class.
 """
 
+from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


khandrew1:

PR #2373, opened 2026-08-30
test(cli): move the CLI scratch fixtures to the OS temp directory

ayaangazali:
## Why

`copyFixture` wrote its mutable copies to `packages/cli/tests/cli/.tmp`, inside the repo. Everything that walks the working tree then has to be told to skip them: ESLint failed on the generated `mcp-env.d.ts` files, `vitest.config.ts` carried an exclude, and the directory is not in `.gitignore` either, so a stray `git add -A` picks up the whole tree of fixtures and the pre-commit hook runs Prettier over them.

This started as an ESLint ignore. @khandrew1 asked for the real fix instead: use the OS temp directory, like the rest of the suite already does. That is what this is now.

## What it took

The two-line version does not work. It relocates the copies and then fails 57 tests:

```
Cannot find module 'zod' imported from
  /private/var/folders/.../mcp-use-cli-tests/dev-zero-views-bound-91607ae0/src/index.ts
Error: TypeScript is not installed in /va…
```

Fixture projects import `zod`, run `tsc`, and start Vite. `copyFixture` only ever linked one dependency:

```ts
symlinkSync(serverPackageRoot, join(nodeModules, "mcp-use"), "junction");
```

Everything else resolved by walking up out of `tests/cli` into the CLI package's own `node_modules`. From the OS temp directory there is nothing to walk up into.

**One link at the scratch root** restores that lookup for every copy beneath it. Node walks up from `<copy>/src` to `<copy>/node_modules` for `mcp-use`, then to `TMP_ROOT/node_modules` for the rest. Cheaper than installing into each copy, and each copy still sees the package layout of an installed consumer.

**`TMP_ROOT` goes through `realpathSync`.** macOS reports `/var/folders/...` for a directory that lives at `/private/var/folders/...`. Vite compares the two, decides a fixture's own assets sit outside the project root, and serves them as `/@fs/private/var/...`. The asset-URL assertion in `dev.test.ts` caught it:

```
- Expected: /http:\/\/localhost:21015\/views\/product-search-result\/badge\.png/
+ Received: "export default \"http://localhost:21015/@fs/private/var/folders/.../badge.png\""
```

## Result

Nothing is written inside the repo, so nothing needs to be excluded from anything. The ESLint ignore this PR originally added is gone, and so is the `tests/cli/.tmp` entry in `vitest.config.ts`. No `.gitignore` entry is needed either.

## Validation

From `libraries/typescript/packages/cli`:

- `vitest run tests/cli/` — 116 passed, 12 files
- `vitest run` — 294 passed, 23 files
- suite wall clock is unchanged at about 15s

From `libraries/typescript`:

- `eslint packages/cli/tests/cli/helpers.ts packages/cli/vitest.config.ts eslint.config.js` — clean with the ignore removed, which is the check that matters
- `prettier --check` on the same files — clean

### DIFF
--- a/libraries/typescript/packages/cli/tests/cli/helpers.ts
+++ b/libraries/typescript/packages/cli/tests/cli/helpers.ts
@@ -2,18 +2,22 @@
 import { randomBytes } from "node:crypto";
 import {
   cpSync,
+  existsSync,
   mkdirSync,
+  realpathSync,
   readFileSync,
   rmSync,
   symlinkSync,
   writeFileSync,
 } from "node:fs";
 import { createServer, type Server } from "node:net";
+import { tmpdir } from "node:os";
 import { dirname, join } from "node:path";
 import { fileURLToPath } from "node:url";
 
 const here = dirname(fileURLToPath(import.meta.url));
 const serverPackageRoot = join(here, "..", "..", "..", "server");
+const cliPackageModules = join(here, "..", "..", "node_modules");
 
 /** Absolute path to the committed basic fixture project. */
 export const FIXTURE_BASIC = join(here, "fixtures", "basic");
@@ -22,18 +26,45 @@ export const FIXTURE_BASIC = join(here, "fixtures", "basic");
 export const FIXTURE_VIEWS = join(here, "fixtures", "views");
 
 /**
- * Scratch root for mutable fixture copies. Each copy receives a local
- * `node_modules/mcp-use` link, matching the package layout of an installed
- * consumer without creating a workspace dependency cycle between the CLI and
- * server packages.
+ * Scratch root for mutable fixture copies, under the OS temp directory like
+ * the rest of the suite.
+ *
+ * Resolved through `realpathSync` because macOS hands out `/var/folders/...`
+ * for a directory that really lives at `/private/var/folders/...`. Vite
+ * compares the two and serves anything it thinks is outside the project root
+ * through `/@fs/`, so a fixture's own assets would come back with the wrong
+ * URL if the root kept the unresolved form.
  */
-export const TMP_ROOT = join(here, ".tmp");
+export const TMP_ROOT = createTmpRoot();
+
+function createTmpRoot(): string {
+  const root = join(tmpdir(), "mcp-use-cli-tests");
+  mkdirSync(root, { recursive: true });
+  return realpathSync(root);
+}
+
+/**
+ * Link the CLI package's installed dependencies next to the scratch copies.
+ *
+ * Fixture projects import `zod`, run `tsc`, and start Vite. While the copies
+ * lived under `tests/cli`, they resolved all of that by walking up into the
+ * CLI package's own `node_modules`; from the OS temp directory there is
+ * nothing to walk up into. One link at the scratch root restores that lookup
+ * for every copy beneath it without giving each one its own install.
+ */
+function linkDependencies(): void {
+  const link = join(TMP_ROOT, "node_modules");
+  if (!existsSync(link)) {
+    symlinkSync(cliPackageModules, link, "junction");
+  }
+}
 
 /** Copy a committed fixture into a fresh scratch dir; returns its path. */
 export function copyFixture(
   label: string,
   fixture: "basic" | "views" = "basic"
 ): string {
+  linkDependencies();
   const source = fixture === "views" ? FIXTURE_VIEWS : FIXTURE_BASIC;
   const dest = join(TMP_ROOT, `${label}-${randomBytes(4).toString("hex")}`);
   mkdirSync(dest, { recursive: true });
--- a/libraries/typescript/packages/cli/tests/cli/helpers.ts
+++ b/libraries/typescript/packages/cli/tests/cli/helpers.ts
@@ -2,7 +2,6 @@
 import { randomBytes } from "node:crypto";
 import {
   cpSync,
-  existsSync,
   mkdirSync,
   realpathSync,
   readFileSync,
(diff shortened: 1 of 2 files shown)


khandrew1 (changes requested):
Thanks for catching this! I don't believe the fix for this issue would be just to ignore it in ESLint. Instead we should be using `os.tmpdir()` in our tests to use the Operating System's global temporary directory as this is what all our other tests do.

ayaangazali:
I tried it that way first and it breaks the suite, so here is what I found before deciding anything.

Pointing `TMP_ROOT` at the OS temp dir is a two-line change:

```diff
-export const TMP_ROOT = join(here, ".tmp");
+export const TMP_ROOT = join(tmpdir(), "mcp-use-cli-tests");
```

The scratch directory does move out of the repo, and the vitest exclude becomes dead. But:

```
Test Files  3 failed | 9 passed (12)
     Tests  57 failed | 59 passed (116)

Cannot find module 'zod' imported from
  /private/var/folders/.../mcp-use-cli-tests/dev-zero-views-bound-91607ae0/src/index.ts
Error: TypeScript is not installed in /va…
```

**The in-repo location is load-bearing.** `copyFixture` only symlinks one dependency:

```ts
symlinkSync(serverPackageRoot, join(nodeModules, "mcp-use"), "junction");
```

Everything else the fixtures import, `zod`, `typescript`, vite and so on, resolves by walking up the directory tree into the monorepo's `node_modules`. Under `os.tmpdir()` there is nothing to walk up into. The tests that already use `tmpdir()` are the ones whose fixtures do not import anything.

So the options as I see them:

1. **Symlink the rest of the dependency set into each scratch copy**, then move to `tmpdir()`. This is the version of your suggestion that actually works, but it means every scratch copy grows a real `node_modules` and the helper has to track which packages the fixtures need.
2. **Keep the scratch dir in the repo and tell the tools about it.** That is this PR. Worth adding: `tests/cli/.tmp` is not in `.gitignore` either, so a leftover fixture shows up in `git status` as well as breaking lint.
3. Something else you have in mind that I have not thought of.

Happy to do 1 if you want it, it is just meaningfully bigger than it looks. I have not pushed the broken version.

ayaangazali:
New evidence for this, from tripping over it myself an hour ago.

Working on an unrelated fix in `packages/server`, I ran `git add -A` and committed. The commit picked up 62 files:

```
 .../typescript/.changeset/loud-pans-repeat.md      |   5 +++
 .../.tmp/dev-custom-skills-113ef209/mcp-env.d.ts   |  10 ++++++
 .../.tmp/dev-custom-skills-113ef209/package.json   |   6 ++++
 .../.tmp/dev-overlapping-skills-views-2b1b5f44/... | ...
 .../views/product-search-result/badge.png          | Bin 0 -> 111 bytes
 ... 60 more under packages/cli/tests/cli/.tmp/ ...
 .../typescript/packages/server/src/oauth/jwt.ts    |   8 +++--
```

The pre-commit hook ran Prettier over the scratch fixtures on the way through, which is the same class of problem as the lint failure this PR started from.

So the cost of leaving the scratch directory in the repo is not only ESLint. Anything that walks the working tree sees it: `git add -A`, `git status`, the pre-commit hook, and grep. That pushes me toward your `os.tmpdir()` instruction being the right call, I just cannot get there with the two-line version for the reasons in my earlier comment. If you want option 1 from that comment, symlinking the fixtures' dependency set into each scratch copy so they resolve outside the workspace, say the word and I will do it here. Otherwise the smallest thing that stops the bleeding today is this PR plus a `.gitignore` entry, and I am happy to add the latter.

ayaangazali:
Done your way, @khandrew1. Pushed on top of your rebase rather than force-pushing over it.

The scratch copies are under `os.tmpdir()` now, the ESLint ignore this PR opened with is reverted, and the `vitest.config.ts` exclude is gone too. Title and description updated to match.

Two things the plain move needed, both in the description with the failure output:

- a single `node_modules` link at the scratch root, because the fixtures were resolving `zod`, `typescript` and Vite by walking up into the CLI package's `node_modules` and there is nothing to walk up into from the temp directory
- `realpathSync` on `TMP_ROOT`, because macOS hands back `/var/folders/...` for a directory that lives at `/private/var/folders/...` and Vite then serves fixture assets through `/@fs/`

116 passing in `tests/cli`, 294 in the package, and `eslint` is clean with the ignore removed, which was the point.

khandrew1:
hey @ayaangazali, sorry need a bit more time to review your findings. going through PRs right now.

ayaangazali:
Windows caught a real hole in the first version of this, now fixed. `typescript/cli (windows-latest)` is green.

The shared `node_modules` link at the scratch root was fine on Linux and macOS and failed 51 tests on Windows:

```
Cannot find module 'zod' imported from
  C:/Users/RUNNER~1/AppData/Local/Temp/mcp-use-cli-tests/build-skills-84948767/src/index.ts
```

The runner puts TEMP on `C:` and the checkout on `D:`, and the entries that link was reaching through are pnpm links into `.pnpm`, so resolution through one junction two directories above the project did not survive the crossing.

Each copy now links what it actually needs into its own `node_modules`: `zod`, `typescript`, `react` and `react-dom`, resolved to their real `.pnpm` directories so each package sits beside the peers pnpm installed for it. That is the layout pnpm creates itself, and it does not depend on anything above the copy. `TMP_ROOT` also moved to `realpathSync.native`, so Windows resolves the `RUNNER~1` short path the same way macOS resolves `/var` to `/private/var`.

```
typescript/cli (windows-latest, node-22.23.1)   pass   2m58s
typescript/cli (ubuntu-latest, node-22.23.1)    pass   1m27s
```

Locally: 294 passed in the package, eslint and prettier clean, monorepo preflight 0 failing packages.

Worth saying plainly: the list of four dependencies is the one thing here that can rot. If a fixture picks up a new import later it will fail with the same `Cannot find module`, and the fix is to add it to `FIXTURE_DEPENDENCIES`. I preferred that to a shared link precisely because the shared link failed silently by platform, but if you would rather the fixtures carried a real `package.json` with these as dependencies and got a proper install, say so and I will do it that way.

khandrew1:
Thanks for investigating this and updating the PR, including tracking down the Windows failures. The failures you found exposed a gap in our fixture setup: we create runnable projects without fully declaring and installing their dependencies, so they rely on the surrounding repository.

Moving them to `os.tmpdir()` is still the right direction. The latest per-fixture dependency links address the shared-link problem, but the fixtures still borrow packages from the repository’s installation and require a separate dependency list in the helper. I agree with the proper-install approach you mentioned: we should prepare complete fixture projects with explicit dependencies and installed local packages, then run the existing tests against them.

I’ve opened [#2419 — Give CLI integration tests complete, isolated fixture projects](https://github.com/mcp-use/mcp-use/issues/2419) to track that setup change. I’m closing this PR so we can take that work forward in a dedicated replacement PR. Your investigation gives us a useful starting point, including the dependency-resolution failures and the macOS and Windows path-normalization requirements.

PR #2374, opened 2026-08-30
test(cli): make the remaining skill rewrites change file length

ayaangazali:
Follow-up to #2338, using the reason you gave there rather than a guess of my own.

You changed `"Guidance v1"` to `"Skills Guidance v2"` because a same-byte-length rewrite may not register as a change. Two more rewrites in `dev.test.ts` have exactly that shape, and both are followed by a `waitFor` on the watcher noticing:

| file under test | old | new | bytes |
|---|---|---|---|
| `skills/returns/SKILL.md` | `description: Process returns` | `description: Updated returns` | 66 and 66 |
| `skills/refunds/references/policy.md` | `Policy v1\n` | `Policy v2\n` | 11 and 11 |

Both give the new value a different length now, and the assertions that read them back are updated to match.

I found them by sweeping the test tree for consecutive `writeFileSync` calls to the same path whose literals are equal length. Those two were the only real hits. A third match in `packages/server/tests/skills.test.ts` was a false positive, `join(directory, "templates", ...)` path segments rather than file contents, so it is untouched.

## Checks

`dev.test.ts`, three consecutive runs: **34 passed, 34 total** each time.

Being straight about one thing: on the first run after the edit I saw a single failure in `runDev (views) > discards a candidate superseded while its server entry is evaluating`, a test this diff does not touch. Three re-runs were clean and unmodified `canary` also passes 34/34, so that was the pre-existing flake rather than anything here. Mentioning it because it happened, not because I think it is related.

---
## Summary by cubic
Fixes two same-byte-length rewrites in `dev.test.ts` so the dev watcher reliably detects the file change. Updates the assertions that read the rewritten `SKILL.md` and `policy.md` contents to match the new values.

<sup>Written for commit 6874c76ad84699fe38ebf0a42258463e8c36a6e6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2374?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/cli/tests/cli/dev.test.ts
+++ b/libraries/typescript/packages/cli/tests/cli/dev.test.ts
@@ -345,7 +345,7 @@ describe("runDev", () => {
 
     writeFileSync(
       addedSkillFile,
-      "---\nname: returns\ndescription: Updated returns\n---\n# Returns\n"
+      "---\nname: returns\ndescription: Updated returns policy\n---\n# Returns\n"
     );
     await waitFor(async () => {
       const body = await mcpRequest(dev.url, "skills/list");
@@ -353,7 +353,7 @@ describe("runDev", () => {
         skills?: Array<{ frontmatter?: { description?: string } }>;
       };
       return result.skills?.some(
-        (skill) => skill.frontmatter?.description === "Updated returns"
+        (skill) => skill.frontmatter?.description === "Updated returns policy"
       )
         ? true
         : undefined;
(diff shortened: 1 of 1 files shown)


khandrew1 (approved):
Reviewed after rebasing onto the latest canary. The PR is limited to the two intended watcher-test rewrites and matching assertions; no blocking issues found.

PR #2375, opened 2026-08-31
fix(ci): fail conformance summary on unexpected warnings

Xander-Xai:
## Problem

`summarize-conformance.mjs` counts `unexpectedWarnings` — conformance checks with `WARNING` status that are not listed in `expected-failures.yml` — but the script's only exit-code gate is:

```js
if (totals.unexpectedFailures > 0) process.exitCode = 1;
```

An unexpected `WARNING` is reported in the summary table but does **not** cause CI to fail. A `WARNING` is a degraded conformance result (not full success), so an unexpected one should not pass silently.

## Reproduction

Create a `checks.json` with a single `WARNING` status check in a suite that is not covered by `expected-failures.yml`:

```json
[{ "id": "check1", "status": "WARNING" }]
```

Run `summarize-conformance.mjs` against it:

- **Before this PR:** exit code `0` — CI stays green.
- **After this PR:** exit code `1` — CI fails.

## Root cause

The script explicitly tracks `unexpectedWarnings` through the aggregation pipeline but only gates the exit code on `unexpectedFailures`. `WARNING` checks are not `FAILURE`, so they slip through.

## Fix

Extend the exit-code condition:

```js
if (totals.unexpectedFailures > 0 || totals.unexpectedWarnings > 0) process.exitCode = 1;
```

## Regression coverage

`summarize-conformance.test.mjs` adds three cases:
- all `SUCCESS` → exit 0
- unexpected `FAILURE` → exit 1 (existing behavior)
- unexpected `WARNING` → exit 1 (new)

## Validation

- One production file changed: `summarize-conformance.mjs` (1 line)
- One regression test file added: `summarize-conformance.test.mjs`
- The regression test is registered through `test:summarize-conformance`
- The conformance workflow explicitly runs the regression test
- No package release changeset is required because this is a repo/CI helper change
- No executable package behavior changed
- Expected warnings listed in `expected-failures.yml` still do not fail

## Risk / compatibility

This may cause CI to fail on runs that currently produce unexpected `WARNING` checks. That is the intended behavior: an unexpected warning is a signal that should not be silenced.


khandrew1:
Thanks for the contribution. The upstream conformance project is currently deciding whether `WARNING` results should fail CI in modelcontextprotocol/conformance#430. Since warnings represent SHOULD-level or advisory outcomes, we don’t want to independently treat them as failures before that policy is resolved upstream. Closing this for now; we can align our behavior once the upstream semantics are settled.

PR #2379, opened 2026-08-31
docs(cli): correct the deployments subcommand table

ayaangazali:
## Why

Seven claims in the `deployments` table do not match the binary. Following #2352 and #2376, I checked every cell by running it rather than reading the help text.

The CLI makes this easy to test without credentials: a bad option or subcommand fails at **exit 2**, while anything that parses reaches `Not logged in` at **exit 1**.

| documented | what happens |
|---|---|
| `list` / **`ls`** | `Usage: mcp-use deployments <list\|get\|logs\|restart\|stop\|delete>` |
| `delete` / **`rm`** | same |
| **`start <deployment-id>`** | same, the subcommand does not exist |
| `list --sort <field:direction>` | `Unknown option '--sort'` |
| `logs -b` | `Unknown option '-b'` |
| `logs -f`, `restart -f` | `Unknown option '-f'` |
| `stop`, options "None" | `Stop deployment dep_x? Pass --yes to confirm.` |

The long forms `--build` and `--follow` do work. Only the short aliases are missing. The dispatch accepts exactly `list`, `get`, `logs`, `restart`, `stop`, `delete` (`deployments.ts:63-71`), so `ls`, `rm` and `start` have nowhere to land.

## What changed

- Dropped the `ls` and `rm` aliases and the `start` row.
- `--sort` removed; **`--org <id-or-slug>` and `--server <id>` added**, both real and previously missing. The command's own usage string advertises them.
- Short flags `-b`, `-f`, `-y` removed, long forms kept.
- `stop` now shows `--yes`, and the notes say it prompts like `delete` does.
- Dropped the `mcp-use deployments start dep_abc123` example, which cannot run.

## Verification

Every row in the corrected table, run against the binary:

```
exit=1  deployments list --org acme --server srv --limit 5 --skip 0 --json
exit=1  deployments get dep_x
exit=1  deployments logs dep_x --build --follow
exit=1  deployments restart dep_x --follow --branch main
exit=1  deployments stop dep_x --yes
exit=1  deployments delete dep_x --yes
```

All reach the auth check instead of failing on arguments. Before the change, four of those six failed at exit 2.

I checked the `servers` table the same way and it is accurate, including every flag on `servers update`, so it is untouched.

---
## Summary by cubic
Corrects the `deployments` subcommand table in the CLI reference so it matches what the binary accepts, and documents the `--json` output flag.

- Removes the `ls` and `rm` aliases, the `start` subcommand, `--sort`, and the short flags `-b`, `-f`, and `-y`; the long forms `--build`, `--follow`, and `--yes` still work.
- Adds `--org <id-or-slug>` and `--server <id>` to `list`, and `--json` to every subcommand.
- Updates the notes: `stop` and `delete` prompt on a TTY; pass `--yes` in non-interactive or JSON runs.
- Removes the `deployments start dep_abc123` example, which cannot run.

<sup>Written for commit 1d516ab52895130494adf0914f00495d81a14ada. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2379?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/v2/typescript/api-reference/cli-reference.mdx
+++ b/docs/v2/typescript/api-reference/cli-reference.mdx
@@ -405,15 +405,14 @@ mcp-use deployments <subcommand> [options]
 
 **Subcommands:**
 
-| Subcommand                                      | Purpose                                      | Options                                                 |
-| ----------------------------------------------- | -------------------------------------------- | ------------------------------------------------------- |
-| `list` / `ls`                                   | List deployments.                            | `--limit <n>`, `--skip <n>`, `--sort <field:direction>` |
-| `get <deployment-id>`                           | Show deployment details.                     | None                                                    |
-| `logs <deployment-id>`                          | View runtime logs.                           | `-b, --build`, `-f, --follow`                           |
-| `restart <deployment-id>`                       | Trigger a new deployment on the same server. | `-f, --follow`, `--branch <name>`                       |
-| `stop <deployment-id>`                          | Stop a running deployment.                   | None                                                    |
-| `start <deployment-id>`                         | Start a stopped deployment.                  | None                                                    |
-| `delete <deployment-id>` / `rm <deployment-id>` | Delete a deployment.                         | `-y, --yes`                                             |
+| Subcommand                | Purpose                                      | Options                                                                          |
+| ------------------------- | -------------------------------------------- | -------------------------------------------------------------------------------- |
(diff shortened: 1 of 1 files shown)


khandrew1 (approved):
Thanks for catching these!

PR #2383, opened 2026-09-01
fix(python): await async telemetry calls

### LINKED ISSUE #2091: Python telemetry() decorator does not await wrapped async methods

Elioooon:
## Type of change

- [x] Python
- [ ] TypeScript
- [ ] Documentation
- [ ] Other

## Summary

The shared `@telemetry` decorator currently wraps coroutine functions with a synchronous wrapper. Calling any decorated async method therefore records a successful, near-zero-duration event as soon as the coroutine object is created, before its body runs. Exceptions raised while awaiting the coroutine are never captured.

This change uses `inspect.iscoroutinefunction` to preserve an async wrapper for coroutine functions, awaits the call before emitting telemetry, and keeps the existing synchronous path unchanged. It also preserves `_record_telemetry=False` while still awaiting and returning the async result.

The impact is shared rather than call-site-specific: current main has 23 async methods using this decorator across the client, sessions, connectors, OAuth, middleware, and server manager.

## Before

- async telemetry is emitted before execution completes
- awaited failures are reported as successful calls
- execution time measures coroutine creation, not execution

## After

- telemetry is emitted after the await completes
- awaited exceptions record `success=False` and the exception type
- elapsed time includes the coroutine execution
- decorated async methods remain recognizable as coroutine functions

## Testing

- `uv run pytest -q tests/unit/test_telemetry_decorator.py` - 4 passed
- `uv run pytest -q tests/unit` - 309 passed (with the `dev` and `e2b` extras)
- `uv run ruff check .` - passed
- `uv run ruff format --check .` - 229 files already formatted
- `uv run ty check mcp_use/telemetry/telemetry.py tests/unit/test_telemetry_decorator.py` - passed

The repository-wide `ty check mcp_use` baseline currently reports 168 diagnostics in unrelated files; the touched files pass directly.

## Risk and compatibility

This uses only the standard library and does not change event fields. Synchronous decorated methods retain their existing behavior. The intentional behavior change is that telemetry for async methods is now emitted at completion rather than coroutine creation.

Closes #2091

---
## Summary by cubic
Fixes the `@telemetry` decorator so async methods emit telemetry after execution completes instead of when the coroutine is created.

- Async wrappers now await the call before recording; previously, telemetry captured a near-zero duration before the body ran.
- Awaited exceptions and `CancelledError` are captured with `success=False` and the exception type, then re-raised.
- `_record_telemetry=False` still skips telemetry while awaiting and returning the async result.
- Synchronous decorated methods keep their existing behavior.
- Adds unit tests covering async success, failure, cancellation, disabled telemetry, and sync paths.

<sup>Written for commit 569debdaa0bc21231def2b68f8885c6ef2e21e23. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2383?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/telemetry/telemetry.py
+++ b/libraries/python/mcp_use/telemetry/telemetry.py
@@ -1,3 +1,4 @@
+import inspect
 import logging
 import os
 import platform
@@ -53,6 +54,51 @@ def telemetry(event_name: str, additional_properties: dict[str, Any] | None = No
     """Decorator to automatically track feature usage"""
 
     def decorator(func: Callable) -> Callable:
+        def capture_event(self, start_time: float, success: bool, error_type: str | None) -> None:
+            telemetry_client = None
+            if hasattr(self, "telemetry"):
+                telemetry_client = self.telemetry
+            elif hasattr(self, "_telemetry"):
+                telemetry_client = self._telemetry
+            else:
+                telemetry_client = Telemetry()
+
+            if telemetry_client:
+                telemetry_client.capture(
+                    event=GenericTelemetryEvent(
+                        EVENT_NAME=event_name,
+                        **{
+                            **(additional_properties or {}),
+                            "success": success,
+                            "execution_time_ms": int((time.time() - start_time) * 1000),
+                            "error_type": error_type,
+                        },
+                    )
+                )
+
+        if inspect.iscoroutinefunction(func):
+
+            @wraps(func)
+            async def async_wrapper(self, *args, **kwargs):
+                record_telemetry = getattr(self, "_record_telemetry", None)
+                if record_telemetry is not None and not record_telemetry:
+                    return await func(self, *args, **kwargs)
+
+                start_time = time.time()
+                success = True
+                error_type = None
+
+                try:
+                    return await func(self, *args, **kwargs)
+                except Exception as e:
+                    success = False
+                    error_type = type(e).__name__
+                    raise
+                finally:
+                    capture_event(self, start_time, success, error_type)
+
+            return async_wrapper
+
         @wraps(func)
         def wrapper(self, *args, **kwargs):
             # Check if telemetry is disabled on this instance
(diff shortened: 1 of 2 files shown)


Elioooon:
Addressed both review findings in 569debd:

- `asyncio.CancelledError` is now recorded as an unsuccessful call and re-raised; a cancellation regression test covers the boundary.
- the elapsed-time test now replaces only the telemetry module's `time` reference, so it cannot affect asyncio, logging, or other process-wide callers.

Revalidated with:

- `uv run pytest -q tests/unit/test_telemetry_decorator.py` (5 passed)
- `uv run pytest -q tests/unit` (310 passed)
- touched-file Ruff, format, and `ty` checks (passed)

khandrew1:

PR #2385, opened 2026-09-01
fix(telemetry): don't send raw prompt/response text by default

unixsvr4:
Follow-up to #1836, rescoped per @tonxxd's review. That PR bundled unrelated
bug fixes across the WebSocket connector, prompt builder, and error formatting.
This one is just the telemetry change: 2 source files, 1 test, 1 doc.

## Problem

`mcp_agent_execution` sends the full `query` and `response` strings to PostHog
whenever anonymized telemetry is on (the default). For a library that runs
inside other people's agents, forwarding raw end-user prompts and model
responses to a third party by default is a privacy/compliance concern.

## Change

Raw content becomes opt-in; the analytically-useful length signals stay on by
default.

- `telemetry/events.py` — widen `MCPAgentExecutionEvent.query` to `str | None`
  (`response` already was).
- `telemetry/telemetry.py` — read `MCP_USE_TELEMETRY_INCLUDE_CONTENT` (default
  `false`) once in `__init__`. `track_agent_execution` always derives
  `query_length` / `response_length` from the real content, then nulls out
  `query` / `response` unless opted in.

Default payload:

```json
"query": null,
"query_length": 44,
"response": null,
"response_length": 38
```

Opt back in (e.g. a self-hosted PostHog you own):

```bash
export MCP_USE_TELEMETRY_INCLUDE_CONTENT=true
```

## Testing

- `tests/unit/test_telemetry_content_optin.py` — content redacted by default
  with lengths preserved, content included when opted in, env default resolves
  to `false`.
- Full unit suite: 307 passed. The single failure
  (`test_config.py::test_create_sandboxed_stdio_connector`) reproduces on a
  clean `main` checkout and is unrelated.
- `ruff check` and `ruff format --check` clean.

## Compatibility

No API break. It does change the default telemetry payload — anyone relying on
that content in their own dashboards restores it with the env var. The docs
change is required, not optional: `telemetry.mdx` currently advertises
"Query and response content (to understand use cases)", which this makes untrue.

---
## Summary by cubic
Stops sending raw prompt and response text to PostHog in `mcp_agent_execution` telemetry. The default payload now reports only `query_length` and `response_length` and nulls out `query` and `response`, keeping usage analytics while preventing end-user content from going to a third party by default. An empty response is still counted as a response, so its `response_length` is `0` rather than `null`.

**Migration**
- Set `MCP_USE_TELEMETRY_INCLUDE_CONTENT=true` to send raw content again, e.g. for a self-hosted PostHog.
- The flag is read once in `Telemetry.__init__`, so set the env var before the telemetry instance is created.
- The telemetry docs now describe the new default, since they previously claimed raw content was collected.

<sup>Written for commit fc38f9e2fc70c3bd5694a40bc049daa18fee114d. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2385?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/python/development/telemetry.mdx
+++ b/docs/python/development/telemetry.mdx
@@ -22,7 +22,7 @@ MCP-use includes an **opt-out telemetry system** that helps us understand how th
 ### Agent Execution Data
 When you use `MCPAgent.run()` or `MCPAgent.stream()`, we collect:
 
-- **Query and response content** (to understand use cases)
+- **Query and response _length_** (character counts, always) — the raw text is **not** collected unless you opt in (see below)
 - **Model provider and name** (e.g., "openai", "gpt-4")
 - **MCP servers connected** (types and count, not specific URLs/paths)
 - **Tools used** (which MCP tools are popular)
@@ -35,10 +35,29 @@ When you use `MCPAgent.run()` or `MCPAgent.stream()`, we collect:
 - **Package download analytics** (via Scarf for distribution insights)
 
 ### What We DON'T Collect
+- **Raw prompt/response content** (only character lengths, unless you opt in — see below)
 - **Personal information** (no names, emails, or identifiers)
 - **Server URLs or file paths** (only connection types)
 - **API keys or credentials** (never transmitted)
 - **IP addresses** (PostHog configured with `disable_geoip=False`)
+
+## Prompt & Response Content
+
+By default, the `mcp_agent_execution` event reports only the **length** of the
+query and response (`query_length`, `response_length`), never the text itself.
+The raw `query` and `response` fields are sent as `null`.
+
+If you want to share the raw content (for example, on a self-hosted PostHog
+instance you control), opt in explicitly:
+
+```bash
+export MCP_USE_TELEMETRY_INCLUDE_CONTENT=true
+```
+
+<Warning>
+Only enable this if you own the telemetry destination or have consent to
+transmit user prompts. This can send end-user data off your machine.
+</Warning>
 
 ## How to Disable Telemetry
 
@@ -91,8 +110,10 @@ Here's what a typical telemetry event looks like:
   "properties": {
     "mcp_use_version": "1.3.0",
     "execution_method": "run",
-    "query": "Help me analyze sales data from the CSV file",
-    "response": "I'll help you analyze the sales data...",
+    "query": null,
+    "query_length": 44,
+    "response": null,
+    "response_length": 38,
     "model_provider": "openai",
     "model_name": "gpt-4",
     "server_count": 2,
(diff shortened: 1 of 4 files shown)


unixsvr4:
@tonxxd this is the rescoped version of #1836 you asked for — 4 files (2 source, 1 test, 1 doc) instead of 10, rebased on current `main`.

The unrelated WebSocket connector / prompt builder / error formatting fixes are dropped from this one. Happy to open those as a separate PR if they're useful.

unixsvr4:
Both review findings were valid — fixed in fc38f9e2. Still 4 files.

**1. `response_length` lost for an empty response.** Correct. `len(response) if response else None` sent `response_length: null` for `""`, which is exactly the signal this PR exists to preserve, and it was inconsistent with `query_length`, which uses an unconditional `len()`. Now `is not None`.

Worth flagging: this is **pre-existing upstream behaviour**, not something this PR introduced — the original line was `response_length=len(response) if response else None` and my change only hoisted it. I fixed it because it's one line, on a line the diff already touches, and directly contradicts the PR's purpose. Happy to drop it if you'd rather keep this strictly additive.

**2. `test_default_env_disables_content` was tautological.** Also correct, and the sharper version of the complaint is that it never constructed a `Telemetry` at all — it re-computed the env expression inline, so it would have passed even against `self._include_content = True`.

The reason it dodged the real code: `Telemetry` is wrapped in `@singleton`, so `Telemetry()` returns the cached instance and `__init__` never re-runs. The replacement reaches the real class via `type(Telemetry())` and initializes a throwaway object. `MCP_USE_ANONYMIZED_TELEMETRY=false` keeps `__init__` on its early branch, so no PostHog/Scarf clients get built. It now covers the default, opt-in (including `"TRUE"`), and non-`"true"` values, plus empty- and absent-response cases.

I mutation-tested the replacements rather than assuming they work:

| Mutation | Result |
|---|---|
| revert to `if response else None` | `test_empty_response_keeps_its_length` fails |
| hardcode `_include_content = True` | `test_include_content_defaults_to_false` + `_ignores_other_values` fail |

The replaced test caught neither.

7 tests in the file, 312 in `tests/unit`, `ruff check` and `ruff format --check` clean.

Unrelated but worth a separate PR if useful: `test_config.py::test_create_sandboxed_stdio_connector` fails on a clean `main` — it patches `mcp_use.connectors.sandbox.AsyncSandbox`, which has moved to `mcp_use.client.connectors.sandbox.Sandbox`. `create=True` masks it. Deliberately kept out of this PR.

tonxxd:
Hi we are currently not accepting ai generated prs and responses. Closing this for now, feel free to open an issue if needed

PR #2389, opened 2026-09-02
fix(cli): preserve sensitive flag on env update when --secret is omitted

### LINKED ISSUE #2388: cli: rotating an env variable without --secret silently clears its sensitive flag
## What happens

`mcp-use servers env set` sends `sensitive` on every write, derived only from whether `--secret` was passed on that invocation (`packages/cli/src/commands/servers.ts:296`):

```ts
const body = {
  key,
  value,
  branch: values.branch ?? null,
  environments: values.branch === undefined ? ["production"] : ["preview"],
};
```

When the key already exists the command PATCHes that body onto the existing variable. So rotating a value you originally created with `--secret`, without repeating the flag, clears the flag:

```bash
mcp-use servers env set srv_abc123 TOKEN=old --secret   # sensitive: true
mcp-use servers env set srv_abc123 TOKEN=rotated        # sensitive: false
```

Driving the second call against the mocked cloud API used by `tests/commands/servers.test.ts`, with `listVariables` returning an existing `{ id: "env_1", key: "TOKEN", sensitive: true }`:

```
PATH: /servers/server_1/env-variables/env_1
BODY: {"key":"TOKEN","value":"rotated","branch":null,"environments":["production"],"sensitive":false}
```

The CLI prints `Set TOKEN.` and exits 0. Nothing reports that the variable stopped being sensitive.

## Why it matters

Rotating a credential is the single most common reason to re-run `env set` on an existing key, and it is exactly the case that silently downgrades it. The docs describe `--secret` as "marks a value sensitive", which reads as a property of the variable rather than something that has to be re-asserted on every write.

## Two ways to go, and I do not want to guess

1. **Preserve the existing flag** when `--secret` is absent on an update. Matches the read of `--secret` as a property of the variable. The cost: there is no `--no-secret`, so this removes the only way to un-mark a variable, accidental as it currently is.
2. **Keep set as a full replace** and make the downgrade visible: mention it in `env set --help` and the CLI reference, and possibly warn when a write clears `sensitive` on an existing variable.

I am happy to send a PR for either. There is no test covering the flag on update today, so no existing behaviour pins it down. If you would rather do this yourself, feel free to close.

Found on `canary` at `b95fbc4e`; no cloud credentials involved, the evidence above is from the mocked API harness already in the repo.

aishwary-dongre:
# Pull Request Description

## Language / Project Scope
Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes
Fixes a bug where rotating an environment variable's value via `mcp-use servers env set` without repeating `--secret` silently cleared the variable's `sensitive` flag. Previously, `sensitive` on the PATCH/POST body was set purely from `values.secret === true` on that specific invocation, ignoring the variable's existing state. Now, when updating an existing variable and `--secret` is not passed, the existing `sensitive` value is preserved instead of being overwritten to `false`. New variables still default to `sensitive: false` when `--secret` is not passed, which is unchanged behavior.

## Review Readiness

## Implementation Details
1. In `packages/cli/src/commands/servers.ts`, `envSet()` already fetches the existing variable via `listVariables()` before building the request body, but was discarding `existing.sensitive` when constructing `body.sensitive`.
2. Added a `sensitive` variable computed as: `true` if `--secret` was passed, otherwise `existing?.sensitive ?? false` (preserves existing state on update, defaults to `false` for new variables).
3. Also fixed the CLI's printed `result.secret` field, which previously echoed `values.secret === true` (whether the flag was passed on this call) rather than the actual resulting `sensitive` state that was sent to the API. It now reports the real value.
4. No new dependencies added.

---

### Packages Modified
- [ ] `docs`
- [x] `tests`
- [x] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist
- [x] Ran `pnpm lint:fix` to auto-fix linting issues (ran via pre-commit hook, which passed)
- [x] Ran `pnpm format` to format code with Prettier (ran via pre-commit hook, which passed)

---

## Example Usage (Before)
```bash
mcp-use servers env set srv_abc123 TOKEN=old --secret   # sensitive: true
mcp-use servers env set srv_abc123 TOKEN=rotated        # sensitive: false (silently downgraded)
```

## Example Usage (After)
```bash
mcp-use servers env set srv_abc123 TOKEN=old --secret   # sensitive: true
mcp-use servers env set srv_abc123 TOKEN=rotated        # sensitive: true (preserved)
```

## Documentation Updates
None. This PR fixes runtime behavior only; the `--secret` flag's documented meaning ("marks a value sensitive") is unchanged, this fix makes actual behavior match that description on updates.

## Testing
- Added two unit tests to `packages/cli/tests/commands/servers.test.ts`:
  - Verifies that updating an existing sensitive variable without `--secret` keeps `sensitive: true` in the PATCH body (reproduces the exact bug scenario from the issue).
  - Verifies that a brand-new variable created without `--secret` still defaults to `sensitive: false`.
- Ran `pnpm exec vitest run tests/commands/servers.test.ts` locally: all 8 tests pass (6 existing + 2 new).
- No manual testing against a live cloud API; all testing uses the repo's existing mocked API harness.

## Backwards Compatibility
This is a backward-compatible bug fix. The only behavior change is that updating an existing sensitive variable without `--secret` no longer clears its `sensitive` flag; this brings actual behavior in line with what the `--secret` flag's documentation already describes. No API surface, CLI flags, or exit codes changed. Marked as a patch bump via changeset.

## Related Issues
Closes #2388

---
## Summary by cubic
Fixes `mcp-use servers env set` so rotating an environment variable's value without repeating `--secret` no longer silently clears its `sensitive` flag. Previously, `sensitive` was set purely from whether `--secret` was passed on that call, ignoring the variable's existing state; now updates preserve the existing value, while new variables still default to `sensitive: false`. Closes #2388.

**Bug Fixes**
- The CLI now preserves the existing `sensitive` value on env updates when `--secret` is omitted.
- The printed `secret` field now reports the actual resulting sensitive state instead of echoing whether the flag was passed.
- Added tests for both update preservation and new-variable default behavior.

<sup>Written for commit a7487ba7683ef923ba6fe537c676203dd8112ecb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2389?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/itchy-worms-hear.md
+++ b/libraries/typescript/.changeset/itchy-worms-hear.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Preserve the existing sensitive flag when rotating an environment variable without --secret
--- a/libraries/typescript/packages/cli/src/commands/servers.ts
+++ b/libraries/typescript/packages/cli/src/commands/servers.ts
@@ -290,12 +290,14 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     (variable) =>
       variable.key === key && (variable.branch ?? undefined) === values.branch
   );
+  const sensitive =
+    values.secret === true ? true : (existing?.sensitive ?? false);
   const body = {
     key,
     value,
     branch: values.branch ?? null,
     environments: values.branch === undefined ? ["production"] : ["preview"],
-    sensitive: values.secret === true,
+    sensitive,
   };
   if (existing === undefined) {
     await api.request<unknown>(
@@ -313,7 +315,7 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     key,
     scope: values.branch === undefined ? "production" : "preview",
     branch: values.branch ?? null,
-    secret: values.secret === true,
+    secret: sensitive,
     updated: existing !== undefined,
   };
   printResult(result, json, `Set ${key}.`);
--- a/libraries/typescript/packages/cli/src/commands/servers.ts
+++ b/libraries/typescript/packages/cli/src/commands/servers.ts
@@ -290,8 +290,7 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     (variable) =>
       variable.key === key && (variable.branch ?? undefined) === values.branch
   );
-  const sensitive =
-    values.secret === true ? true : (existing?.sensitive ?? false);
+  const sensitive = values.secret === true || (existing?.sensitive ?? false);
   const body = {
     key,
     value,
--- a/libraries/typescript/packages/cli/tests/commands/servers.test.ts
+++ b/libraries/typescript/packages/cli/tests/commands/servers.test.ts
@@ -84,6 +84,62 @@ describe("server environment output safety", () => {
       updated: true,
     });
   });
+
+  it("preserves an existing sensitive flag when --secret is not repeated on update", async () => {
+    api.request
+      .mockResolvedValueOnce([{ id: "env_1", key: "TOKEN", sensitive: true }])
+      .mockResolvedValueOnce({
+        id: "env_1",
+        key: "TOKEN",
+        value: "rotated",
+      });
+    vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+
+    await expect(
+      runServers(["env", "set", "server_1", "TOKEN=rotated", "--json"])
+    ).resolves.toBe(0);
+
+    expect(api.request).toHaveBeenLastCalledWith(
+      "/servers/server_1/env-variables/env_1",
+      {
+        method: "PATCH",
+        body: JSON.stringify({
+          key: "TOKEN",
+          value: "rotated",
+          branch: null,
+          environments: ["production"],
+          sensitive: true,
+        }),
+      }
+    );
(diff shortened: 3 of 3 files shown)


aishwary-dongre:
Good catch, and matches what the issue itself flagged: there's currently no --no-secret, so un-sensitizing isn't reachable through this command either before or after this fix. Kept this PR scoped to the reported bug (silent downgrade on rotation) rather than adding a new flag, since that felt like a separate feature decision for a maintainer to weigh in on. Happy to add --no-secret in a follow-up if that's wanted.

khandrew1:
Thanks for the contribution. Closing this out, see #2388 for more info

PR #2392, opened 2026-09-02
fix(create-mcp-use-app): drop the dead --no-git flag from help

MohammedAlkindi:
`create-mcp-use-app --help` lists `--no-git   Skip initializing a git repository`, but nothing reads that flag. `parseArgs` declares no `git` option, and with `strict: false` an unknown flag is accepted and dropped, so `--no-git` behaves exactly like omitting it. The neighbouring `--no-install` and `--no-skills` are real, both read via `argv.includes(...)` at lines 81 and 87.

a62db703 (#820) removed `tryGitInit` and its `options.git` call site to avoid scanning `node_modules`, and left the help line behind. This drops the line rather than restoring the behaviour, since that removal was deliberate.

Checks, run in `libraries/typescript`: `pnpm --filter create-mcp-use-app lint` clean, `test` 28/28 passing, `build` succeeds, and prettier reports no issues on the committed file.

---
## Summary by cubic
Removes the dead `--no-git` option from `create-mcp-use-app --help`. Git initialization was already removed, so the flag had no effect and only advertised unsupported behavior.

<sup>Written for commit c7f562e46dd4e4c4c75fc637b7126d2f3d14ee87. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2392?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/scaffold-drop-dead-no-git-flag.md
+++ b/libraries/typescript/.changeset/scaffold-drop-dead-no-git-flag.md
@@ -1,0 +1,5 @@
+---
+"create-mcp-use-app": patch
+---
+
+Remove `--no-git` from the help output. Scaffolding stopped initializing a git repository in #820, and nothing has read the flag since, so the line described behaviour the tool no longer had.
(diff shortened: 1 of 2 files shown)


khandrew1 (approved):
Thanks for catching this!

PR #2393, opened 2026-09-02
fix(client): rebind the OAuth loopback listener after a failed bind

MohammedAlkindi:
`startLoopback()` assigns `this.server` before `listen()` resolves, so a failed bind leaves a handle that never listened. The guard at the top of the method then makes every later call return early without binding. `stopLoopback()` is the only thing that clears it, and it is unreachable on this path: `this.pending` is created after `await this.startLoopback()`, so neither `resolvePending` nor `rejectPending` runs, and nothing in the repo calls `dispose()`.

The bind can fail because `reservePort()` probes the port free and binds later, leaving a window for something else to take it. The caller then opens a browser against a port with nothing listening and waits out `authTimeoutMs` (default 300000) rather than rebinding.

The added test occupies the port, asserts the first attempt rejects with `EADDRINUSE`, frees it, then requires the retry to serve a real request. Against unfixed source it fails with `ECONNREFUSED` on that retry; it passes with the fix.

`pnpm --filter @mcp-use/client test:run`: 342 passed, 1 skipped. Two failures remain in `browser-entry.test.ts`, which needs a built `dist/` that a fresh checkout lacks. `stdio-errlog` timed out once and passed on re-run, so it looks flaky here.

---
## Summary by cubic
Fixes OAuth loopback authorization after a failed port bind, which previously left a dead server handle and caused later attempts to wait until `authTimeoutMs`; failed binds now clean up so retries can listen normally.

**Bug Fixes**
- Keeps the in-flight listener visible to concurrent authorization calls and `dispose()`, then clears and closes it if binding fails.
- Reproduces the `reservePort()` race by occupying the selected port and verifies that a retry serves a callback.

<sup>Written for commit cd16b4e5369513fdec5e1e19cfb69223ae344b96. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2393?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/oauth-loopback-rebind-after-failed-bind.md
+++ b/libraries/typescript/.changeset/oauth-loopback-rebind-after-failed-bind.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Rebind the OAuth loopback listener after a failed bind. `startLoopback()` assigned `this.server` before `listen()` resolved, so a bind failure left a server that was never listening; the `if (this.server) return` guard then made every later authorization skip binding and hang until the auth timeout.
--- a/libraries/typescript/packages/client/src/auth/node.ts
+++ b/libraries/typescript/packages/client/src/auth/node.ts
@@ -492,10 +492,7 @@ export class NodeOAuthClientProvider implements OAuthClientProvider {
     const server = createHttpServer((req, res) => {
       this.handleCallback(req.url ?? "/", res);
     });
-    // Assign before awaiting the bind so an overlapping call and `dispose()`
-    // both observe the in-flight listener, then drop it if the bind fails: a
-    // server that never listened must not satisfy the guard above, or every
-    // later call short-circuits without ever binding.
+    // Track the server during binding so dispose() can find it.
     this.server = server;
     try {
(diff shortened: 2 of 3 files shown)


khandrew1 (approved):
Thanks for the PR! For future PRs that fix issues like these, could you also open up an issue beside it as it makes it easier for me to diagnose the root cause and track. Thanks!

khandrew1:
> pnpm --filter @mcp-use/client test:run: 342 passed, 1 skipped. Two failures remain in browser-entry.test.ts, which needs a built dist/ that a fresh checkout lacks. stdio-errlog timed out once and passed on re-run, so it looks flaky here.

Regarding this line in your PR description: I built the client and ran the full test suite—all 345 tests passed, so I’m merging this. Going forward, please make sure your PRs are fully built, tested, and verified before requesting review.

PR #2395, opened 2026-09-02
fix(python): preserve streamable HTTP initialization result

dlowzzxx:
## Scope

- [x] Python

## Problem

For Streamable HTTP, `HttpConnector.connect()` performs the MCP initialization handshake as a transport probe. It then marks the connector initialized and populates cached lists, but discards the successful `InitializeResult`. A later `MCPSession.initialize()` therefore returns `None`, and connector capabilities are unavailable.

## Fix

- Cache the successful Streamable HTTP `InitializeResult` and capabilities during `connect()`.
- Return that cached result from `initialize()` without a second wire handshake.
- Clear the cache during connector cleanup.
- Add regression coverage to the existing Streamable HTTP connector test.

## Validation

- [x] `ruff format --check mcp_use/client/connectors/http.py tests/unit/test_http_connector.py`
- [x] `ruff check mcp_use/client/connectors/http.py tests/unit/test_http_connector.py`
- [x] `python -m unittest tests.unit.test_http_connector -v` — 21 passed
- [x] `git diff --check`
- [x] Checked for existing open PRs addressing this behavior.

## Documentation

No documentation change is needed: this restores the existing initialization contract without changing the public API.

## Backwards Compatibility

Fully backwards compatible. This restores the expected initialization result and capability availability after a successful Streamable HTTP connection.

## Related Issues

None found.

---
## Summary by cubic
Fixes Streamable HTTP initialization so the successful handshake result is preserved and returned from `initialize()` instead of being lost. Capabilities are now available after connect, no second wire handshake is performed, and reconnects no longer reuse stale initialization state.

- Caches the probe `InitializeResult` and capabilities during `connect()`, and clears them during connector cleanup.
- Resets initialization state at the start of `connect()`, so an SSE fallback after a dropped Streamable HTTP session starts fresh.
- Adds regression coverage for the cached result, reconnect reset, and SSE fallback path.

<sup>Written for commit 9597be3674520dc3ef448c22412d8313ae7cd103. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2395?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/client/connectors/http.py
+++ b/libraries/python/mcp_use/client/connectors/http.py
@@ -136,6 +136,14 @@ class HttpConnector(BaseConnector):
             logger.debug("Already connected to MCP implementation")
             return
 
+        # A reconnect uses a new transport, so discard initialization state from the dropped session.
+        self._initialized = False
+        self._streamable_initialize_result = None
+        self.capabilities = None
+        self._tools = None
+        self._resources = None
+        self._prompts = None
+
         # Handle OAuth if needed
         if self._oauth:
             try:
--- a/libraries/python/mcp_use/client/connectors/http.py
+++ b/libraries/python/mcp_use/client/connectors/http.py
@@ -11,7 +11,7 @@ import httpx
 from mcp import ClientSession
 from mcp.client.session import ElicitationFnT, ListRootsFnT, LoggingFnT, MessageHandlerFnT, SamplingFnT
 from mcp.shared.exceptions import McpError
-from mcp.types import Root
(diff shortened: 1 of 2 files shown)


khandrew1:

PR #2363, opened 2026-08-29
docs(inspector): remove unsupported WebSocket and stdio transport claims

merttcetn:
## Description

Fixes #2361

The Inspector README advertised transports that are not implemented:

- **Key Features** claimed "Works with HTTP/SSE and WebSocket connections"
- **Adding an MCP Server** instructed users to select "WebSocket" and configure "stdin/stdio"

The client only supports Streamable HTTP, with SSE still accepted for legacy saved connections (see `ConnectionSettingsForm.tsx` — `transportType: "http", // HTTP only - SSE is deprecated`; the transport selector was removed from the connect form).

This PR updates both spots to describe only the supported options.

## Checklist

- [x] Documentation-only change (no code affected)
- [x] Verified no WebSocket/stdio transport plumbing exists in the Inspector client/server
- [x] Scoped to avoid overlap with #2357 (which touches other sections of this README)

---
## Summary by cubic
Removes unsupported WebSocket and stdio transport claims from the Inspector README so docs match the Streamable HTTP-only client, fixing #2361.

- Updates the Key Features table and "Adding an MCP Server" steps to describe only Streamable HTTP.
- Notes that SSE remains supported for legacy servers.

<sup>Written for commit f9a32d0a8e5be2168a944b12170d833545e266bd. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2363?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

PR #2397, opened 2026-09-02
docs: correct two README claims that contradict the code

MohammedAlkindi:
Two README lines that state something the code does not do. One commit each.

**Inspector feature table** claims "Works with HTTP/SSE server connections". `ConnectionSettingsForm.tsx:159` hardcodes `transportType: "http"` with the comment "HTTP only - SSE is deprecated", and this same README says "The Inspector supports Streamable HTTP only" a few sections down. #2363 corrected the WebSocket half of that row and left the SSE half.

**create-mcp-use-app quick-start** imports `MCPAgent` from `@mcp-use/agent` and passes `llm: new ChatOpenAI()`. The root entry types `llm` as `string | ProviderConfig`, so the snippet does not typecheck; the LangChain `MCPAgent` is the one that takes a model instance. The agent package README pairs `ChatOpenAI` with `@mcp-use/agent/langchain`, which is what this changes it to.

Docs-side only, no code touched. Prettier reports both files clean.

---
## Summary by cubic
Corrects two README claims that contradicted the code.

- Inspector feature table now says "Streamable HTTP" instead of "HTTP/SSE"; the connection panel hardcodes `transportType: "http"` and SSE is deprecated.
- `create-mcp-use-app` quick-start now imports `MCPAgent` from `@mcp-use/agent/langchain` so the example typechecks when passing a `ChatOpenAI` instance.

<sup>Written for commit c23516855467fc2389c4f597e96fedede58b7cd0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2397?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/create-mcp-use-app/README.md
+++ b/libraries/typescript/packages/create-mcp-use-app/README.md
@@ -341,7 +341,7 @@ Use the MCP server with any MCP-compatible client:
 
 ```typescript
 import { MCPClient } from "@mcp-use/client";
-import { MCPAgent } from "@mcp-use/agent";
+import { MCPAgent } from "@mcp-use/agent/langchain";
 import { ChatOpenAI } from "@langchain/openai";
 
 const client = new MCPClient({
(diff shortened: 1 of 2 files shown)


khandrew1 (approved):
Thanks for catching these!

PR #2403, opened 2026-09-02
fix(client): stabilize useMcp configuration dependencies to prevent reconnection loops

### LINKED ISSUE #2402: client: useMcp enters reconnect loop on parent re-renders when passing configuration objects
## Description

When passing unmemoized configuration object literals (`clientInfo`, `proxyConfig`, `headers`, or `clientOptions`) to the `useMcp` React hook, every re-render in the parent component triggers a connection teardown (`disconnect(true)`) followed immediately by a new `connect()` attempt.

This leads to continuous disconnect/reconnect loops that reset active sessions, abort in-flight tool calls, and spam reconnection requests.

## Reproduction

```tsx
import React, { useState } from "react";
import { useMcp } from "@mcp-use/client/react";

export function Dashboard() {
  const [count, setCount] = useState(0);

  // Inline object literals create fresh references on each parent render
  const mcp = useMcp({
    url: "http://localhost:3000/mcp",
    clientInfo: {
      name: "Dashboard App",
      version: "1.0.0",
    },
  });

  return (
    <button onClick={() => setCount((c) => c + 1)}>
      Render (Count: {count}) - State: {mcp.state}
    </button>
  );
}
```

**Steps to Reproduce:**
1. Render a component calling `useMcp` with an inline `clientInfo`, `proxyConfig`, or `headers` object.
2. Trigger any state update in the parent component.
3. Observe that `useMcp` teardown cleanup runs and a new transport connection is initiated on every click.

## Root Cause

In `libraries/typescript/packages/client/src/react/useMcp.ts`:
1. `mergedClientInfo` is computed with `[options.clientInfo, defaultClientInfo]` as dependencies (`useMcp.ts:208-214`). When `options.clientInfo` is passed inline, `mergedClientInfo` is recreated on every render.
2. The primary connection `useEffect` lists `mergedClientInfo` and `proxyConfig` as raw dependencies (`useMcp.ts:2086-2089`).
3. Each parent re-render causes the effect cleanup (`disconnect(true)`) to run and triggers `connect()`.

## Suggested Fix

1. Add a structural comparison helper (`isEqualDeep`) and a stable reference hook (`useStableValue`) in `useMcp-helpers.ts` to stabilize object/array options at the `useMcp` boundary.
2. Use stabilized configuration references for `mergedClientInfo`, `activeProxyConfig`, and `effectiveClientOptions`.

rohith500:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Stabilizes `clientInfo` and `proxyConfig` configuration dependencies in `useMcp` to prevent infinite reconnection loops when parents re-render with inline object literals.

## Review Readiness

## Implementation Details

1. **Targeted Stabilization for `clientInfo`**:
   - Compares scalar metadata properties (`name`, `title`, `version`, `description`, `websiteUrl`) and serializes `icons` into deterministic `[src, mimeType, sorted sizes]` tuples via `serializeClientIcons`.
   - Prevents re-renders with inline client information or nested icon literals from recomputing `mergedClientInfo` and triggering transport-level reconnections.
   - Legitimate changes to metadata or icons recompute `mergedClientInfo` and reconnect with updated configurations.

2. **Targeted Stabilization for `proxyConfig`**:
   - Compares `proxyAddress` alongside sorted header key-value tuples via `serializeProxyHeaders`.
   - Prevents reconnections when equivalent proxy headers are passed with differing key insertion orders.
   - Legitimate updates to `proxyAddress` or header values recompute `proxyConfig` and trigger reconnects.

3. **Preserved Effect Policy**:
   - Strictly scoped to `clientInfo` and `proxyConfig` (the two object dependencies responsible for reconnect loops). Reconnect effect dependencies remain identical to base.

4. **Comprehensive Regression Tests**:
   - Added unit tests in `useMcp-helpers.test.ts` for `serializeClientIcons` and `serializeProxyHeaders`.
   - Added tests in `useMcp-reference-stability.test.tsx` verifying:
     - Inline `clientInfo` and nested `icons` stay connected across multiple re-renders.
     - Changing `name` or `icons` reconnects with updated configuration.
     - Inline `proxyConfig` and re-ordered header keys stay connected.
     - Changing `proxyAddress` or header values reconnects with updated configuration.

---

### Packages Modified

- [x] `mcp-use` (client)

### Pre-commit Checklist

- [x] Ran `pnpm changeset` to create a changeset

---

## Testing

- Ran focused stability tests: `pnpm --filter @mcp-use/client exec vitest run tests/unit/react/useMcp-helpers.test.ts tests/unit/react/useMcp-reference-stability.test.tsx` (18 tests passed).
- Ran full client test suite: `pnpm --filter @mcp-use/client test` (57 files, 359 tests passed).
- Ran formatting and linting: `pnpm format && pnpm lint` (clean, 0 warnings/errors).
- Ran build: `pnpm --filter @mcp-use/client build` (clean build with types generated).

## Backwards Compatibility

Fully backwards compatible. Eliminates reconnection loops when passing inline `clientInfo` or `proxyConfig`.

## Related Issues

Closes #2402


libraries/typescript/packages/client/src/react/useMcp.ts
```diff
@@ -148,13 +149,21 @@ export function useMcp(options: UseMcpInternalOptions): UseMcpResult {
...
-    oauth: oauthOptions,
+    oauth: rawOauthOptions,
   } = options;
+
+  const clientInfo = useStableValue(options.clientInfo);
+  const proxyConfig = useStableValue(rawProxyConfig);
+  const headers = useStableValue(rawHeadersOption ?? {});
+  const clientOptions = useStableValue(rawClientOptions);
+  const oauthOptions = useStableValue(rawOauthOptions);
+  const reconnectionOptions = useStableValue(rawReconnectionOptions);
```
khandrew1: Please limit this fix to `clientInfo` and `proxyConfig`, the existing object dependencies responsible for the reported reconnect loop. Fresh `headers` and `clientOptions` references can recreate `connect()`, but `connect` itself is not a dependency of the lifecycle effect. OAuth already uses primitive dependencies, and `reconnectionOptions` is not a lifecycle dependency.

Please remove the unrelated stabilization and preserve the existing reconnect policy.

libraries/typescript/packages/client/src/react/useMcp.ts
```diff
@@ -2088,6 +2094,9 @@ export function useMcp(options: UseMcpInternalOptions): UseMcpResult {
     proxyConfig, // Triggers reconnection when proxy config (including headers) changes
     autoProxyFallbackConfig.proxyAddress,
     providedAuthProvider,
+    headers, // Triggers reconnection when custom headers legitimately change
+    effectiveClientOptions, // Triggers reconnection when client capabilities change
+    protocolNegotiation, // Triggers reconnection when protocol era negotiation mode changes
```
khandrew1: Please remove these three new dependencies from this fix. They introduce additional reconnect behavior, and an inline `protocolNegotiation: { pin: "2026-07-28" }` now creates a new effect dependency on every render, causing cleanup and reconnection even though its value is unchanged. This introduces the same bug the PR is intended to fix.

libraries/typescript/packages/client/src/react/useMcp-helpers.ts
```diff
@@ -12,6 +13,59 @@ export function assert(condition: unknown, message: string): asserts condition {
   }
 }
 
+/**
+ * Performs a deep structural equality check for primitives, arrays, and plain objects.
+ * Handles key ordering differences, null/undefined, and nested data structures.
+ */
+export function isEqualDeep(a: unknown, b: unknown): boolean {
```
khandrew1: Please remove `isEqualDeep` and `useStableValue`, along with the comparator-specific tests, and replace their usage with comparisons specific to `clientInfo` and `proxyConfig`.

For example, use the scalar client metadata fields as effect dependencies, serialize icons as tuples of `[src, mimeType, sizes]`, and compare `proxyAddress` plus serialized header entries sorted by key. Small named helpers for those configurations would also be reasonable.

Recreating `{ name: "Dashboard", version: "1.0.0" }` should keep the connection alive; changing the name should reconnect. Equivalent proxy headers should stay stable regardless of insertion order, while changing a header value should reconnect.

This also avoids applying enumerable-property comparison to SDK objects: different validator classes with prototype methods can compare equal here, causing `useStableValue` to retain the old `clientOptions.jsonSchemaValidator`.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
```diff
@@ -0,0 +1,381 @@
...
+  it("does not reconnect when parent re-renders with inline proxyConfig", async () => {
+    let triggerRender: () => void = () => {};
+
+    function TestComponent() {
+      const [count, setCount] = useState(0);
+      triggerRender = () => setCount((c) => c + 1);
+
+      useMcp({
+        url: "http://localhost:3000/mcp",
+        proxyConfig: { proxyAddress: "https://gateway.example.com" },
```
khandrew1: Please focus the regression tests on `clientInfo` and `proxyConfig`. Verify that equivalent inline configurations cause neither session closure nor reconnection, including nested icons and proxy headers supplied in different key orders.

Also verify that changing client metadata, icons, proxy address, or proxy header values reconnects with the updated configuration. The current proxy test only covers an unchanged address; the new reconnect expectations for top-level `headers` and `clientOptions` should be excluded from this scoped fix.

khandrew1 (changes requested):
Thanks for working on this. The reconnect bug is valid, but I’d like to narrow this PR to `clientInfo` and `proxyConfig`, which are the existing dependencies causing it.

Please remove `isEqualDeep` and `useStableValue` and compare the relevant values of those two configurations directly, including nested icons and proxy headers. Equivalent values in a new object should keep the connection alive; actual value changes should still reconnect.

Please also remove the new reconnect dependencies for `headers`, `clientOptions`, and `protocolNegotiation`. Those expand the behavior beyond this fix, and inline `protocolNegotiation` objects now introduce the same reconnect problem.

Keep the regression tests focused on unchanged inline configurations staying connected and actual changes reconnecting with the updated configuration. I’ve left inline comments with examples and suggested coverage.

rohith500:
@khandrew1 Thanks for the detailed review and guidance!

I have addressed all the feedback in commit `2f06e9a`:

1. **Narrowed Scope to `clientInfo` & `proxyConfig`**:
   - Removed `isEqualDeep` and `useStableValue` completely from `useMcp-helpers.ts` (along with their unit tests).
   - Removed the extra dependencies (`headers`, `effectiveClientOptions`, and `protocolNegotiation`) from the reconnect effect, preserving the existing reconnection policy.

2. **Deterministic Stabilization for `clientInfo`**:
   - Extracted scalar dependencies (`name`, `title`, `version`, `description`, `websiteUrl`) and added `serializeClientIcons(icons)` to serialize icons into deterministic `[src, mimeType, sorted sizes]` tuples.
   - Inline `clientInfo` literals and nested `icons` stay stable across parent re-renders, while legitimate metadata or icon changes trigger reconnection with the updated configuration.

3. **Deterministic Stabilization for `proxyConfig`**:
   - Added `serializeProxyHeaders(headers, customHeaders)` which merges headers and sorts keys alphabetically.
   - Inline `proxyConfig` objects and varying header key insertion orders remain reference-stable without triggering reconnects, while legitimate updates to `proxyAddress` or header values trigger reconnection with the updated configuration.

4. **Updated Regression Tests**:
   - Added unit tests in `useMcp-helpers.test.ts` for `serializeClientIcons` and `serializeProxyHeaders`.
   - Updated `useMcp-reference-stability.test.tsx` to verify:
     - Inline `clientInfo` and nested `icons` do not reconnect across re-renders.
     - Changing `name` or `icons` reconnects with updated configuration.
     - Inline `proxyConfig` and reordered proxy header keys do not reconnect.
     - Changing `proxyAddress` or a header value reconnects with updated configuration.
     - Removed top-level `headers` and `clientOptions` reconnect tests.

All 57 client test suites (359 tests) pass, formatting and lint checks are green, and the client build succeeds cleanly.

libraries/typescript/packages/client/src/react/useMcp-helpers.ts
```diff
@@ -12,6 +12,43 @@ export function assert(condition: unknown, message: string): asserts condition {
...
+ */
+export function serializeClientIcons(
+  icons?: Array<{ src: string; mimeType?: string; sizes?: string[] }>
+): string {
+  if (!icons || icons.length === 0) return "";
+  return JSON.stringify(
+    icons.map((icon) => [
+      icon.src,
+      icon.mimeType ?? "",
+      icon.sizes ? [...icon.sizes].sort() : [],
```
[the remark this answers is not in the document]
rohith500: Resolved in bbd7216: preserved absent versus empty values (`null` for omitted `mimeType` and `sizes`, empty string/array when provided) while retaining deterministic sorting.

libraries/typescript/packages/client/src/react/useMcp.ts
```diff
@@ -155,6 +157,40 @@ export function useMcp(options: UseMcpInternalOptions): UseMcpResult {
...
+  const rawProxyAddress = rawProxyConfig?.proxyAddress;
+  const serializedProxyHeaders = serializeProxyHeaders(
+    rawProxyConfig?.headers,
+    rawProxyConfig?.customHeaders
+  );
+
+  const proxyConfig = useMemo(() => {
+    if (!rawProxyConfig) return undefined;
+    return rawProxyConfig;
+  }, [rawProxyAddress, serializedProxyHeaders]);
```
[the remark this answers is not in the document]
rohith500: Resolved in bbd7216: normalized the retained `proxyConfig` so `headers` is always populated with the merged headers, and distinguished `headers` vs `customHeaders` in the memo identity.

libraries/typescript/packages/client/src/react/useMcp.ts
```diff
@@ -155,6 +157,47 @@ export function useMcp(options: UseMcpInternalOptions): UseMcpResult {
...
+
+  const proxyConfig = useMemo(() => {
+    if (!rawProxyConfig) return undefined;
+    const mergedHeaders =
+      rawProxyConfig.headers || rawProxyConfig.customHeaders
+        ? { ...rawProxyConfig.customHeaders, ...rawProxyConfig.headers }
+        : undefined;
+    return {
+      ...rawProxyConfig,
+      ...(mergedHeaders !== undefined ? { headers: mergedHeaders } : {}),
```
khandrew1: Please remove this normalization and retain the raw proxy configuration in the memo. The updated serializer already compares `headers` and `customHeaders` separately, which is sufficient to fix stale references. Merging `customHeaders` into `headers` additionally changes the request behavior: a token supplied only in `customHeaders` was previously ignored by this hook, but will now be sent. That also means removing `headers` while retaining `customHeaders` keeps sending the token. Let’s preserve the existing behavior in this scoped fix and adjust the new migration test accordingly.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
```diff
@@ -0,0 +1,515 @@
...
+    };
+
+    // Re-render multiple times with new inline clientInfo object instances
+    for (let i = 0; i < 5; i++) {
+      await reRender();
+    }
+
+    // Must NOT have added server / reconnected again
+    expect(addServerCalls.length).toBe(initialAddCount);
+    expect(removeServerCalls.length).toBe(0);
```
khandrew1: The hook disconnects through `closeSession`, not `removeServer`, so this assertion does not establish that the session stayed open. Before unmounting, assert `expect(sharedClient.closeSession).not.toHaveBeenCalled()` and `expect(sharedClient.connect).toHaveBeenCalledTimes(1)` in each stability test. Also require the initial `addServer` count to equal one instead of merely being greater than zero. Keep the changed-value tests that inspect the updated configuration passed to `addServer`; those are useful.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
```diff
@@ -0,0 +1,515 @@
...
+
+    root.unmount();
```
khandrew1: Please unmount with `await act(async () => { root.unmount(); })`, preferably in `afterEach` or `finally` so cleanup also runs after failed assertions. In isolated verification of the previous revision, adding the stronger `closeSession` assertions exposed two failures caused by prior tests’ unmount cleanup running after mocks were cleared. Awaiting unmount inside `act` resolved those failures. The same bare unmount pattern remains here; please update it consistently across the file.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
```diff
@@ -0,0 +1,515 @@
...
+    await flushMicrotasks();
+
+    expect(addServerCalls.length).toBe(2);
+    expect(addServerCalls[1].headers["x-api-key"]).toBe("key_v2");
+    expect(sharedClient.closeSession).toHaveBeenCalled();
+
+    root.unmount();
+  });
+
+  it("reconnects when an icon changes from omitted sizes to explicit empty array", async () => {
```
khandrew1, line 550: The new optional-sizes test is useful, but please also cover the actual default-icon transitions: omit `clientInfo.icons`, then set it to `[]`, and test the reverse. Assert that the next connection receives an empty list or the restored default logo respectively. For proxy headers, keep the migration test and add the removal case: start with the same token in both fields, then remove `headers` while retaining `customHeaders`. With the existing request behavior preserved, the next connection should omit that token. These transitions caught the previous regressions and should remain protected at the hook level.

khandrew1 (changes requested):
Thanks for narrowing the implementation. The latest commit now distinguishes omitted versus empty icons and preserves the two proxy header sources in the comparison key, which addresses the serializer issues.

There are a few remaining changes before I’m comfortable approving:

- Please remove the new merging of `customHeaders` into the request’s `headers`. Comparing the two maps separately is sufficient for this fix; merging them changes the existing request behavior.
- Strengthen the stability tests to assert that `closeSession` is not called and that `connect` is called exactly once, rather than checking `removeServer`.
- Await unmount cleanup inside `act`, including when assertions fail.
- Add the missing hook-level transitions for clearing/restoring default icons and removing `headers` while an equivalent `customHeaders` entry remains. Keep checking the configuration passed to the next connection.

The scope and overall approach are now reasonable. I’ve left inline comments with the specific remaining changes; I’m not asking for a broader redesign.

libraries/typescript/packages/client/src/react/useMcp.ts
rohith500: Addressed in commit `4a0a0b7`. Removed the `customHeaders` normalization from the memo so `proxyConfig` retains the raw reference unchanged across re-renders when keys/values match. Existing request behavior is fully preserved, and the tests have been updated accordingly.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
rohith500: Addressed in commit `4a0a0b7`. Updated all stability tests to assert `expect(addServerCalls.length).toBe(1)` on initial connect, and before unmounting assert both `expect(sharedClient.closeSession).not.toHaveBeenCalled()` and `expect(sharedClient.connect).toHaveBeenCalledTimes(1)`.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
rohith500: Addressed in commit `4a0a0b7`. Added suite-level `mountedRoots` tracking cleaned up in `afterEach` via `await act(async () => { root?.unmount(); })` to guarantee unmount runs even if prior assertions fail. In addition, all inline unmount calls across the file have been updated to `await act(async () => { root.unmount(); })`.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
rohith500: Addressed in commit `4a0a0b7`. Added hook-level test coverage verifying the default-icon transition when omitting `clientInfo.icons` (receives default logo `https://mcp-use.com/logo.png`), updating to `[]` (reconnects with empty list), and reversing back to omitted (reconnects with restored default logo). Also added the proxy header removal test starting with the token in both `headers` and `customHeaders`, then removing `headers` while retaining `customHeaders`, verifying that the next connection reconnects and omits the token.

rohith500:
@khandrew1 Thank you for the follow-up review! All remaining items have been addressed in commit `4a0a0b7`:

1. **Preserved Raw Proxy Config**: Removed the merging of `customHeaders` into `headers` within the memo. The raw `proxyConfig` object is now retained as-is across re-renders when keys and values match, preserving existing request behavior and ensuring headers in `customHeaders` alone are not inadvertently sent.
2. **Strengthened Stability Assertions**: Updated all stability tests to assert `expect(addServerCalls.length).toBe(1)` initially, and prior to unmounting assert both `expect(sharedClient.closeSession).not.toHaveBeenCalled()` and `expect(sharedClient.connect).toHaveBeenCalledTimes(1)`.
3. **Guaranteed Unmount Cleanup in Act**: Added suite-level `mountedRoots` tracking cleaned up in `afterEach` with `await act(async () => { root?.unmount(); })` to guarantee cleanup runs even if assertions fail. Also updated all inline unmount calls to `await act(async () => { root.unmount(); })`.
4. **Default Icon Transitions**: Added a comprehensive transition test covering:
   - Omit `clientInfo.icons` (receives default logo `https://mcp-use.com/logo.png`).
   - Set to `[]` (reconnects with empty list `[]`).
   - Reverse to omitted (reconnects and restores default logo).
5. **Proxy Header Removal Transition**: Updated the legacy migration test and added a removal test starting with the token in both `headers` and `customHeaders`, then removing `headers` while retaining `customHeaders`, verifying that the hook reconnects and omits the token from the next connection.

All 14 tests in `useMcp-reference-stability.test.tsx` and all 366 tests in `@mcp-use/client` are passing cleanly with zero ESLint/formatting regressions.

libraries/typescript/packages/client/src/react/useMcp.ts
```diff
@@ -155,6 +157,40 @@ export function useMcp(options: UseMcpInternalOptions): UseMcpResult {
...
+
+  const rawProxyAddress = rawProxyConfig?.proxyAddress;
+  const serializedProxyHeaders = serializeProxyHeaders(
+    rawProxyConfig?.headers,
+    rawProxyConfig?.customHeaders
+  );
+
+  const proxyConfig = useMemo(() => {
+    if (!rawProxyConfig) return undefined;
+    return rawProxyConfig;
```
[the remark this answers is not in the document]
rohith500: Addressed in commit `cef08a6`. Returned a shallow copy (`{ ...rawProxyConfig }` and `{ ...rawClientInfo }`) so that in-place mutations of existing configuration objects produce a new object identity when serialized dependencies change, allowing the reconnect effect to trigger. Added regression test coverage for in-place `proxyConfig` mutations in `useMcp-reference-stability.test.tsx`.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
```diff
@@ -0,0 +1,799 @@
+// @vitest-environment jsdom
```
khandrew1: Please remove this file and keep the focused helper tests. The hook behavior has been validated during review, and I’d prefer to keep this fix small rather than iterate further on the test suite.

khandrew1 (changes requested):
Thanks for addressing the changes. I think this 800 line test file has too much coverage for this PR and doesn't give us much. Please remove it and just keep the focused helper tests.

libraries/typescript/packages/client/tests/unit/react/useMcp-reference-stability.test.tsx
rohith500: Removed in commit 00f2ef7. Retained the focused unit tests in `useMcp-helpers.test.ts` as requested.

rohith500:
Thanks @khandrew1! I’ve removed `useMcp-reference-stability.test.tsx` in commit `00f2ef7` and kept only the focused helper unit tests in `useMcp-helpers.test.ts`. All client tests, lint, formatting, and build checks are 100% green.

khandrew1 (approved):
LGTM! Thanks

PR #2410, opened 2026-09-02
docs(skills): point the openapi-to-mcp handoff at a skill that exists

MohammedAlkindi:
`skills/openapi-to-mcp/SKILL.md` twice tells the reader to switch to a `build-mcp-app` skill. That name appears nowhere else in the repo and no such skill ships here: `skills/` contains `chatgpt-app-builder`, `mcp-apps-builder`, `mcp-builder` and `openapi-to-mcp`, and those four names are what the SKILL.md frontmatter and `.claude-plugin/marketplace.json` list.

An agent following the handoff looks for a skill that does not exist, so it either stalls or stays on the tools-only path when the user asked for widgets.

Changed both to `mcp-apps-builder`. That skill and `chatgpt-app-builder` are byte-identical apart from the `name:` field, so their content cannot settle which was meant. Three other things do: the same sentence already routes widgets to `--template mcp-apps`, the second reference is about a "widget-driven MCP App", and `create-mcp-use-app` installs `mcp-apps-builder` for that template (`src/index.ts:260-269`) while never referencing `chatgpt-app-builder` anywhere.

Say the word if you meant the ChatGPT one and I will switch it.


khandrew1 (approved):
Awesome, thanks for catching this!

PR #2421, opened 2026-09-03
fix(client): clean up sandbox message listener and cancel readiness wait on unmount in ViewRenderer

### LINKED ISSUE #2414: client: ViewRenderer leaks message listener on unmount and hangs indefinitely on failed sandbox handshake
## Description

In `ViewRenderer.tsx`, `waitForSandboxProxyReady(iframe)` attaches a `message` event listener to `window` with no cleanup on abort or unmount. When `ViewRenderer` unmounts prior to receiving `SANDBOX_PROXY_READY`, the listener is permanently leaked on `window`.

Additionally, if the iframe fails to load (e.g. 404, script error before proxy initialization, or strict CSP blockage), `waitForSandboxProxyReady` has no timeout or rejection mechanism. As a result, `await readyPromise` stays suspended indefinitely in memory, leaving the component permanently in `{ status: "connecting" }` without transitioning to `{ status: "error" }` or invoking `onError`.

## Reproduction

```tsx
import React, { useState } from "react";
import { ViewRenderer } from "@mcp-use/client/react";

export function App() {
  const [showView, setShowView] = useState(true);

  return (
    <div>
      <button onClick={() => setShowView(false)}>Unmount View</button>
      {showView && (
        <ViewRenderer
          viewId="test-view"
          source={{ kind: "preloaded", html: "<html>...</html>" }}
          sandboxUrl={new URL("https://sandbox.example/unreachable")}
        />
      )}
    </div>
  );
}
```

**Steps to Reproduce:**
1. Render `ViewRenderer` with a failing or slow sandbox URL.
2. Unmount the component while it is still waiting for `SANDBOX_PROXY_READY`.
3. Observe that the `message` event listener remains attached to `window`.
4. If left mounted, observe that `ViewRenderer` hangs indefinitely in `connecting` and never times out or fires `onError`.

## Root Cause

In `libraries/typescript/packages/client/src/react/view/ViewRenderer.tsx`:
1. `waitForSandboxProxyReady` (lines 73–86) creates a `Promise` and adds a `message` listener to `window` without accepting an `AbortSignal` or implementing unmount cleanup.
2. If `SANDBOX_PROXY_READY` is never received, the promise never resolves/rejects, keeping the bridge initialization effect suspended forever.

## Suggested Fix

1. Add `signal?: AbortSignal` and `timeoutMs?: number` (default 15s) to `waitForSandboxProxyReady` with a guaranteed cleanup callback that removes the listener and clears any pending timer.
2. Pass an `AbortSignal` from `ViewRendererBase`'s effect and call `abort()` on unmount.
3. If the handshake times out, reject the promise so `setLoadError`, `onError`, and `onLifecycleChange({ status: "error" })` are triggered.

rohith500:
# Pull Request Description

## Language / Project Scope

Check all that apply:
- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Cleans up global window `message` event listeners and cancels pending sandbox proxy readiness waits upon `ViewRenderer` unmount or effect teardown.

## Review Readiness

## Implementation Details

1. Added `options?: { signal?: AbortSignal }` to `waitForSandboxProxyReady` in `ViewRenderer.tsx`.
2. Created a guaranteed `cleanup()` callback that detaches the `message` event listener on `window` and cleans up signal listeners whenever the promise resolves or aborts.
3. Connected an `AbortController` in `ViewRendererBase`'s lifecycle effect, invoking `abortController.abort()` on unmount or effect replacement.
4. Ensures unmount cancellations are handled silently without triggering false `onError` callbacks or erroneous error lifecycle states, while allowing mounted views to wait for initialization without an arbitrary timeout deadline.

---

### Packages Modified

- [ ] `docs`
- [ ] `tests`
- [ ] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [x] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

---

## Example Usage (Before)

```tsx
// If ViewRenderer unmounted before receiving SANDBOX_PROXY_READY, window message listener was permanently leaked.
<ViewRenderer viewId="test" source={source} sandboxUrl={sandboxUrl} />
```

## Example Usage (After)

```tsx
// Unmounting cleanly detaches window message listeners and cancels pending wait silently.
<ViewRenderer viewId="test" source={source} sandboxUrl={sandboxUrl} />
```

## Documentation Updates

* N/A (Internal lifecycle bugfix and listener cleanup)

## Testing

- Added `packages/client/tests/unit/view-renderer-sandbox-cleanup.test.tsx` testing:
  - Exact window `message` event listeners are removed and cancelled silently when unmounting prior to sandbox ready.
  - Event listener is detached upon receiving `SANDBOX_PROXY_READY`.
  - Slow initialization beyond 15s allows mounted views to wait without timeout and continues when ready.
  - Effect replacement before readiness removes previous wait's exact listener and cleanly accepts readiness on replacement effect.
  - Unmounting during blob loading aborts cleanly without attaching readiness listeners or resuming initialization.
  - All listener assertions match the exact function instance registered by the wait.
- Ran all related ViewRenderer tests: `pnpm --filter @mcp-use/client test tests/unit/view-renderer*` (3 files, 14 tests passed).
- Ran formatting and linting: `pnpm format && pnpm lint` (0 errors, 0 warnings).

## Backwards Compatibility

Fully backwards compatible.

## Related Issues

Closes #2414

### DIFF
--- a/libraries/typescript/.changeset/fix-viewrenderer-sandbox-cleanup.md
+++ b/libraries/typescript/.changeset/fix-viewrenderer-sandbox-cleanup.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Clean up sandbox proxy message listener on unmount and add timeout in `ViewRenderer`.
--- a/libraries/typescript/.changeset/fix-viewrenderer-sandbox-cleanup.md
+++ b/libraries/typescript/.changeset/fix-viewrenderer-sandbox-cleanup.md
@@ -2,4 +2,4 @@
 "@mcp-use/client": patch
 ---
 
-Clean up sandbox proxy message listener on unmount and add timeout in `ViewRenderer`.
+Clean up sandbox proxy message listener and cancel pending readiness wait on `ViewRenderer` unmount.
--- a/libraries/typescript/packages/client/src/react/view/ViewRenderer.tsx
+++ b/libraries/typescript/packages/client/src/react/view/ViewRenderer.tsx
@@ -70,13 +70,17 @@ function CloseIcon() {
   );
 }
 
-function waitForSandboxProxyReady(
+/** @internal */
+export function waitForSandboxProxyReady(
   iframe: HTMLIFrameElement,
-  options?: { signal?: AbortSignal }
+  options?: { signal?: AbortSignal; timeoutMs?: number }
 ): Promise<void> {
   return new Promise((resolve, reject) => {
+    let timer: ReturnType<typeof setTimeout> | undefined;
+
     const cleanup = () => {
       window.removeEventListener("message", listener);
+      if (timer !== undefined) clearTimeout(timer);
       options?.signal?.removeEventListener("abort", onAbort);
     };
 
@@ -100,6 +104,17 @@ function waitForSandboxProxyReady(
       return;
     }
     options?.signal?.addEventListener("abort", onAbort, { once: true });
+
+    if (options?.timeoutMs !== undefined && options.timeoutMs > 0) {
+      timer = setTimeout(() => {
+        cleanup();
+        reject(
+          new Error(
+            `Sandbox proxy did not become ready within ${options.timeoutMs}ms`
+          )
+        );
+      }, options.timeoutMs);
+    }
 
     window.addEventListener("message", listener);
   });
--- a/libraries/typescript/packages/client/src/react/view/ViewRenderer.tsx
+++ b/libraries/typescript/packages/client/src/react/view/ViewRenderer.tsx
@@ -73,15 +73,11 @@ function CloseIcon() {
 /** @internal */
 export function waitForSandboxProxyReady(
   iframe: HTMLIFrameElement,
-  options?: { signal?: AbortSignal; timeoutMs?: number }
+  options?: { signal?: AbortSignal }
 ): Promise<void> {
-  const timeoutMs = options?.timeoutMs ?? 15_000;
   return new Promise((resolve, reject) => {
-    let timer: ReturnType<typeof setTimeout> | undefined;
-
     const cleanup = () => {
       window.removeEventListener("message", listener);
-      if (timer !== undefined) clearTimeout(timer);
       options?.signal?.removeEventListener("abort", onAbort);
     };
 
@@ -105,15 +101,6 @@ export function waitForSandboxProxyReady(
       return;
     }
     options?.signal?.addEventListener("abort", onAbort, { once: true });
-
-    if (timeoutMs > 0 && timeoutMs !== Infinity) {
-      timer = setTimeout(() => {
-        cleanup();
-        reject(
-          new Error(`Sandbox proxy did not become ready within ${timeoutMs}ms`)
-        );
-      }, timeoutMs);
-    }
 
     window.addEventListener("message", listener);
   });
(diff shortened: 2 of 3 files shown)


khandrew1 (changes requested):
The listener cleanup fix looks good: tying the readiness wait to the effect's AbortController removes the listener on teardown and keeps cancellation silent. I ran the three related test files; all 13 tests passed.

Please remove the sandbox readiness timeout before merging. As discussed in https://github.com/mcp-use/mcp-use/issues/2414#issuecomment-5518279805, mounted views should allow slow sandbox initialization to complete without an arbitrary deadline. The current `options?.timeoutMs ?? 15_000` in `ViewRenderer.tsx` applies a 15-second timeout at both call sites because they pass only a signal. A slow but valid sandbox therefore loses its readiness listener, triggers `onError` and an error lifecycle event, and has its iframe replaced by the error UI.

Commit `714476241` removed the timeout as requested, but the subsequent commits restored it, with `5f2efa3d8` reinstating the default. Please remove the timeout option/timer while retaining abort cleanup. The PR description and issue reply currently describe waiting without a timeout, so please keep them aligned with the final code.

Please also update/add regression coverage for:

- Slow initialization while mounted: replace the timeout-failure assertions with a test that advances beyond 15 seconds, verifies no `onError` or error lifecycle event and that the readiness listener remains attached, then delivers `SANDBOX_PROXY_READY` and verifies initialization can continue.
- Effect replacement before readiness: change a dependency that restarts the bridge effect, verify the previous wait's exact listener is removed and cancellation is silent, and verify the replacement wait can receive readiness and continue initialization.
- Unmount during blob loading: defer the blob fetch or response body, unmount before it resolves, then resolve it and verify no readiness listener is attached, no iframe initialization resumes, and no `onError`, error lifecycle event, or unhandled rejection occurs.

For listener cleanup assertions, please match the actual listener function registered by the wait rather than only checking listener counts or removal of any function.

rohith500:
@khandrew1 Thank you for the detailed feedback! All requested changes have been addressed in commit `868e704`:

1. **Removed Readiness Timeout**: Completely removed `timeoutMs` and the `setTimeout` timer from `waitForSandboxProxyReady` in `ViewRenderer.tsx`. Sandbox readiness now waits without an arbitrary deadline while strictly maintaining abort cleanup on unmount/effect replacement.
2. **Slow Initialization While Mounted**: Replaced timeout tests with a test that advances fake timers past 30s, verifies no `onError` or error status occurs and that the readiness listener remains attached, then delivers `SANDBOX_PROXY_READY` and verifies initialization completes.
3. **Effect Replacement Before Readiness**: Added a test verifying that when a dependency changes (e.g. `viewId`), the previous wait's exact listener function is removed silently without error, and the replacement wait can receive readiness and continue initialization.
4. **Unmount During Blob Loading**: Added a test deferring the HTML blob fetch, unmounting prior to resolution, then resolving the fetch and verifying no readiness listener is attached, `iframe.srcdoc` is never set, and no error occurs.
5. **Exact Listener Function Assertions**: All listener cleanup assertions now match the exact callback function registered by the readiness wait rather than relying on invocation counts.

All 14 tests across the 3 ViewRenderer suites (`view-renderer.test.ts`, `view-renderer-handler-stability.test.tsx`, `view-renderer-sandbox-cleanup.test.tsx`) are passing cleanly, and ESLint / Prettier formatting is fully verified.

khandrew1 (approved):
Thanks for the contribution!

PR #2422, opened 2026-09-03
test(cli): give the CLI integration fixtures a real install

### LINKED ISSUE #2419: Give CLI integration tests complete, isolated fixture projects
Our CLI integration tests effectively create a tiny customer project and run `mcp-use dev`, `build`, or `typecheck` against it.

There are two separate pieces of code involved:

- The test itself runs inside the repository and has access to the repository’s dependencies.
- The generated customer project imports packages such as `mcp-use` and `zod`, which need to be available from that project’s location.

Currently, we create the customer project without properly installing its dependencies. The helper links `mcp-use`, while other dependencies resolve from the surrounding repository. For example, the basic fixture imports `mcp-use` and `zod` despite declaring no dependencies.

This works because the fixture lives under `packages/cli/tests/cli/.tmp`. Moving it to the OS temporary directory exposes the incomplete setup: the project can no longer find those packages.

PR #2373 exposed this dependency on the repository. Its latest revision moves fixtures to `os.tmpdir()` and explicitly links `zod`, `typescript`, `react`, and `react-dom` from the repository’s installation into each fixture, alongside the existing `mcp-use` link. That restores access, but the fixture still depends on the repository’s installation and a separate dependency list maintained in the helper.

We should instead prepare these fixtures as complete customer projects:

1. Create the project in a unique OS temporary directory.
2. Declare the dependencies it needs.
3. Install those dependencies, using packed local packages so the tests exercise the code under review.
4. Run the existing CLI tests against that prepared project and clean up afterward.

This lets the project run outside the repository because it has its own complete installation. It also prevents missing fixture dependencies from being silently supplied by the repository.

Most other packages already use temporary directories successfully because they put data there—JSON files, images, or configuration—rather than applications that import packages.

The initial scope should be the shared CLI fixture helper and the dev/build/typecheck tests that use it. Preserve existing test coverage and reuse the installation approach already present in `scripts/verify-release-install.mjs` where appropriate. Carry forward the cross-platform path-normalization findings from #2373, including macOS symlinked temp paths and Windows short paths.

## Acceptance criteria

- Runnable fixtures declare and receive their required dependencies.
- Tests pass from OS temporary directories without manually linking back to the repository’s `node_modules`.
- Concurrent runs and worktrees have isolated mutable state.
- Generated fixture files stay outside the repository and are cleaned up.
- Measure setup cost and reuse preparation where practical instead of installing everything separately for every test.

ayaangazali:
Closes #2419.

## Why

The fixtures are customer projects. They import `mcp-use` and `zod`, they declared neither, and they installed nothing. That worked only because they sat under `packages/cli/tests/cli/.tmp`, inside the repository, and resolved by walking up into its `node_modules`. #2373 moved them out and the incomplete setup surfaced immediately as `Cannot find module 'zod'`.

## What this does

Each fixture manifest now declares what the project needs, and `globalSetup` prepares it once per run:

1. `pnpm --filter mcp-use pack`, so the project installs the build under test rather than whatever the registry would return for `mcp-use`
2. copy the fixture to a temp directory and rewrite **only** its `mcp-use` entry to `file:<tarball>`
3. `npm install`, which resolves everything else from the fixture's own manifest
4. tests copy from the prepared template; teardown removes it

The manifest is the single list of dependencies, so there is no parallel list in the helper to keep in sync. Nothing reaches back into the repository's `node_modules`, which is what lets the scratch copies live in `os.tmpdir()`. The `tests/cli/.tmp` entry in `vitest.config.ts` is gone with them, and no ESLint ignore or `.gitignore` entry is needed because nothing is written inside the repo.

## Two details carried over from #2373

**`TMP_ROOT` resolves through `realpathSync.native`.** macOS reports `/var/folders/...` for a directory that lives at `/private/var/folders/...`, and Windows reports an 8.3 short path. Vite compares the project root against resolved file paths and serves anything it believes is outside the root through `/@fs/`, which changes the URLs a fixture's own views come back under. `dev.test.ts` asserts one of those URLs and caught it.

**Copies link the installed packages entry by entry rather than copying the tree.** The install is about 100MB and there are over a hundred fixture tests, so copying it per test would add minutes. Linking the individual entries rather than `node_modules` itself leaves that directory writable, which matters because Vite puts its optimizer cache there and concurrent tests need their own.

## Acceptance criteria

- **Fixtures declare and receive their dependencies.** Both manifests list them; `npm install` provides them.
- **Tests pass from the OS temp directory without linking back to the repo.** The only links are into the prepared install.
- **Concurrent runs and worktrees isolated.** Each copy gets its own sources and its own writable `node_modules`; the template root is an `mkdtemp` per run.
- **Generated files stay outside the repo and are cleaned up.** Teardown removes the template root; `removeDir` still handles the per-test copies.
- **Setup cost measured and reused.** One pack plus two installs per run, not per test.

## Cost

```
[cli-tests] prepared 2 fixture projects in 10s     # warm npm cache
[cli-tests] prepared 2 fixture projects in 20s     # cold
```

Package suite: 294 passed, 23 files, same as before. Wall clock goes from about 15s to 26s warm, and the preparation is the whole difference.

## Validation

From `libraries/typescript/packages/cli`: `vitest run` — 294 passed. `tsc --noEmit` — clean.
From `libraries/typescript`: `eslint` and `prettier --check` on the changed files — clean.

I can only measure macOS locally. The Windows and Linux legs of `typescript/cli` are the real check on the two things that differ per platform here, the short-path resolution and spawning `npm`/`pnpm` (`.cmd` on Windows), and I will watch them on this PR rather than claim them.

---
## Summary by cubic
Fixes the CLI integration fixtures so they install their own dependencies (`mcp-use`, `zod`, `typescript`) instead of resolving them by walking up to the repository's `node_modules`. Closes #2419.

**What changed**
- Each fixture declares its dependencies; `globalSetup` installs one template per run by packing the local `mcp-use` build, rewriting its entry to the tarball, and running `npm install`.
- The committed manifests point `mcp-use` at a non-existent tarball so a manual `npm install` fails loudly rather than fetching the published framework.
- Packing fails with a clear error when `packages/server` is unbuilt.
- Tests copy from the template into `os.tmpdir()`, with the template root resolved through `realpathSync.native` so Vite does not serve fixture assets under `/@fs/`.
- Copies link installed packages entry-by-entry (directories as junctions, files copied) to keep `node_modules` writable for Vite's optimizer cache.
- Whole copies wait until worker exit because Vite's optimizer keeps writing after the dev server closes; paths inside a copy delete immediately.
- Each worker gets its own `mkdtemp` scratch root for isolation.
- On Windows, package managers run through `cmd.exe` with allow-listed, quoted arguments.
- The first skills reload in `dev.test.ts` gets a 30s timeout for Vite's pre-bundling.

**Cost**
- Setup adds about 11s on Linux, 5–30s on macOS, 87s on Windows.
- Suite still passes 294 tests.

<sup>Written for commit dadf0065b438fd62ece2062e2add8a9f3430da44. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2422?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


ayaangazali:
Status, and one open question for you @khandrew1 before I push anything else.

**Ubuntu is fully green.** Windows is at 293/294 with one failure, and I would rather ask than guess at it.

Progress across the three runs, each a real bug the matrix caught:

1. `spawnSync pnpm.cmd EINVAL`. Since the fix for CVE-2024-27980 node refuses to spawn `.cmd` shims without a shell, so the bare name is ENOENT and `.cmd` is EINVAL. Now goes through `cmd.exe` with quoted arguments on Windows only.
2. Ubuntu ran all 294 tests green and then failed the job on an unhandled rejection: rolldown could not write into `.mcp-use/cache/deps_temp_*` because the copy had already been deleted. Vite's optimizer keeps writing after the dev server is told to close. Whole copies now wait for the worker to exit; paths inside a copy still go immediately, since `build.test.ts` deletes `src/` during a test to set up the missing-entry case.
3. Now: `dev.test.ts > omits invalid skills on reload while preserving valid siblings` times out after 15s on Windows. No `lastError`, so the reload simply never arrived in time.

**The setup cost you asked to measure:**

```
ubuntu   prepared 2 fixture projects in  ~11s      job 1m52s   294 passed
macOS    prepared 2 fixture projects in 5 to 20s   local       294 passed
windows  prepared 2 fixture projects in  87s       job 4m24s   293 passed
```

**The question.** I do not think that last failure is a broken reload. The basic fixture now has a real `node_modules` with mcp-use's transitive dependencies in it, where before it had a single symlink, so Vite has genuinely more to scan and pre-bundle and the whole file went from about 60s to 120s on Windows. The 15s wait is the thing that no longer fits, not the behaviour under test.

That leaves a tradeoff I would rather you make than me:

1. **Raise the wait for the skills-reload tests.** Smallest change, and honest about the extra work, but it is a timeout number I would be picking to fit one runner.
2. **Keep the optimizer out of it**, so the fixtures install their dependencies but the dev server does not pre-bundle them. Better runtime, but it needs a lever from the CLI side that does not exist today.
3. **Decide 87s of Windows setup is too much regardless** and prepare only the fixture a given test file needs, or prepare lazily per worker.

Option 3 is worth weighing on its own: this adds about 90s to the Windows leg whatever we do about the timeout. Tell me which way you want it and I will finish it out. Everything through step 2 above is pushed and Ubuntu is green on it.

ayaangazali:
Correcting my last comment: this is not general slowness, and a bigger timeout would be papering over something specific. Same run, no new push.

The other skills-reload tests on the same Windows job are fast:

```
✓ watches supporting files in a configured skills directory              566ms
✓ reloads skill files when the configured skills directory overlaps views 1705ms
× omits invalid skills on reload while preserving valid siblings        15820ms
```

So the reload path works on Windows. One test does not. `waitFor` also reported no `lastError`, meaning every poll of `skills/list` answered fine and the invalid skill was simply never dropped, so the watcher event for that one overwrite did not arrive.

What I ruled out while looking, in case it saves anyone repeating it:

- **node_modules being watched.** `resolveChokidarOptions` merges rather than replaces, so `**/node_modules/**` is always ignored even though `server.watch.ignored` in `dev.ts` sets a single pattern.
- **The optimizer cache being watched.** `cacheDir: paths.cache` is set, and Vite feeds the resolved cacheDir into the same ignore list.
- **The same-length rewrite problem from #2374.** The two writes here are 56 and 41 bytes.
- **Setup cost.** Prep is 87s on Windows, but it is charged once before any test runs and the siblings pass in under two seconds.

What is left is that this test overwrites a `SKILL.md` it wrote moments earlier, in place, while the passing siblings write to a different file or create a new one. That is the same family as #2374, just a different trigger, and it only became reachable once the project had a real install.

I am not pushing a fix for it. I cannot reproduce on Windows, every candidate above is ruled out by reading rather than by running, and anything I changed now would be me guessing at a number or a write pattern and using your CI to check it. Ubuntu is green and Windows is 293/294.

@khandrew1 the question from my previous comment still stands, with this correction folded in: given the 87s Windows setup cost, do you want this shape at all, or preparation scoped per worker or per test file? Happy to take the Windows test on separately once the shape is settled, ideally with someone who can run it locally.

ayaangazali:
One more correction, and it goes back the other way. My "this is not general slowness" comment leaned on a comparison that does not hold.

Measured locally, reproducibly, on the same branch:

```
omits invalid skills on reload while preserving valid siblings   2936ms
watches supporting files in a configured skills directory         628ms
reloads skill files when the configured skills directory overlaps views  640ms
```

The failing test is roughly five times heavier than the two I compared it against, on a machine where it passes. It drives three full reload cycles rather than one, and the first of them absorbs Vite's initial dependency pre-bundling for the project. So the siblings passing in under two seconds on Windows says nothing about this one, and I should not have drawn a conclusion from it.

The Windows failure is on the first `waitFor` at `dev.test.ts:277`, and it hit the wall at 15820ms against a 15000ms budget. A test with a 2.9s baseline, on a leg where this file runs 120s against 35s locally, sitting 800ms over the limit reads as headroom, not a missed event. My earlier reading of "no `lastError`" as proof the event never arrived was wrong: it only shows the predicate had not flipped yet when the clock ran out.

That puts option 1 from my earlier comment back on the table as the honest fix rather than a band-aid, since the per-reload work genuinely grew when the fixtures gained a real install.

I am still not pushing it, for the same reason as before: it is a number I would be choosing to fit one runner I cannot measure. If you want it, say what you would consider reasonable headroom for that first reload and I will set it there and confirm the leg goes green. If you would rather scope the preparation differently first, that changes the number anyway.

ayaangazali:
Reopened. I closed this by accident: a pre-commit changeset gate rejected the commit, my push ran anyway and briefly put the branch at the base commit, and GitHub closed the PR. The branch is back to one commit on current canary and the diff is unchanged. Sorry for the noise.

ayaangazali:
I ran a control for this, and it sharpens the picture in a way that is worth recording before you decide.

On #2448, a one line reorder in `servers.ts` that cannot touch `dev.test.ts`, the Windows leg failed on `reloads skill files when the configured skills directory overlaps views` at 15463ms. I re-pushed the identical tree with no content change and it went green. So the skills tests do sit close to the 15s wall on Windows on `canary` as it stands, and they do flake there.

This PR does not flake. Same experiment, identical tree re-pushed:

```
run 1   omits invalid skills on reload   15820ms
run 2   omits invalid skills on reload   15748ms
run 3   omits invalid skills on reload   15887ms
```

Three runs, same test every time, always about 0.85s over the budget, while the two sibling skills tests pass in the same job. So this is deterministic rather than flaky, and the install change is implicated for that one test specifically.

Two things follow. My earlier claim that the sibling comparison ruled out slowness was wrong, and I have already corrected that. But so was treating this as the same flakiness the rest of the file shows: #2448 recovered on a re-run and this one has not in three.

That leaves the same question, now better evidenced. A test with a 2.9s baseline locally, landing 0.85s over a 15s budget on a runner where the whole file takes 120s against 35s, reads as a budget that no longer fits the work rather than a broken reload. If you are happy with option 1, tell me the headroom you want and I will set it and confirm the leg. If the 87s Windows setup means you would rather scope the preparation differently first, that changes the number, so it makes sense to settle that first.

ayaangazali:
The timeout theory is dead, and I killed it with the obvious experiment rather than shipping it. Worth reporting because it rules out the option you and I were both leaning towards.

I raised that one wait from 15s to 30s and pushed. Windows:

```
× omits invalid skills on reload while preserving valid siblings   30509ms
Error: waitFor timed out after 30000ms
```

Twice the budget, same failure. So the reload never arrives on Windows for this test, and option 1 from my earlier comment would only have made the red slower to reach. I have reverted it, and the branch is back to the single fixture commit with no timeout change in the diff.

That also means my "budget no longer fits the work" reading was wrong, and the original reading of `waitFor` reporting no `lastError` was right: `skills/list` answers every poll and the invalid skill is simply never dropped.

What is now established:

- **Deterministic, not flakiness.** Four Windows runs, same test every time. The control on #2448 failed a different skills test once and went green on an identical re-push, so the file's ordinary flakiness looks nothing like this.
- **Not a budget.** 15s and 30s both time out.
- **Specific to this test.** The two sibling skills-reload tests pass in the same job.
- **Introduced by the install.** The same tests pass on Windows with the old symlink-only fixture.

What I have ruled out by reading rather than guessing: `node_modules` being watched (`resolveChokidarOptions` merges rather than replaces, so `**/node_modules/**` is always ignored), the optimizer cache being watched (`cacheDir: paths.cache` is fed into the same ignore list), the same-length rewrite class from #2374 (56 and 41 bytes), and the skills walker following links into the dependency tree (`walk()` starts at the skills root and skips symlinks).

The one difference left that I can see between this test and the two that pass: it overwrites a `SKILL.md` it wrote moments earlier, in place, under the conventional `skills/` directory, while the passing pair write a different file under a configured directory.

I am going to stop guessing here. I do not have a Windows machine, and every remaining hypothesis needs one to test. If someone who does can run `vitest run tests/cli/dev.test.ts -t "omits invalid skills"` on this branch, that will settle it faster than another round of CI. Otherwise I am happy to park this PR until the setup-cost question is decided, since that may change the shape anyway.

ayaangazali:
Moved this to draft so it stops reading as awaiting review. Nothing has changed in the diff; it is parked on the open question rather than abandoned.

Where it stands: the Windows leg fails one test, `omits invalid skills on reload`, deterministically across five runs, and it is not a timeout (15s and 30s both expire). Ruled out so far: `node_modules` being watched, the optimizer cache being watched, the same-length rewrite class from #2374, the skills walker following links, a path-shape or casing mismatch in the skills comparison, the single-shared-junction layout (which breaks a views test locally), and today the skill name-versus-directory validation, which uses `basename(resolve(path, ".."))` and is platform-correct.

What is left needs a Windows machine I do not have. If anyone can run `vitest run tests/cli/dev.test.ts -t "omits invalid skills"` on this branch I will take it from there.

@khandrew1 the question from #2419 still gates the shape anyway: Windows setup is 87s, so if you would rather the fixture preparation be scoped per worker or per test file, that changes this code before the Windows test is worth chasing. Mark it ready whenever you want it back in the queue.

khandrew1:
@ayaangazali thanks for converting this to a draft. I still need to think about how exactly we should handle this properly. We should probably take inspiration from how other open source repos handle testing their create CLIs.

ayaangazali:
Rebased onto current canary and re-ran, because the branch had fallen behind and I had never eliminated base drift as the cause. It is not the cause.

```
[cli-tests] prepared 2 fixture projects in 81s
× omits invalid skills on reload while preserving valid siblings   15715ms
Tests  1 failed | 298 passed (299)
```

Canary's own `typescript/cli (windows-latest)` leg is green on its last two runs, and the branch now contains everything canary has, so the failure is genuinely this diff and not the base. Sixth Windows run, same single test, same margin. Ubuntu passes in 1m22s.

That leaves the position unchanged but with one more variable closed. Established: deterministic across six runs, not a budget (15s and 30s both expire), specific to this one test while both sibling skills-reload tests pass in the same job, and introduced by giving the fixtures a real install.

Ruled out so far, each by reading or by experiment rather than guessing: `node_modules` being watched (`resolveChokidarOptions` merges rather than replaces, so `**/node_modules/**` is always ignored), the optimizer cache being watched (`cacheDir: paths.cache` is fed into the same ignore list), the same-length rewrite class from #2374 (the two writes are 56 and 41 bytes), the skills walker following links (it starts at the skills root and skips symlinks), a path shape or casing mismatch in the skills comparison (`assertSafeDirectory` returns an absolute path for both the conventional and configured cases, and all three skills tests share the comparison), the name-versus-directory validation (`basename(resolve(path, ".."))`, platform-correct), a single shared `node_modules` junction instead of per-entry links (breaks a views test locally), and base staleness (today).

Still parked on your call in #2419 about the 87s Windows setup, since scoping the preparation differently changes this code. If anyone with a Windows machine can run `vitest run tests/cli/dev.test.ts -t "omits invalid skills"` on this branch, that will settle it faster than another CI round.

ayaangazali:
Six cubic threads were sitting unanswered here because my own sweep filtered bot reviewers out. That was my mistake, and the findings were good. All six are addressed, five by fixing them.

**The link loop junctioned plain files.** `readdirSync` over the template's `node_modules` includes npm's `.package-lock.json`, and a junction can only point at a directory. On macOS node ignores the type and makes an ordinary symlink, which is why this passed locally for a week. Entries are now handled by type: directories junctioned, plain files copied. The one part I would qualify is "`copyFixture` fails for every fixture" — 300 Windows tests pass, so it is evidently not fatal there, but the type mismatch was real and is gone.

**`TMP_ROOT` was one fixed directory.** Correct, and it contradicted the isolation this PR claims: concurrent runs and worktrees on the same host shared it, and an interrupted run left copies behind permanently. It is now an `mkdtemp` per worker, removed whole at worker exit, which also lets the fixed-path backstop in teardown go.

**`pnpm pack` does not build.** Correct and worth guarding: `files` is `["dist"]` and `prepack` only checks the README, so packing an unbuilt framework produces a tarball with nothing in it. `packFramework` now fails with a clear message when `packages/server/dist/bin.js` is missing, rather than letting the fixtures install an empty `mcp-use`.

**`quoteForCmd` under-counted.** Correct: `(`, `)` and `%` all mean something to `cmd.exe`, and `%` expands inside double quotes. It now allow-lists the plainly safe characters instead of enumerating dangerous ones, and escapes `%` as `%%`.

**The `mcp-use: "*"` placeholder.** Correct that a stray `npm install` in a fixture directory would fetch the published framework and silently test the wrong code. The manifests now point at a tarball path that does not exist, so that fails loudly.

**The `filename` shape.** I verified pnpm's `pack --json` returns an absolute path today, so the premise does not hold on this version. Rather than argue the point I took the robustness half: the path is now resolved against the destination, which is a no-op for an absolute path and correct if a version ever returns a bare name.

None of this fixed the Windows test, which I did not expect it to: still `omits invalid skills on reload` at 15826ms, 300 passed. `tests/cli/` locally is 117 passed, and the whole file suite is green.

khandrew1:
Thanks for investigating this @ayaangazali! Closing in favor of #2524 as a bandaid fix, as well as draft PR #2537 which attempts to standardize the lifecycle with a project fixture

PR #2423, opened 2026-09-03
fix(cli): warn on env set secret downgrade, document --secret required on every write

### LINKED ISSUE #2388: cli: rotating an env variable without --secret silently clears its sensitive flag

aishwary-dongre:
### Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

### Changes

Follow-up to #2389 (closed), addressing @khandrew1's feedback on #2388. Scoped to help text, docs, and the success message. `--secret` stays an explicit opt-in required on every write, and the write contract is unchanged.

- `env set --help` and the CLI reference now state that `--secret` must be passed on **every** write to a variable, including updates and rotations, and that omitting it stores the value as non-sensitive even if the variable was previously write-only.
- The success message reports the resulting state: `Set TOKEN (saved as write-only).` when `--secret` is passed, `Set TOKEN.` otherwise.

No downgrade detection, no warning branch, no new JSON fields, no new dependencies.

### Review Readiness

### Implementation Details

`envSet()` in `servers.ts` builds the request body with `sensitive: values.secret === true` as before. The only change is the message passed to `printResult()`, which depends solely on whether `--secret` was passed on this invocation. `--secret`'s help text in the detailed `env set --help` block was reworded; the top-level command summary is unchanged.

`docs/v2/typescript/api-reference/cli-reference.mdx` carries the same clarification.

#### Packages Modified

- [x] docs
- [x] tests
- [x] cli
- [ ] create-mcp-use-app
- [ ] mcp-use (server)
- [ ] mcp-use (client)
- [ ] inspector

#### Pre-commit Checklist

- [x] Ran `pnpm lint:fix` to auto-fix linting issues (via pre-commit hook)
- [x] Ran `pnpm format` to format code with Prettier (via pre-commit hook)
- [x] Ran `pnpm changeset` to create a changeset
- [x] Added or updated tests
- [x] Updated documentation in `docs/` folder

### Example Usage (Before)

```bash
mcp-use servers env set srv_abc123 TOKEN=old --secret   # Set TOKEN.
mcp-use servers env set srv_abc123 TOKEN=rotated        # Set TOKEN.  (stored non-sensitive, not obvious from the output)
```

### Example Usage (After)

```bash
mcp-use servers env set srv_abc123 TOKEN=old --secret   # Set TOKEN (saved as write-only).
mcp-use servers env set srv_abc123 TOKEN=rotated        # Set TOKEN.
```

### Documentation Updates

`docs/v2/typescript/api-reference/cli-reference.mdx`: the `--secret` note now explains that the flag must be repeated on every write, including updates and rotations.

### Testing

`tests/commands/servers.test.ts` gained two cases that pin the existing write semantics: an update without `--secret` sends `sensitive: false` and reports `secret: false` in `--json`, and a brand-new variable is not marked sensitive by default.

Ran `pnpm exec vitest run tests/commands/servers.test.ts` in `packages/cli`: 8 tests pass. `pnpm build` at the workspace root succeeds.

### Backwards Compatibility

No behavior change. `--secret` remains the sole determinant of `sensitive` on every write, and `--json` output is unchanged. This PR only clarifies the help text and docs and makes the human-readable success message report the resulting state. Patch bump via changeset.

## Related Issues

Closes #2388

### DIFF
--- a/libraries/typescript/.changeset/itchy-worms-hear.md
+++ b/libraries/typescript/.changeset/itchy-worms-hear.md
@@ -1,5 +1,0 @@
----
-"@mcp-use/cli": patch
----
-
-Preserve the existing sensitive flag when rotating an environment variable without --secret
--- a/libraries/typescript/.changeset/itchy-worms-hear.md
+++ b/libraries/typescript/.changeset/itchy-worms-hear.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Preserve the existing sensitive flag when rotating an environment variable without --secret
--- a/libraries/typescript/packages/cli/src/commands/servers.ts
+++ b/libraries/typescript/packages/cli/src/commands/servers.ts
@@ -290,12 +290,14 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     (variable) =>
       variable.key === key && (variable.branch ?? undefined) === values.branch
   );
+  const sensitive =
+    values.secret === true ? true : (existing?.sensitive ?? false);
   const body = {
     key,
     value,
     branch: values.branch ?? null,
     environments: values.branch === undefined ? ["production"] : ["preview"],
-    sensitive: values.secret === true,
+    sensitive,
   };
   if (existing === undefined) {
     await api.request<unknown>(
@@ -313,7 +315,7 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     key,
     scope: values.branch === undefined ? "production" : "preview",
     branch: values.branch ?? null,
-    secret: values.secret === true,
+    secret: sensitive,
     updated: existing !== undefined,
   };
   printResult(result, json, `Set ${key}.`);
--- a/libraries/typescript/packages/cli/src/commands/servers.ts
+++ b/libraries/typescript/packages/cli/src/commands/servers.ts
@@ -290,8 +290,7 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     (variable) =>
       variable.key === key && (variable.branch ?? undefined) === values.branch
   );
-  const sensitive =
-    values.secret === true ? true : (existing?.sensitive ?? false);
+  const sensitive = values.secret === true || (existing?.sensitive ?? false);
   const body = {
     key,
     value,
--- a/libraries/typescript/packages/cli/src/commands/servers.ts
+++ b/libraries/typescript/packages/cli/src/commands/servers.ts
@@ -29,7 +29,7 @@ Commands:
   update <id-or-slug>          Update server metadata or build configuration
   delete <id-or-slug>          Delete a server
   env list <server>            List environment variable metadata
-  env set <server> <KEY=VALUE> Create or update an environment variable (--secret required on every write to keep it sensitive)
+  env set <server> <KEY=VALUE> Create or update an environment variable
   env unset <server> <key>     Delete an environment variable
 
 Run mcp-use servers <command> --help for all options.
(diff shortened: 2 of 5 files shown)


libraries/typescript/.changeset/itchy-worms-hear.md
```diff
@@ -0,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Preserve the existing sensitive flag when rotating an environment variable without --secret
```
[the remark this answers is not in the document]
aishwary-dongre: Good catch, that was a leftover changeset from the original PR (#2389) before the maintainer's redirect. Removed it in 93d7696, the current changeset (fancy-crabs-unite.md) accurately describes the actual behavior now shipped.

khandrew1 (changes requested):
Thanks for the contribution! Could we make it a little tighter:

- Keep the top-level command summary unchanged and shorten the `--secret` help text.
- Use `Set TOKEN (saved as write-only).` for secrets and `Set TOKEN (saved as non-sensitive).` otherwise.
- Shorten the downgrade warning to `Warning: TOKEN is no longer write-only.`
- Remove `downgradedFromSecret`; the existing `secret` JSON field already reports the resulting state.

Keep the existing write semantics unchanged.

aishwary-dongre:
Thanks, all five changes made: reverted the top-level summary line, shortened the --secret help text, updated both success messages to the new wording, shortened the downgrade warning, and removed downgradedFromSecret since the existing secret field already covers it. All 9 tests pass and the build succeeds.

libraries/typescript/packages/cli/src/commands/servers.ts
```diff
@@ -316,7 +319,12 @@ async function envSet(argv: readonly string[], json: boolean): Promise<number> {
     secret: values.secret === true,
     updated: existing !== undefined,
   };
-  printResult(result, json, `Set ${key}.`);
+  const message = downgradesExistingSecret
+    ? `Warning: ${key} is no longer write-only.`
+    : values.secret === true
+      ? `Set ${key} (saved as write-only).`
+      : `Set ${key} (saved as non-sensitive).`;
+  printResult(result, json, message);
```
khandrew1: Please simplify this to depend only on whether `--secret` was passed on this invocation. No comparison with the variable’s previous sensitivity is needed:

```suggestion
  printResult(
    result,
    json,
    values.secret === true
      ? `Set ${key} (saved as write-only).`
      : `Set ${key}.`
  );
```

khandrew1 (changes requested):
Thanks for the updates. To clarify my earlier feedback, I don’t want any downgrade detection or warning logic. Please keep this scoped to help/docs clarification and a simple success message:

- With `--secret`: `Set TOKEN (saved as write-only).`
- Without `--secret`: `Set TOKEN.`

Please remove `downgradesExistingSecret` and the warning branch, restore `sensitive: values.secret === true` directly in the request body, and remove the warning-specific test and documentation.

Keep the clarification that `--secret` must be passed on every write, including updates and rotations. Write semantics and JSON output should remain unchanged. Please also update the PR description, which still mentions `downgradedFromSecret`.

libraries/typescript/packages/cli/src/commands/servers.ts
aishwary-dongre: Applied exactly as suggested; the message now depends only on whether `--secret` was passed on this invocation.

aishwary-dongre:
Thanks for the clarification, that's much simpler. Removed the downgrade detection and the warning entirely:

- `sensitive: values.secret === true` goes straight into the request body again; `downgradesExistingSecret` is gone.
- Success message is `Set TOKEN (saved as write-only).` with `--secret` and `Set TOKEN.` without, based only on this invocation.
- Dropped the warning-specific test and the warning wording in the docs.
- Kept the clarification that `--secret` must be passed on every write, including updates and rotations, in both `env set --help` and the CLI reference.

Write semantics and `--json` output are unchanged. Updated the changeset and the PR description so neither mentions `downgradedFromSecret` any more. `pnpm exec vitest run tests/commands/servers.test.ts` passes (8 tests).

khandrew1 (approved):
LGTM!

PR #2437, opened 2026-09-03
fix(agent): apply per-call maxSteps to the model-call limit

### LINKED ISSUE #2413: agent: per-call maxSteps never reaches the model-call limit in local mode
## What happens

`MCPAgent.run({ prompt, maxSteps })` documents a per-call step budget, and telemetry records it as `maxStepsUsed`. In local mode it does not change how many model calls the agent is allowed to make.

The limit is enforced by middleware baked in when the executor is built (`mcp_agent_langchain.ts:563`):

```ts
const middleware = [modelCallLimitMiddleware({ runLimit: this.maxSteps })];
```

`createAgent()` runs during `initialize()`. Nothing rebuilds the executor when a later call asks for a different budget.

Measured with the mock harness in `tests/unit/telemetry/agent-telemetry.test.ts`, spying on the `runLimit` the middleware actually receives. Agent constructed with `maxSteps: 5`, then called with `maxSteps: 20`:

```
run()          runLimit seen: [5]    telemetry maxStepsConfigured: 5   maxStepsUsed: 20
streamEvents() runLimit seen: [5]    telemetry maxStepsConfigured: 20
```

So there are two separate gaps.

**`run()` / `stream()` ignore the argument entirely.** In the body from `mcp_agent_langchain.ts:1240` onward, the destructured `steps` is used in exactly two places: forwarding to the remote agent (`:1268`) and the telemetry field (`:1611`). It never updates `this.maxSteps`, so even `recursionLimit: this.maxSteps * 3` (`:1364`) uses the constructor value.

**`streamEvents()` updates the field but not the limit.** It does `this.maxSteps = steps ?? this.maxSteps` (`:1836`), which is why `maxStepsConfigured` moves to 20, but the executor is already built, so `runLimit` stays 5 and only `recursionLimit` changes.

Remote mode is unaffected: `steps` is passed to `remoteAgent.run` and honoured there. So the same call behaves differently depending on whether the agent is remote.

## Why it matters

The visible symptom is an agent that stops early with no explanation. You pass `maxSteps: 20` for a task that needs more than five model calls, telemetry confirms `maxStepsUsed: 20`, and the run still stops at five. Nothing in the logs contradicts the number you passed.

## Fix direction, and the part I would rather not decide alone

The mechanical part is clear: the executor has to be rebuilt when the effective limit changes, since `modelCallLimitMiddleware` takes its budget at construction. Something like recording the limit the current executor was built with, and rebuilding when a call asks for a different one.

What I do not want to guess at is whether a per-call `maxSteps` should mutate agent state. `streamEvents` already writes it back to `this.maxSteps`, so today one call with `maxSteps: 20` raises the ceiling for every later call that does not pass one. Reasonable arguments both ways, and it changes what the fix looks like:

1. **Per-call is per-call.** Use the argument for that invocation, leave `this.maxSteps` alone. Means `streamEvents` stops writing the field, which is a small behaviour change for anyone relying on it.
2. **Keep the write-back**, and just make the executor follow it. Smaller diff, keeps the current stickiness in `streamEvents` and extends it to `run`.

Happy to send a PR for either once you say which. Reproduced on `canary` at `b95fbc4e`.

ayaangazali:
Closes #2413. Implements option 1 as you specified it: constructor value is the default, a per-call value overrides it for that invocation only, and it applies to `run`, `stream` and `streamEvents`.

## Why

`modelCallLimitMiddleware` takes its budget when the executor is built (`mcp_agent_langchain.ts:563`), so a per-call `maxSteps` never reached it.

- `run` and `stream` destructured the argument and used it in exactly two places: forwarding to a remote agent, and the telemetry field. The limit and `recursionLimit` both read `this.maxSteps`.
- `streamEvents` assigned it to `this.maxSteps`. That moved the recursion limit but not the middleware, and it left the raised ceiling in place for every later call that passed no override.

## The change

You were right that the middleware can be steered at runtime. It resolves `runtime.context.runLimit ?? middlewareOptions?.runLimit`, so passing the budget through the run context overrides the baked-in value with no executor rebuild:

```diff
+          context: { runLimit: stepBudget },
-          recursionLimit: this.maxSteps * 3,
+          recursionLimit: stepBudget * 3,
```

`stepBudget` is `steps ?? this.maxSteps`, computed once per call on both paths. The recursion limit and the `maxStepsUsed` telemetry field follow the same number, and `this.maxSteps` is no longer written to.

Measured against the mocked executor, agent constructed with `maxSteps: 5`:

```
run({ maxSteps: 20 })            context {"runLimit":20}  recursionLimit 60
streamEvents({ maxSteps: 20 })   context {"runLimit":20}  recursionLimit 60
  this.maxSteps afterwards       5
streamEvents({}) next call       context {"runLimit":5}   recursionLimit 15
```

## Tests

New `tests/unit/agent/max-steps-budget.test.ts` asserts the budget the executor actually receives on both paths, including that the third call falls back to the constructor default. On the unfixed source both cases fail with `expected { streamMode: 'updates', …(5) } to match object { context: { runLimit: 20 }, …(1) }`.

One existing test changed. `stream_events.test.ts > should respect maxSteps parameter` asserted the old behaviour directly:

```ts
// Verify that maxSteps was set on the agent instance
expect(agent["maxSteps"]).toBe(5);
```

That is the stickiness you asked to remove, so it now checks the context the executor was invoked with and that the constructor default survived.

## Validation

From `libraries/typescript/packages/agent`: `vitest run` — 147 passed. The 8 remaining failures are in `tests/integration/agent`, which fail identically on `canary` without this diff; they need live provider credentials.

---
## Summary by cubic
Fixes per-call `maxSteps` so it actually limits model calls on `run`, `stream`, and `streamEvents`. Previously `run` and `stream` dropped the limit entirely, and `streamEvents` wrote it to the instance, permanently raising the ceiling for later calls. Closes #2413.

**Bug Fixes**
- Passes the per-call budget through the run context, which `modelCallLimitMiddleware` reads before its baked-in value.
- Keeps the constructor `maxSteps` as the default for calls that pass no override.
- Makes recursion limit and `maxStepsUsed` telemetry follow the same per-call budget.
- Adds tests that exercise the `stream` options-object overload directly.

<sup>Written for commit c602df425be8ad110aa911f61fe3d000c6e71fe9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2437?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khandrew1 (changes requested):
Could we simplify these tests and cover a couple more cases?

The new file adds 251 lines for two tests, with the first 178 lines copied from the telemetry suite—including an unrelated header and unused helpers/imports. Reusing existing setup or a smaller executor fixture would make this easier to maintain.

The main coverage gaps are:

- **Lower overrides:** every case increases the budget. An implementation that ignores values below the constructor default still passes.
- **Default preservation after `run`:** only `streamEvents` checks a subsequent call without an override. Making `run`/`stream` overrides sticky still passes.

A compact parameterized test covering a lower override, a higher override, and a subsequent omitted override for both execution paths would provide stronger coverage.

Also, please rename “enforces the budget” to “passes the budget to the executor.” These tests mock both the executor and middleware, so they verify forwarding rather than enforcement. That’s a reasonable unit-test boundary; the names should reflect it.

ayaangazali:
Both cubic findings were right and are fixed in 6814f296.

I built that test file by copying the mock preamble out of `agent-telemetry.test.ts`, and two pieces of that suite came along with it:

- **The file header** described `trackAgentExecution` assertions that do not exist here. Replaced with what the file actually asserts: that the per-call budget reaches the executor config, since `modelCallLimitMiddleware` takes its `runLimit` at construction and the invocation config is the only place the effect is visible.
- **`lastExecution()`** was dead. Removed.

One part of the finding I did not apply, with the reason: `mockTrackAgentExecution` stays. The `vi.mock("@mcp-use/client")` factory returns it from `Telemetry.getInstance()`, and the agent reaches for that during a run, so removing the stub breaks construction rather than tidying it. The comment above it now says that is all it is there for, instead of implying the suite verifies telemetry.

`vitest run tests/unit/agent/max-steps-budget.test.ts` — 2 passed. eslint and prettier clean on the file.

ayaangazali:
All three points addressed, @khandrew1.

**Simplified.** 251 lines to 124, and the 178 copied lines are gone. It now uses the lighter pattern `stream_events.test.ts` already has: mock `@mcp-use/client` and the LangChain adapter, inject the executor, and mark the agent initialized so `initialize()` does not build a real one over the top of the injected mock. That removes the `langchain`, `node:fs`, `node:os` and observability mocks entirely.

**Lower overrides.** You were right that this was the real hole: every case raised the budget, so an implementation that ignored anything below the constructor default would have passed. There is now a lower override (2 against a default of 5) for both paths.

**Default preservation after `run`.** Also right. Both paths now make an override call followed by a call with no override and assert the default is back, plus that `this.maxSteps` itself is untouched. Making `run` sticky now fails.

Parameterized with `it.each` over the two paths, so it is six cases in two blocks:

```
run() passes a budget of 2 to the executor
run() passes a budget of 20 to the executor
streamEvents() passes a budget of 2 to the executor
streamEvents() passes a budget of 20 to the executor
run() leaves the constructor default in place for the next call
streamEvents() leaves the constructor default in place for the next call
```

**Renamed.** "enforces the budget" is now "passes a budget of N to the executor", and the file header says outright that the executor and the middleware are both mocked so these verify forwarding rather than enforcement. That was a fair correction; the old name claimed more than the test does.

All six fail on the unfixed source, verified by checking out canary's `mcp_agent_langchain.ts`. `tests/stream_events.test.ts` and `tests/unit` together: 30 passed. eslint and prettier clean.

ayaangazali:
Rebased onto current canary, which had drifted enough to conflict. The diff is now four files: the changeset, `mcp_agent_langchain.ts`, the rewritten `max-steps-budget.test.ts`, and the one assertion in `stream_events.test.ts` that pinned the old sticky behaviour. Nothing else, and none of the already-merged work my earlier branch had been carrying.

Local verification skipped this round on purpose: the machine is at 3.3GB swap against a 3GB ceiling after an out-of-memory incident earlier in the week, so I am not running suites until it settles. CI is the check. The content is byte-identical to what I ran green yesterday, 6 passed and all six failing on reverted source.

ayaangazali:
cubic's latest is right and is fixed in 570fcd31.

`stream` had no cases of its own. The `run` cases reach `stream` through the positional-argument path, so a regression in the public `stream({ prompt, maxSteps })` overload would have passed the suite. It is now a third path in the parameterization with the same three cases as the others: lower override, higher override, and a later call with no override.

Nine cases now, and all nine fail on the unfixed source, verified by checking out canary's `mcp_agent_langchain.ts`.

ayaangazali:
Rebased onto current canary in c602df4 and dropped the explanatory comments from the diff, so it is now the source change, the changeset, the new `max-steps-budget` cases and the one assertion in `stream_events.test.ts` that pinned the old sticky behaviour. All checks are green and there are no unresolved threads.

Everything you asked for on 2026-09-08 is in: the test file is the lighter `stream_events` shape, there are lower-override and post-`run` default cases for each of `run`, `stream` and `streamEvents`, and the names say "passes to the executor" rather than "enforces", since the executor and the middleware are both mocked.

Whenever you have a moment, this one is ready for another look. @khandrew1

khandrew1 (approved):
Thanks for the contribution!

PR #442, opened 2025-11-06
CI: approve fork PRs by promotion + call reusable CI

gabrielekarra:
Fixes #436

Summary
- Add .github/workflows/approve-fork-pr.yml: maintainer-run workflow that
  - takes a PR number
  - promotes the fork PR head commit into a temporary base-repo branch (fork/<author>/<branch>)
  - calls the existing CI (now reusable) with secrets on that exact commit
- Update .github/workflows/ci.yml to support workflow_call with inputs:
  - ref: commit SHA/ref to test
  - python / typescript / create_mcp_use_app booleans to scope jobs
  - run_changeset_check boolean (off by default when called)
- detect-changes job now works for both normal push/PR (paths-filter) and workflow_call (uses inputs) and exposes the same outputs, so downstream jobs remain unchanged.
- All checkouts are ref-aware when called via workflow_call.

Why
- Fork PRs cannot access repo secrets, so integration tests were skipped/failing.
- This PR adds an explicit maintainer approval step that safely re-runs full CI (with secrets) on the same commit, without exposing secrets to untrusted PR events.

Usage
- Actions → "Approve fork PR CI (promote + call CI)" → enter PR number → Run.
- The workflow promotes the commit and then calls CI on that SHA; results show under the called workflow run.

Notes
- No change to normal CI behavior for push/pull_request events.
- Branch protection can continue to use existing CI checks. For fork PRs, run this workflow to get authoritative, secrets-enabled results.

PR #2452, opened 2026-09-04
fix(ci): post test reports when CI is called for a fork PR

MohammedAlkindi:
`approve-fork-pr` is dispatched manually and then calls `ci.yml` as a reusable workflow. A called workflow sees the caller's `github` context, so `github.event_name` there is `workflow_dispatch`, never `pull_request`. Both python report jobs gate on that value, so neither runs on this path. `context.issue.number` is undefined off the `pull_request` trigger too, so the comment would have had no target even if they did run.

This is the same caller-context problem #2418 fixed for the checkouts. Those tested for `workflow_call`; these two compare against `pull_request`, so that sweep did not reach them. The fix passes the dispatched PR number through as an input and uses it for both the job condition and the comment target, falling back to the existing context on the `pull_request` path exactly as before.

Worth saying plainly: `approve-fork-pr` has never run, 0 runs since #442 added it, so this is latent rather than live breakage and I cannot show you a skipped job. The evidence is structural.

Disclosure: AI-assisted. I checked that both files parse, that the new input reaches both jobs, and that the two conditions read as intended.

---
## Summary by cubic
Fixes test report jobs not running when CI is triggered from an approved fork PR. Previously, the report jobs checked `github.event_name == 'pull_request'`, but that event is never seen when `ci.yml` is called as a reusable workflow from `approve-fork-pr`. Now the PR number is passed as an input and used to gate the jobs and target the comment, falling back to the existing context on the normal pull request path.

<sup>Written for commit 7ebeec18099f7550af6c20a865132cc565865804. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2452?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/.github/workflows/approve-fork-pr.yml
+++ b/.github/workflows/approve-fork-pr.yml
@@ -106,4 +106,5 @@ jobs:
       typescript: ${{ needs.promote.outputs.typescript == 'true' }}
       create_mcp_use_app: ${{ needs.promote.outputs.create_mcp_use_app == 'true' }}
       run_changeset_check: false
+      pr_number: ${{ inputs.pull_number }}
     secrets: inherit
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,6 +79,10 @@ on:
         required: false
         type: boolean
         default: false
+      pr_number:
+        description: "PR number for the reporting jobs to comment on"
+        required: false
+        type: number
 
 jobs:
   detect-changes:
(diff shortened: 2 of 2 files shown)


khandrew1:
Thanks for the contribution! Our CI has been having a lot of issues lately. I opened #2457 to refactor it so I'm going to close this out.

PR #2458, opened 2026-09-05
fix(client): coalesce concurrent connector operations and prevent process leaks

### LINKED ISSUE #2459: client: concurrent connect() / disconnect() calls leak child processes and drop in-flight teardown
## Description

In `@mcp-use/client`, connection state management was previously partitioned across `BaseConnector`, `StdioConnector`, and `HttpConnector` with only a synchronous `if (this.connected)` check. Because `this.connected` is only set to `true` after asynchronous child process spawning, transport creation, and protocol handshakes complete:

1. **Child Process Leaks in `StdioConnector`**: Concurrent calls to `connect()` (e.g. during React `StrictMode` mounting, auto-reconnect racing manual triggers, or parallel startup) both pass the synchronous `if (this.connected)` check. Each call instantiates a separate `StdioConnectionManager` and spawns a new OS child process via `child_process.spawn`. The second assignment overwrites `this.connectionManager`, leaving the first child process running in the background as an orphaned, unkillable zombie process.
2. **Transport & Client Reference Overwrites in `HttpConnector`**: Concurrent calls create parallel `StreamableHTTPClientTransport` instances. If one fails or times out, its `catch` block invokes `cleanupResources()`, tearing down resources while the parallel connection is mid-handshake.
3. **Dropped `disconnect()` During In-Flight `connect()`**: When `disconnect()` is invoked while `connect()` is in flight, it evaluates `if (!this.connected) return;`, logs `"Not connected"`, and returns immediately. Seconds later, `connect()` completes and leaves the connection permanently open, leaking processes and streams despite the caller's explicit disconnect command.
4. **Duplicate Teardowns on Concurrent `disconnect()`**: Multiple `disconnect()` callers simultaneously pass `if (!this.connected)` because `this.connected = false` occurs only after `cleanupResources()`, racing calls to `client.close()` and `connectionManager.stop()`.

## Reproduction

```ts
import { StdioConnector } from "@mcp-use/client";

const connector = new StdioConnector({
  command: "npx",
  args: ["-y", "@modelcontextprotocol/server-everything"],
});

// 1. Concurrent connect() calls:
// Triggers two parallel child_process.spawn calls.
// The second overwrites this.connectionManager; the first process is leaked.
await Promise.all([connector.connect(), connector.connect()]);

// 2. Disconnect during in-flight connect():
const freshConnector = new StdioConnector({ command: "node", args: ["server.js"] });
const connectPromise = freshConnector.connect();
await freshConnector.disconnect(); // Returns immediately because connected is false!
await connectPromise; // Completes and leaves process running in background!
```

## Root Cause

In `libraries/typescript/packages/client/src/transport/`:
- `BaseConnector.disconnect()`:
  ```ts
  if (!this.connected) {
    this.log("Not connected");
    return;
  }
  await this.cleanupResources();
  this.connected = false;
  ```
- `StdioConnector.connect()`:
  ```ts
  if (this.connected) return;
  this.connectionManager = new StdioConnectionManager(...);
  await this.connectionManager.start();
  this.connected = true;
  ```
- `HttpConnector.connect()`:
  ```ts
  if (this.connected) return;
  // Creates transport and client asynchronously before setting this.connected = true
  ```

Because there are no locks or coalescing promises (`connectPromise`, `disconnectPromise`), any asynchronous delay between start and completion exposes race windows.

## Suggested Solution

1. **Promise Coalescing in `BaseConnector`**:
   - Track `connectPromise` and `disconnectPromise`.
   - Subsequent concurrent `connect()` calls return the active `connectPromise`.
   - Subsequent concurrent `disconnect()` calls return the active `disconnectPromise`.
2. **In-Flight Synchronization**:
   - If `disconnect()` is called while `connect()` is in flight, `disconnect()` awaits completion of `connectPromise` before tearing down resources, and `connect()` detects cancellation and rejects with an explicit error (`Connection cancelled by disconnect`).
   - If `connect()` is called while `disconnect()` is in flight, `connect()` awaits the teardown before starting fresh.
3. **Template Method Pattern**:
   - Move lifecycle locking into `BaseConnector.connect()` and `BaseConnector.disconnect()`.
   - Subclasses (`StdioConnector`, `HttpConnector`, `CodeModeConnector`) implement `protected establishTransport(): Promise<void>`.

rohith500:
Fixes #2459

### Summary
In `@mcp-use/client`, connection state management was previously partitioned across `BaseConnector`, `StdioConnector`, and `HttpConnector` with only a synchronous `if (this.connected)` guard. Because `this.connected` was only set to `true` after child process spawning, transport creation, and protocol handshakes completed:

1. **Child Process Leaks in `StdioConnector`**: Concurrent calls to `connect()` (e.g. during React `StrictMode` mounting, auto-reconnect racing manual triggers, or parallel startup) both passed the `if (this.connected)` guard. Each call instantiated a separate `StdioConnectionManager` and spawned a new OS child process via `child_process.spawn`. The second assignment overwrote `this.connectionManager`, leaving the first child process running in the background as an orphaned, unkillable zombie process.
2. **Transport & Client Reference Overwrites in `HttpConnector`**: Concurrent calls created parallel `StreamableHTTPClientTransport` instances. If one failed or timed out, its `catch` block invoked `cleanupResources()`, tearing down resources while the parallel connection was mid-handshake.
3. **Dropped `disconnect()` During In-Flight `connect()`**: When `disconnect()` was invoked while `connect()` was in flight, it evaluated `if (!this.connected) return;`, logged `"Not connected"`, and returned immediately. Seconds later, `connect()` completed and left the connection permanently open, leaking processes and streams despite the caller's explicit disconnect command.
4. **Duplicate Teardowns on Concurrent `disconnect()`**: Multiple `disconnect()` callers simultaneously passed `if (!this.connected)` because `this.connected = false` occurred after `cleanupResources()`, racing calls to `client.close()` and `connectionManager.stop()`.

### Key Changes
- **Promise Coalescing in `BaseConnector`**:
  - Added internal `connectPromise` and `disconnectPromise` locks.
  - `connect()` coalesces concurrent calls onto the active `connectPromise`.
  - `disconnect()` coalesces concurrent calls onto the active `disconnectPromise`.
- **In-Flight Synchronization**:
  - If `disconnect()` is invoked while `connect()` is in flight, `disconnect()` awaits the in-flight connection before tearing down, ensuring any newly created resources are cleanly stopped and not orphaned. In this case, `connect()` rejects with `Error("Connection cancelled by disconnect")` to prevent callers from attempting operations on a disconnected transport.
  - If `connect()` is invoked while `disconnect()` is in flight, `connect()` waits for teardown to finish before opening a fresh connection.
- **Template Method Pattern**:
  - `BaseConnector` exposes `protected establishTransport(): Promise<void>`, with a no-op fallback for backwards compatibility.
  - `StdioConnector`, `HttpConnector`, and `CodeModeConnector` implement `establishTransport()` and inherit unified lifecycle synchronization.
- **Test Suite**:
  - Added `tests/unit/transport/connector-concurrency.test.ts` verifying:
    - Coalescing of concurrent `connect()` calls (ensuring `StdioConnectionManager.start()` is called only once and spawns a single process).
    - Coalescing of concurrent `HttpConnector.connect()` calls without duplicate transports.
    - Coalescing of concurrent `disconnect()` calls.
    - Synchronous teardown and cancellation rejection when `disconnect()` is called during in-flight `connect()`.
    - Sequential reconnection after disconnect.
    - Error cleanup and clean retry after connection failure.

### Testing
- `vitest run tests/unit/transport/` (all 16 tests pass)
- Full client test suite: `vitest run tests/unit/` (all 57 test files, 352 tests pass)
- TypeScript compilation and type emit: `tsup && tsc --emitDeclarationOnly --declaration` passed
- Code formatting and linting: `prettier` and `eslint` clean

### DIFF
--- a/libraries/typescript/.changeset/quiet-connectors-close.md
+++ b/libraries/typescript/.changeset/quiet-connectors-close.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Prevent duplicate subprocesses and HTTP transports from concurrent connection attempts. Coordinate connection and disconnection calls so a later disconnect cancels in-flight and queued reconnects, releasing resources before shutdown completes.
--- a/libraries/typescript/packages/client/src/code-mode/connector.ts
+++ b/libraries/typescript/packages/client/src/code-mode/connector.ts
@@ -114,12 +114,8 @@ export class CodeModeConnector extends BaseConnector {
     this._tools = this._createToolsList();
   }
 
-  async connect(): Promise<void> {
-    this.connected = true;
-  }
-
-  async disconnect(): Promise<void> {
-    this.connected = false;
+  protected override async establishTransport(): Promise<void> {
+    // No-op: wraps an existing MCPClient without separate transport setup
   }
 
   get publicIdentifier(): Record<string, string> {
--- a/libraries/typescript/packages/client/src/transport/base.ts
+++ b/libraries/typescript/packages/client/src/transport/base.ts
@@ -140,6 +140,8 @@ export abstract class BaseConnector {
   protected serverInfoCache: MCPServerInfo | null = null;
   protected authorizationCache: MCPAuthorizationInfo | undefined;
   protected connected = false;
+  private connectPromise: Promise<void> | null = null;
+  private disconnectPromise: Promise<void> | null = null;
   protected readonly opts: ConnectorInitOptions;
   protected notificationHandlers: NotificationHandler[] = [];
   protected rootsCache: Root[] = [];
@@ -466,9 +468,51 @@ export abstract class BaseConnector {
   /**
    * Establishes the transport connection and creates the SDK client.
    *
+   * Coalesces concurrent connection attempts, synchronizes with ongoing
+   * disconnections, and guarantees that orphaned processes or transports
+   * are not leaked on failure or rapid disconnect.
+   *
    * @returns A promise that resolves when the connector is connected.
    */
-  abstract connect(): Promise<void>;
+  async connect(): Promise<void> {
+    while (this.disconnectPromise) {
+      await this.disconnectPromise;
+    }
+
+    if (this.connected) {
+      logger.debug("Already connected to MCP implementation");
+      return;
+    }
+
+    if (this.connectPromise) {
+      return this.connectPromise;
+    }
+
+    const currentConnect = (async () => {
+      try {
+        await this.establishTransport();
+        if (this.disconnectPromise) {
+          return;
+        }
+        this.connected = true;
+      } catch (err) {
+        this.connected = false;
+        await this.cleanupResources();
+        throw err;
+      } finally {
+        this.connectPromise = null;
+      }
+    })();
+
+    this.connectPromise = currentConnect;
+    return this.connectPromise;
+  }
+
+  /**
+   * Subclasses override this method to perform transport-specific connection logic.
+   * Default implementation is a no-op for backwards compatibility.
+   */
+  protected async establishTransport(): Promise<void> {}
 
   /**
    * Returns transport-specific fields suitable for logs and telemetry.
@@ -506,18 +550,44 @@ export abstract class BaseConnector {
   /**
    * Disconnects the SDK client and releases transport resources.
    *
+   * Coalesces concurrent calls and coordinates with in-flight connections
+   * to ensure no child processes or transports are orphaned.
+   *
    * @returns A promise that resolves after cleanup completes.
    */
   async disconnect(): Promise<void> {
-    if (!this.connected) {
+    if (this.disconnectPromise) {
+      return this.disconnectPromise;
+    }
+
+    if (!this.connected && !this.connectPromise) {
       logger.debug("Not connected to MCP implementation");
       return;
     }
 
-    logger.debug("Disconnecting from MCP implementation");
-    await this.cleanupResources();
-    this.connected = false;
-    logger.debug("Disconnected from MCP implementation");
+    const currentDisconnect = (async () => {
+      try {
+        // If a connection attempt is in flight, await it first so that partial
+        // or just-created resources are settled and safely cleaned up.
+        if (this.connectPromise) {
+          try {
+            await this.connectPromise;
+          } catch {
+            // Failure in connect() already invokes cleanupResources()
+          }
+        }
+
+        logger.debug("Disconnecting from MCP implementation");
+        await this.cleanupResources();
+        this.connected = false;
+        logger.debug("Disconnected from MCP implementation");
+      } finally {
+        this.disconnectPromise = null;
+      }
+    })();
+
+    this.disconnectPromise = currentDisconnect;
+    return this.disconnectPromise;
   }
 
   /** Whether an SDK client currently exists for this connector. */
--- a/libraries/typescript/packages/client/src/transport/base.ts
+++ b/libraries/typescript/packages/client/src/transport/base.ts
@@ -492,7 +492,7 @@ export abstract class BaseConnector {
       try {
         await this.establishTransport();
         if (this.disconnectPromise) {
-          return;
+          throw new Error("Connection cancelled by disconnect");
         }
         this.connected = true;
       } catch (err) {
@@ -566,6 +566,7 @@ export abstract class BaseConnector {
     }
 
     const currentDisconnect = (async () => {
+      let connectFailed = false;
       try {
         // If a connection attempt is in flight, await it first so that partial
         // or just-created resources are settled and safely cleaned up.
@@ -573,12 +574,14 @@ export abstract class BaseConnector {
           try {
             await this.connectPromise;
           } catch {
-            // Failure in connect() already invokes cleanupResources()
+            connectFailed = true;
           }
         }
 
-        logger.debug("Disconnecting from MCP implementation");
-        await this.cleanupResources();
+        if (!connectFailed) {
+          logger.debug("Disconnecting from MCP implementation");
+          await this.cleanupResources();
+        }
         this.connected = false;
         logger.debug("Disconnected from MCP implementation");
       } finally {
--- a/libraries/typescript/packages/client/src/transport/base.ts
+++ b/libraries/typescript/packages/client/src/transport/base.ts
@@ -142,6 +142,7 @@ export abstract class BaseConnector {
   protected connected = false;
   private connectPromise: Promise<void> | null = null;
   private disconnectPromise: Promise<void> | null = null;
+  private disconnectGeneration = 0;
   protected readonly opts: ConnectorInitOptions;
   protected notificationHandlers: NotificationHandler[] = [];
   protected rootsCache: Root[] = [];
@@ -475,8 +476,15 @@ export abstract class BaseConnector {
    * @returns A promise that resolves when the connector is connected.
    */
   async connect(): Promise<void> {
+    const generation = this.disconnectGeneration;
     while (this.disconnectPromise) {
       await this.disconnectPromise;
+    }
+
+    // A later disconnect also cancels attempts queued behind an earlier
+    // teardown. No resources belong to this attempt yet, so do not clean up.
+    if (generation !== this.disconnectGeneration) {
+      throw new Error("Connection cancelled by disconnect");
     }
 
     if (this.connected) {
@@ -491,7 +499,7 @@ export abstract class BaseConnector {
     const currentConnect = (async () => {
       try {
         await this.establishTransport();
-        if (this.disconnectPromise) {
+        if (generation !== this.disconnectGeneration) {
           throw new Error("Connection cancelled by disconnect");
         }
         this.connected = true;
@@ -556,6 +564,9 @@ export abstract class BaseConnector {
    * @returns A promise that resolves after cleanup completes.
    */
   async disconnect(): Promise<void> {
+    // Advance even when joining existing cleanup: callers may have queued
+    // a reconnect since that cleanup began.
+    this.disconnectGeneration++;
     if (this.disconnectPromise) {
       return this.disconnectPromise;
     }
--- a/libraries/typescript/packages/client/src/transport/http.ts
+++ b/libraries/typescript/packages/client/src/transport/http.ts
@@ -531,12 +531,7 @@ export class HttpConnector extends BaseConnector {
    * @returns A promise that resolves after protocol negotiation completes.
    * @throws An error with `code: 401` when authentication is required.
    */
-  async connect(): Promise<void> {
-    if (this.connected) {
-      logger.debug("Already connected to MCP implementation");
-      return;
-    }
-
+  protected override async establishTransport(): Promise<void> {
     const baseUrl = this.baseUrl;
     logger.debug(`Connecting to MCP implementation via HTTP: ${baseUrl}`);
 
@@ -558,8 +553,6 @@ export class HttpConnector extends BaseConnector {
       logger.debug("Streamable HTTP connect failed", err);
       const { fallbackReason, is401Error, httpStatusCode } =
         this.classifyStreamableHttpFailure(err);
-
-      await this.cleanupResources();
 
       if (is401Error) {
         logger.info("Authentication required");
@@ -802,7 +795,6 @@ export class HttpConnector extends BaseConnector {
         },
       } as any;
 
-      this.connected = true;
       this.transportType = "streamable-http";
       // Inbound request handlers (roots/sampling/elicitation) were registered before connect()
       logger.debug(
--- a/libraries/typescript/packages/client/src/transport/stdio.ts
+++ b/libraries/typescript/packages/client/src/transport/stdio.ts
@@ -101,12 +101,7 @@ export class StdioConnector extends BaseConnector {
    *
    * @returns A promise that resolves after protocol negotiation completes.
    */
-  async connect(): Promise<void> {
-    if (this.connected) {
-      logger.debug("Already connected to MCP implementation");
-      return;
-    }
-
+  protected override async establishTransport(): Promise<void> {
     logger.debug(`Connecting to MCP implementation via stdio: ${this.command}`);
     try {
       // 1. Build server parameters for the transport
@@ -192,7 +187,6 @@ export class StdioConnector extends BaseConnector {
       await this.client.connect(transport);
       this.setupRoundProgressForwarding();
 
-      this.connected = true;
       this.setupNotificationHandler();
       // Inbound request handlers (roots/sampling/elicitation) were registered before connect()
       logger.debug(
@@ -207,7 +201,6 @@ export class StdioConnector extends BaseConnector {
       });
     } catch (err) {
       logger.error(`Failed to connect to MCP implementation: ${err}`);
-      await this.cleanupResources();
       throw err;
     }
   }
--- a/libraries/typescript/packages/client/tests/fixtures/lifecycle-stdio-server.mjs
+++ b/libraries/typescript/packages/client/tests/fixtures/lifecycle-stdio-server.mjs
@@ -1,0 +1,25 @@
+import { appendFileSync } from "node:fs";
+import { createInterface } from "node:readline";
+
+// Record every real spawn, including children a broken connector loses track of.
+appendFileSync(process.argv[2], `${process.pid}\n`);
+// Safety net if the test worker itself crashes before its cleanup hook runs.
+setTimeout(() => process.exit(1), 10_000).unref();
+
+const input = createInterface({ input: process.stdin });
+input.on("close", () => process.exit(0));
+input.on("line", (line) => {
+  const request = JSON.parse(line);
+  if (request.method !== "initialize") return;
+  process.stdout.write(
+    JSON.stringify({
+      jsonrpc: "2.0",
+      id: request.id,
+      result: {
+        protocolVersion: request.params.protocolVersion,
+        capabilities: {},
+        serverInfo: { name: "lifecycle-fixture", version: "1.0.0" },
+      },
+    }) + "\n"
+  );
+});
--- a/libraries/typescript/packages/client/tests/helpers/deferred.ts
+++ b/libraries/typescript/packages/client/tests/helpers/deferred.ts
@@ -1,0 +1,10 @@
+/** A gate tests can release explicitly instead of relying on elapsed time. */
+export function deferred() {
+  let resolve!: () => void;
+  let reject!: (error: Error) => void;
+  const promise = new Promise<void>((done, fail) => {
+    resolve = done;
+    reject = fail;
+  });
+  return { promise, resolve, reject };
+}
--- a/libraries/typescript/packages/client/tests/unit/transport/connector-concurrency.test.ts
+++ b/libraries/typescript/packages/client/tests/unit/transport/connector-concurrency.test.ts
@@ -173,7 +173,10 @@ describe("Connector lifecycle concurrency", () => {
       // Trigger disconnect while connect is in flight
       const disconnectPromise = connector.disconnect();
 
-      await Promise.all([connectPromise, disconnectPromise]);
+      await expect(connectPromise).rejects.toThrow(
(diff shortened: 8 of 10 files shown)


rohith500:
Addressed in `e119a28`: Updated `BaseConnector.connect()` to reject with `Error("Connection cancelled by disconnect")` when superseded by an in-flight `disconnect()`, preventing callers from attempting subsequent operations in a disconnected state. Updated `disconnect()` to gracefully handle in-flight cancellation teardowns and verified all unit tests and concurrency scenarios pass.

Elioooon:
I ran a real-subprocess follow-up against `e119a284` and its current main base `08bbc7a1`, through `MCPConnection → StdioConnector → StdioConnectionManager → SDK Client/StdioClientTransport`. The fixture speaks the legacy initialize/tools protocol over actual stdin/stdout; handshake release and teardown are explicitly gated rather than scheduled with sleeps.

Results on this head:

- Three concurrent `connect()` calls create **one actual child**, complete initialization and `tools/list`, and concurrent disconnects leave that PID gone. On main, the same one-process assertion fails with **three transports/processes**.
- Disconnect during the gated initialize rejects all three connect waiters with `Connection cancelled by disconnect`, then the child exits. On main, those connect promises instead fulfill.
- Reconnect waits behind a gated normal transport close; the old PID is gone when that disconnect completes.
- An injected initialize error allows a fresh retry on the same `MCPConnection`. There is a narrower, **pre-existing SDK shutdown boundary** here: the failure can settle and a new child can start while the previous child is still exiting. The same characterization passes on main and on this PR.

The last point comes from `@modelcontextprotocol/client@2.0.0`: `Client._legacyHandshake()` calls `this.close()` without awaiting it, and `StdioClientTransport.close()` clears `_process` before waiting for child exit. The subsequent close calls in connector/manager cleanup can therefore return without joining that first shutdown. I reproduced it with the SDK alone below. This establishes transient overlap during failed-handshake cleanup; it does **not** establish a permanent orphan. The held child exits when the probe releases its shutdown gate.

A useful follow-up boundary is whether “clean retry” promises predecessor-process exit. If that is required, joining the original SDK/transport close is necessary; adding another `await client.close()` alone does not establish it. I have kept the characterization separate from the two regressions this PR fixes.

Validation: **21/21 transport tests passed** (16 existing + 5 subprocess scenarios), client source typecheck, probe typecheck, ESLint and Prettier. Linux, Node 22.22.2, SDK client/core 2.0.0, Vitest 4.1.0. This adds real-process coverage for legacy stdio; it does not establish real HTTP/OAuth, modern negotiation, React lifecycle, Windows, or performance coverage. The normal-close gate wraps the real transport close; spawning, JSON-RPC, pipe I/O and process exit are real.

<details>
<summary>Standalone SDK-only failure-close reproduction (Linux)</summary>

In a temporary directory with Node 22, install `@modelcontextprotocol/client@2.0.0`, save this as `repro.mjs`, and run `node repro.mjs`. It signals only the child it just created and awaits cleanup.

```sh
npm install @modelcontextprotocol/client@2.0.0
node repro.mjs
```

```js
import assert from "node:assert/strict";
import { Client } from "@modelcontextprotocol/client";
import { StdioClientTransport } from "@modelcontextprotocol/client/stdio";

// Keep the failed server alive at EOF until the parent releases it explicitly.
const child = `
  const lines = require('node:readline').createInterface({ input: process.stdin });
  let held;
  let released = false;
  process.on('SIGUSR2', () => { released = true; clearInterval(held); });
  lines.on('close', () => { if (!released) held = setInterval(() => {}, 1000); });
  lines.on('line', line => {
    const request = JSON.parse(line);
    if (request.method === 'initialize') process.stdout.write(JSON.stringify({
      jsonrpc: '2.0', id: request.id,
      error: { code: -32603, message: 'injected initialization failure' }
    }) + '\\n');
  });
`;
const transport = new StdioClientTransport({
  command: process.execPath,
  args: ["-e", child],
});
let pid;
let closing;
let closeFinished = false;
const start = transport.start.bind(transport);
transport.start = async () => {
  await start();
  pid = transport.pid;
};
const close = transport.close.bind(transport);
transport.close = () => {
  const attempt = close();
  if (!closing) {
    closing = attempt;
    void attempt.then(() => {
      closeFinished = true;
    });
  }
  return attempt;
};
const client = new Client(
  { name: "close-boundary-repro", version: "1.0.0" },
  { versionNegotiation: { mode: "legacy" } },
);
try {
  await assert.rejects(
    client.connect(transport),
    /injected initialization failure/,
  );
  // Match the cleanup's second call after SDK _legacyHandshake starts closing.
  await client.close();
  assert.equal(typeof pid, "number");
  assert.equal(closeFinished, false);
  assert.equal(transport.pid, null);
  process.kill(pid, 0); // The original child is still alive.
  console.log(
    "connect rejected; second close returned; first close pending; child alive",
  );
} finally {
  if (pid) {
    try {
      process.kill(pid, "SIGUSR2");
    } catch {
      /* Already exited. */
    }
  }
  await closing;
  await client.close();
}
assert.throws(() => process.kill(pid, 0), { code: "ESRCH" });
console.log("signal released shutdown; original close completed; child exited");
```

Expected output:

```text
connect rejected; second close returned; first close pending; child alive
signal released shutdown; original close completed; child exited
```

</details>


---

**HTTP follow-up on the same head (`e119a284`, SDK client/core `2.0.0`)**

I also exercised `MCPConnection → HttpConnector → SDK Client/StreamableHTTPClientTransport` against an actual loopback HTTP server, with native fetch and explicit initialize/DELETE gates. No transport methods are mocked.

| Scenario | This PR | main `08bbc7a1` |
| --- | --- | --- |
| Three concurrent connects, then tools/list and concurrent disconnects | One session, one DELETE, no session remaining | Creates three sessions; single-session assertion fails |
| Disconnect while initialize is held | All three connect waiters reject; session deleted | All three connect waiters fulfill; cancellation assertion fails |
| Reconnect while DELETE is held | New initialize occurs after old DELETE completes; new session usable | Reconnect is a no-op; expected second initialize is absent |
| Initialize assigns a session, then notifications/initialized returns HTTP 500 | Connect rejects, local client cleared, old session receives no DELETE; retry uses/deletes a new session | Same characterization passes |

The fourth row is a **pre-existing failure boundary**, separate from the three concurrency regressions. `HttpConnector.connectWithStreamableHttp` only stores `streamableTransport` after the handshake succeeds. In addition, the SDK's `_legacyHandshake` catch calls `close()`, which aborts the transport controller. The probe confirms that even retaining the failed transport and explicitly calling `terminateSession()` afterwards rejects with `AbortError` and sends no DELETE. Moving the assignment earlier alone is therefore insufficient.

If best-effort server-session termination on failed initialization is part of the intended contract, it needs a teardown path that remains usable after the SDK abort, or SDK-supported termination before close. A separate DELETE path would need to preserve authentication, gateway routing, a bounded cleanup deadline, and the original connection error. Alternatively, retaining current behavior relies on server-side expiry and should bound the cleanup guarantee accordingly. This probe demonstrates missing DELETE, not permanent production retention or leaked TCP sockets.

Validation: all **25 transport tests passed** (16 existing + 5 stdio probes above + 4 HTTP probes below); client `tsc --noEmit`, strict probe typecheck, ESLint and Prettier passed. The identical HTTP file on main gives **3 expected failures + 1 characterization pass**. Scope: Linux/Node 22.22.2, legacy `2025-11-25` JSON responses; optional GET returns 405. This does not cover SSE streaming/reconnection, modern negotiation, OAuth, gateways, browser runtime, or production server expiry.

<details>
<summary>Complete HTTP probe: save as packages/client/tests/unit/transport/connector-http-lifecycle.test.ts</summary>

From `libraries/typescript/packages/client`, with the repository dependencies installed:

```bash
pnpm exec vitest run tests/unit/transport/connector-http-lifecycle.test.ts --testTimeout=10000
```

The final test intentionally characterizes existing behavior; it is not a desired permanent no-DELETE contract.

```typescript
import {
  createServer,
  type IncomingMessage,
  type ServerResponse,
} from "node:http";
import { once } from "node:events";
import { StreamableHTTPClientTransport } from "@modelcontextprotocol/client";
import { afterEach, describe, expect, it } from "vitest";
import { MCPConnection } from "../../../src/core/session.js";
import { HttpConnector } from "../../../src/transport/http.js";

function deferred() {
  let resolve!: () => void;
  const promise = new Promise<void>((done) => {
    resolve = done;
  });
  return { promise, resolve };
}

const cleanups: Array<() => Promise<void>> = [];
afterEach(async () => {
  await Promise.all(cleanups.splice(0).map((cleanup) => cleanup()));
});

async function fixture(
  options: {
    holdInitialize?: boolean;
    holdDelete?: boolean;
    failInitialized?: boolean;
  } = {}
) {
  const initializeSeen = deferred();
  const initializeGate = deferred();
  const deleteSeen = deferred();
  const deleteGate = deferred();
  if (!options.holdInitialize) initializeGate.resolve();
  if (!options.holdDelete) deleteGate.resolve();
  const activeSessions = new Set<string>();
  const transports: StreamableHTTPClientTransport[] = [];
  const events: string[] = [];
  const errors: unknown[] = [];
  let sequence = 0;
  let failInitialized = options.failInitialized ?? false;

  async function handle(req: IncomingMessage, res: ServerResponse) {
    const session = String(req.headers["mcp-session-id"] ?? "");
    if (req.method === "DELETE") {
      events.push(`delete:start:${session}`);
      deleteSeen.resolve();
      await deleteGate.promise;
      activeSessions.delete(session);
      events.push(`delete:end:${session}`);
      res.writeHead(200).end();
      return;
    }
    // A standalone SSE stream is optional; these probes exercise JSON responses.
    if (req.method !== "POST") {
      res.writeHead(405).end();
      return;
    }
    let body = "";
    for await (const chunk of req) body += String(chunk);
    const message = JSON.parse(body) as { id?: number; method: string };
    events.push(`${message.method}:${session}`);
    if (message.method === "initialize") {
      const id = `session-${++sequence}`;
      activeSessions.add(id);
      initializeSeen.resolve();
      await initializeGate.promise;
      res.writeHead(200, {
        "content-type": "application/json",
        "mcp-session-id": id,
      });
      res.end(
        JSON.stringify({
          jsonrpc: "2.0",
          id: message.id,
          result: {
            protocolVersion: "2025-11-25",
            capabilities: { tools: {} },
            serverInfo: { name: "lifecycle-probe", version: "1" },
          },
        })
      );
    } else if (message.method === "notifications/initialized") {
      res.writeHead(failInitialized ? 500 : 202).end();
    } else if (message.method === "tools/list") {
      res.writeHead(200, { "content-type": "application/json" });
      res.end(
        JSON.stringify({
          jsonrpc: "2.0",
          id: message.id,
          result: { tools: [] },
        })
      );
    } else {
      res.writeHead(202).end();
    }
  }
  const server = createServer((req, res) => {
    void handle(req, res).catch((error: unknown) => {
      errors.push(error);
      if (!res.headersSent) res.writeHead(500);
      res.end();
    });
  });
  server.listen(0, "127.0.0.1");
  await once(server, "listening");
  const address = server.address();
  if (!address || typeof address === "string")
    throw new Error("Missing server address");
  const connector = new HttpConnector(`http://127.0.0.1:${address.port}/mcp`, {
    protocolNegotiation: "legacy",
    detectMixedAuth: false,
    timeout: 2000,
    // Explicitly use native fetch, independent of external proxy configuration.
    fetch: globalThis.fetch,
    wrapTransport: (transport) => {
      if (!(transport instanceof StreamableHTTPClientTransport)) {
        throw new Error("Expected the real HTTP transport");
      }
      transports.push(transport);
      return transport;
    },
  });
  const connection = new MCPConnection(connector, false);
  cleanups.push(async () => {
    initializeGate.resolve();
    deleteGate.resolve();
    await connection.disconnect();
    server.closeAllConnections();
    await new Promise<void>((resolve, reject) =>
      server.close((error) => (error ? reject(error) : resolve()))
    );
    expect(errors).toEqual([]);
  });
  return {
    connection,
    activeSessions,
    transports,
    events,
    initializeSeen,
    initializeGate,
    deleteSeen,
    deleteGate,
    stopFailing: () => {
      failInitialized = false;
    },
    get initializes() {
      return sequence;
    },
  };
}

describe("real HTTP connector lifecycle", () => {
  it("coalesces concurrent connects into one usable session and one DELETE", async () => {
    const f = await fixture({ holdInitialize: true });
    const connects = Promise.allSettled([
      f.connection.connect(),
      f.connection.connect(),
      f.connection.connect(),
    ]);
    await f.initializeSeen.promise;
    f.initializeGate.resolve();
    expect((await connects).map((r) => r.status)).toEqual([
      "fulfilled",
      "fulfilled",
      "fulfilled",
    ]);
    expect(f.initializes).toBe(1);
    await f.connection.initialize();
    expect(await f.connection.listTools()).toEqual([]);
    await Promise.all([
      f.connection.disconnect(),
      f.connection.disconnect(),
      f.connection.disconnect(),
    ]);
    expect(f.activeSessions.size).toBe(0);
    expect(f.events.filter((e) => e.startsWith("delete:start:"))).toEqual([
      "delete:start:session-1",
    ]);
    expect(f.connection.isConnected).toBe(false);
  });

  it("rejects concurrent connect waiters and deletes the session when disconnect races initialize", async () => {
    const f = await fixture({ holdInitialize: true });
    const connects = Promise.allSettled([
      f.connection.connect(),
      f.connection.connect(),
      f.connection.connect(),
    ]);
    await f.initializeSeen.promise;
    const disconnect = f.connection.disconnect();
    f.initializeGate.resolve();
    const outcomes = await connects;
    await disconnect;
    expect(outcomes.map((r) => r.status)).toEqual([
      "rejected",
      "rejected",
      "rejected",
    ]);
    for (const result of outcomes) {
      if (result.status === "rejected")
        expect(String(result.reason)).toContain(
          "Connection cancelled by disconnect"
        );
    }
    expect(f.initializes).toBe(1);
    expect(f.activeSessions.size).toBe(0);
    expect(f.connection.isConnected).toBe(false);
  });

  it("waits for the old session DELETE before reconnecting", async () => {
    const f = await fixture({ holdDelete: true });
    await f.connection.connect();
    const disconnect = f.connection.disconnect();
    await f.deleteSeen.promise;
    const reconnect = f.connection.connect();
    f.deleteGate.resolve();
    await Promise.all([disconnect, reconnect]);
    expect(f.initializes).toBe(2);
    const starts = f.events.flatMap((e, i) => (e === "initialize:" ? [i] : []));
    expect(starts[1]).toBeGreaterThan(f.events.indexOf("delete:end:session-1"));
    expect([...f.activeSessions]).toEqual(["session-2"]);
    await f.connection.initialize();
    expect(await f.connection.listTools()).toEqual([]);
    await f.connection.disconnect();
    expect(f.activeSessions.size).toBe(0);
  });

  it("characterizes a session left on the server after initialized notification failure", async () => {
    const f = await fixture({ failInitialized: true });
    await expect(f.connection.connect()).rejects.toThrow();
    expect(f.connection.isConnected).toBe(false);
    expect(f.events).toContain("notifications/initialized:session-1");
    expect(f.events.filter((e) => e.startsWith("delete:"))).toEqual([]);
    expect([...f.activeSessions]).toEqual(["session-1"]);
    // The SDK already aborted this transport during its handshake catch.
    // Merely retaining it and calling terminateSession later cannot send DELETE.
    expect(f.transports[0].sessionId).toBe("session-1");
    await expect(f.transports[0].terminateSession()).rejects.toMatchObject({
      name: "AbortError",
    });
    expect(f.events.filter((e) => e.startsWith("delete:"))).toEqual([]);
    f.stopFailing();
    await f.connection.connect();
    await f.connection.initialize();
    expect(await f.connection.listTools()).toEqual([]);
    await f.connection.disconnect();
    expect([...f.activeSessions]).toEqual(["session-1"]);
    expect(f.events).toContain("delete:end:session-2");
  });
});
```

</details>

rohith500:
@Elioooon This is exceptional analysis. Thank you for building an empirical Linux subprocess verification harness and sharing the standalone reproduction script. Testing concurrency against real OS pipes (`child_process`, stdin/stdout, and signals) with deterministic protocol gating is far more authoritative than synthetic mock timers, which often mask OS-level timing quirks and file descriptor drain semantics.

### 1. Verification of Core Guarantees
It is great to see the real-subprocess characterization empirically confirm the three primary fixes against `main`:
- **Process Deduplication**: 3 concurrent `connect()` invocations coalescing onto a single child process and cleanly reaping the PID upon disconnect (whereas `main` leaked 3 parallel processes and left orphaned handles).
- **Cancellation Barrier**: Calling `disconnect()` while `connect()` is mid-handshake now deterministically rejects all callers with `Connection cancelled by disconnect` and triggers process teardown (whereas `main` silently ignored `disconnect()` due to the premature `if (!this.connected) return` guard and allowed the child to stay permanently alive).
- **Teardown Serialization**: Reconnection deterministically queuing behind in-flight transport teardown.

---

### 2. Deep Dive: The Upstream `@modelcontextprotocol/client` Teardown Boundary
Your characterization of `@modelcontextprotocol/client@2.0.0` is spot on and highlights a subtle lifecycle nuance in the official SDK:

```text
client.connect(transport) [fails in _legacyHandshake]
  ├──> client calls this.close() (un-awaited!)
  └──> transport.close() synchronously clears this._process = null
        └──> Child process 'exit' event has NOT fired yet when close() resolves!
```

As your `repro.mjs` clearly demonstrates, because `StdioClientTransport.close()` nullifies `this._process` immediately rather than returning a promise that joins on the child's `'exit'` / `'close'` event, any caller awaiting `client.close()` or `transport.close()` resolves **before** the OS actually reaps the process.

#### Production Implications of This Boundary:
In most stdio servers, this transient overlap is benign since the child process receives EOF on its closed stdin and terminates shortly thereafter. However, there are two production scenarios where this transient overlap can bite:
1. **Exclusive Locks**: If an MCP server acquires an exclusive lock on startup (e.g., an embedded SQLite/DuckDB lock file, leveldb, or local state directory), an immediate retry can fail with `EBUSY` / `EWOULDBLOCK` because the dying predecessor hasn't released the OS file lock.
2. **Fixed IPC / Port Bindings**: If the stdio process binds an internal loopback socket or local IPC pipe for sidecar communication, a rapid retry can collide with `EADDRINUSE`.

---

### 3. Path Forward & Architecture Recommendations
I completely agree with your separation: keeping this PR strictly focused on the **concurrency deduplication and cancellation state machine** (fixing #2459) ensures this fix remains clean, targeted, and safe to merge.

For the predecessor process exit guarantee on failed handshake retry, there are two natural avenues:
1. **Upstream Contribution (`@modelcontextprotocol/client`)**: 
   Submit an issue/PR to `@modelcontextprotocol/client` ensuring `StdioClientTransport.close()` returns a promise that resolves only after the spawned child's `'exit'` event fires (with a timeout fallback to `SIGKILL`). That solves it at the root for all MCP SDK consumers.
2. **Framework-Level Guard (`mcp-use` StdioConnectionManager)**:
   In `StdioConnectionManager`, track the spawned `ChildProcess` reference directly. In `stop()` / `cleanupResources()`, after calling `await transport.close()`, add an explicit join barrier:
   ```ts
   if (this.childProcess && !this.childProcess.killed && this.childProcess.exitCode === null) {
     await Promise.race([
       once(this.childProcess, 'exit'),
       new Promise((resolve) => setTimeout(resolve, 2000)), // Grace period before SIGKILL
     ]);
   }
   ```

I’m opening a dedicated tracking issue for this next so we can follow up with clean predecessor-process exit joins.

Thanks again for the rigorous testing and verification—having 21/21 transport scenarios green across real subprocesses is a huge milestone for client stability!

khandrew1 (approved):
Thanks for fixing this—the shared lifecycle handling addresses a real concurrency problem.

I pushed a follow-up fix for one remaining race: a reconnect queued during shutdown could still open a connection after a later disconnect. Each disconnect now invalidates older connection attempts, while allowing explicitly requested reconnects after it.

I also reworked the tests to use explicit synchronization gates, verify cleanup after resources have been allocated, and exercise real stdio subprocesses and HTTP sessions. All 357 client tests pass locally, and the new tests catch the original race when the fix is removed.

Approved from my side, pending green CI on the updated commit.

PR #2463, opened 2026-09-06
fix(agent): forward RunOptions.messages in LangChain agent

### LINKED ISSUE #2462: agent: RunOptions.messages is silently dropped by the LangChain agent
## What happens

`RunOptions.messages` is documented as "Provider-neutral messages appended before `prompt`" and is honoured by the native agent (`mcp_agent.ts:380`):

```ts
if (options.messages?.length) {
  messages.push(...options.messages);
}
```

The LangChain agent never reads it. `normalizeRunOptions` (`mcp_agent_langchain.ts:63`) maps six of the seven `RunOptions` fields and leaves `messages` out, so nothing downstream can see it.

Driving it against the mock harness in `tests/unit/telemetry/agent-telemetry.test.ts` and inspecting what the executor is invoked with:

```ts
await agent.run({
  prompt: "what is the weather",
  messages: [{ role: "user", content: "CANARY-PROVIDER-MESSAGE" }],
});
```

```
FORWARDED: false
INPUTS: {"messages":[{"content":"what is the weather"},{"type":"ai","content":"Test response","tool_calls":[]}]}
```

The prompt is there; the provided message is gone. No warning, no error. The remote agent does not forward it either.

## Why I think this is unintended rather than a native-only option

`RunOptions` is shared by all three implementations and marks per-implementation limits explicitly. `schema` says "Structured output is supported by the LangChain entry point and remote agents. The native local agent rejects this option." `messages` carries no such caveat, and the `externalHistory` doc describes the ordering in terms of it: "appends it after memory-enabled stored conversation and before `messages` and the current `prompt`."

So the documented contract reads as general, and one of the three implementations quietly ignores it.

## The part I do not want to decide alone

There is no provider-to-LangChain converter today. `messageFormat.ts` exports `convertExternalHistoryToProvider` for the other direction only, so honouring this properly means writing the inverse, including tool-call messages and multimodal content parts. That is a real chunk of new code with mapping decisions in it, which is why I am asking rather than pushing.

1. **Implement it.** Add the inverse converter and insert the messages in the native agent's documented order: stored memory, `externalHistory`, `messages`, `prompt`. Most useful, largest diff, and the tool-call and content-part mappings want your eye.
2. **Reject it on the paths that cannot honour it**, the way the native agent already rejects `schema`, and note the limit in the `RunOptions.messages` doc. Small and honest, but it removes a documented capability instead of adding it.

Happy to send either. Reproduced on `canary` at `37dc3c8a`.

Christian-Sidak:
## Summary

- `normalizeRunOptions` in `mcp_agent_langchain.ts` omitted the `messages` field from its return value, so `RunOptions.messages` was silently dropped for all three entry points (`run`, `stream`, `streamEvents`)
- Add `messages` propagation through `normalizeRunOptions` and both streaming methods
- Add `convertProviderMessagesToLangChain` helper to convert provider-neutral `ProviderMessage[]` to LangChain `BaseMessage[]` (handles user, assistant with optional tool calls, tool, and system roles)
- Delegate from `run()` to `stream()` via the options-object overload so the `messages` field is not lost when bridging the positional-argument call
- Converted messages are inserted between `externalHistory` and the current prompt, matching the ordering documented in `RunOptions.externalHistory`

## Test plan

- [x] New unit test: `stream()` forwards a provider-neutral canary message before the prompt when `RunOptions.messages` is set
- [x] New unit test: `stream()` without `messages` continues to work correctly
- [x] New unit test: `streamEvents()` forwards a provider-neutral canary message before the prompt when `RunOptions.messages` is set
- [x] All 143 existing unit tests pass (`pnpm test:unit`)

Fixes #2462

---
## Summary by cubic
Fixes issue #2462 where the LangChain agent silently dropped `RunOptions.messages`. The messages are now converted from provider-neutral to LangChain `BaseMessage[]` and inserted between `externalHistory` and the current prompt in both `run()` and streaming paths. Both the LangChain and native agents now throw when `messages` is set in remote mode.

**Bug Fixes**
- `normalizeRunOptions` now propagates the `messages` field, and `run()` delegates to `stream()` through the options-object overload.
- Added `convertProviderMessagesToLangChain` for user, assistant (with tool calls), tool, and system roles; tool messages missing `toolCallId` throw an explicit error.
- `RunOptions.messages` in remote mode throws in `run()` and `stream()` for both agents, `streamEvents()` now always throws in remote mode, and the field's JSDoc documents the limitation.

<sup>Written for commit a4783279e1e0a203f40e8131bf8f091cd0c817aa. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2463?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/fix-langchain-run-options-messages.md
+++ b/libraries/typescript/.changeset/fix-langchain-run-options-messages.md
@@ -1,0 +1,7 @@
+---
+"@mcp-use/agent": patch
+---
+
+Forward RunOptions.messages in LangChain agent (fixes #2462).
+
+The LangChain agent silently dropped RunOptions.messages because normalizeRunOptions did not propagate the field. The messages are now converted from provider-neutral format to LangChain BaseMessage instances and inserted between externalHistory and the current prompt in both stream() and streamEvents().
--- a/libraries/typescript/packages/agent/src/agents/mcp_agent.ts
+++ b/libraries/typescript/packages/agent/src/agents/mcp_agent.ts
@@ -478,6 +478,12 @@ export class MCPAgent {
       );
     }
     if (this.isRemote && this.remoteAgent) {
+      if (options.messages?.length) {
+        throw new Error(
+          "RunOptions.messages is not supported in remote mode. " +
+            "The remote agent protocol does not carry a pre-seeded message list."
+        );
+      }
       return this.remoteAgent.run(
         options.prompt ?? "",
         options.maxSteps,
@@ -538,6 +544,12 @@ export class MCPAgent {
       signal
     );
     if (this.isRemote && this.remoteAgent) {
+      if (options.messages?.length) {
+        throw new Error(
+          "RunOptions.messages is not supported in remote mode. " +
+            "The remote agent protocol does not carry a pre-seeded message list."
+        );
+      }
       const result = await this.remoteAgent.run(
         options.prompt ?? "",
         options.maxSteps,
--- a/libraries/typescript/packages/agent/src/agents/mcp_agent_langchain.ts
+++ b/libraries/typescript/packages/agent/src/agents/mcp_agent_langchain.ts
@@ -1330,6 +1330,12 @@ export class MCPAgent {
 
     // Delegate to remote agent if in remote mode
     if (this.isRemote && this.remoteAgent) {
+      if (extraMessages?.length) {
+        throw new Error(
+          "RunOptions.messages is not supported in remote mode. " +
+            "The remote agent protocol does not carry a pre-seeded message list."
+        );
+      }
       const result = await this.remoteAgent.run(
         query,
         steps,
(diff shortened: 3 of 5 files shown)


Christian-Sidak:
Fixed in a478327.

The cubic review was correct: the run_options.ts comment promised a throw for remote mode but only the LangChain agent enforced it. The native agent's run() and stream() remote branches silently discarded messages instead.

Added the same check to both branches in mcp_agent.ts so the behavior matches the documented contract for all callers. Updated the run_options.ts JSDoc to say 'Both the native agent and the LangChain agent throw' rather than implying LangChain only.

Christian-Sidak:
Checking in on this one. The fix is for LangChain agent callers who provide RunOptions.messages — before the patch the messages were silently dropped; after it they're forwarded as expected. The native agent remote path was also aligned to match the documented contract. Happy to answer questions or make changes.

khandrew1:
Thanks for jumping on this and for iterating on the review feedback.

I'm going to close this in favor of a smaller change, #2658 (context in #2462): documenting `messages` as native-agent only, rather than adding a converter from provider-neutral messages to LangChain messages. LangChain users already have `externalHistory` for this.

Two parts of this PR I also wouldn't want to take on their own:
- **Throwing on remote agents.** The existing behavior for unsupported options in remote mode is a warning (see `externalHistory` in `remote.ts`). Turning calls that work today into runtime errors, especially in the native agent, is a breaking change I'd avoid in a patch.
- **The converter drops image parts and tool metadata**, so it would trade one silent drop for another.

PR #2466, opened 2026-09-06
fix(cli): report whether env unset actually deleted a variable

### LINKED ISSUE #2465: cli: env unset reports "Deleted" even when the variable did not exist
## What happens
`mcp-use servers env unset <server> <key>` unconditionally prints `Deleted ${key}.` and returns exit code 0, even when no variable matching that key exists.

In `envUnset()` (servers.ts):

if (existing !== undefined) {
  await api.request(
    `/servers/${encodeURIComponent(server)}/env-variables/${encodeURIComponent(existing.id)}`,
    { method: "DELETE" }
  );
}
printResult({ deleted: key }, json, `Deleted ${key}.`);

The DELETE API call is correctly skipped when `existing === undefined`, but the success message and JSON result are printed unconditionally, regardless of whether anything was actually deleted.

## Why it matters
A user or automated script has no way to tell "a real variable was deleted" from "there was nothing to delete." This is the same category of output-accuracy issue as #2388 (env set reporting a sensitivity state that did not match what was actually written).

## Reproduction
mcp-use servers env unset srv_abc123 DOES_NOT_EXIST --yes --json
# Prints: {"deleted":"DOES_NOT_EXIST"}
# No API call was made, but the output implies a real deletion happened.

## Proposed fix
Track whether a variable was actually found before attempting deletion, and reflect that in the result, e.g.:
- print `${key} does not exist.` (or similar) when `existing === undefined`
- only print `Deleted ${key}.` when a real deletion occurred
- reflect this as a boolean field in the JSON result (e.g. `deleted: true/false`) so scripts can detect it too

Happy to send a PR for this.

aishwary-dongre:
Closes #2465.

`mcp-use servers env unset` skipped the DELETE call when no matching variable
existed, but printed `Deleted <key>.` and a JSON result echoing the key either
way. A script had no way to tell a real deletion from a no-op.

## Changes

- Track whether the lookup found a variable, and reflect it in the message and
  the JSON result.
- The message names the branch when `--branch` is supplied. The lookup is
  branch-scoped, so `TOKEN does not exist.` would otherwise be false for a key
  that exists in production.
- Documented the behavior in the CLI reference.

## Breaking change to `--json` output

`deleted` changes from a string (the key) to a boolean. The key is still
present as a separate `key` field. Worth a look before merge.

## Not changed

- Exit code stays `0` when nothing was deleted — `unset` is idempotent, like
  `rm -f`. Deliberate.
- The confirmation prompt still fires before the existence check, so a user can
  be asked to confirm deleting something that isn't there. Fixing that changes
  prompting behavior beyond this issue's scope.
- `servers delete` still emits `deleted` as a string. Left alone; happy to
  align it in a follow-up if you want consistency.

## Verification

- `npx vitest run` in `packages/cli` — 295 passed, 23 files.
- `npx tsc --noEmit` clean.
- Three new tests, including the issue's exact repro asserting no DELETE
  request is issued.

---
## Summary by cubic
Closes #2465. `servers env unset` always printed `Deleted <key>.` and echoed the key in `--json` output even when no matching variable existed, and `env set` silently downgraded an existing write-only secret when `--secret` was omitted on update.

**Bug Fixes**
- `env unset` now prints `Deleted <key>.` only when a variable existed, and `<key> does not exist.` otherwise, naming the branch when `--branch` is supplied.
- `--json` output now carries a boolean `deleted` field instead of echoing the key.
- `env set` warns when a write would downgrade an existing write-only variable; `--secret` is required on every write to keep a variable write-only.
- Exit code stays `0` when nothing was deleted — `unset` is idempotent.
- `servers delete` still emits `deleted` as a string; left unchanged.

**Migration**
- `deleted` in `--json` output changes from a string to a boolean; the key is still available in the `key` field.
- Scripts that update secrets must pass `--secret` on every write to preserve write-only status.

<sup>Written for commit c066254c8e923f1e9955be5042508455d049da51. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2466?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/docs/v2/typescript/api-reference/cli-reference.mdx
+++ b/docs/v2/typescript/api-reference/cli-reference.mdx
@@ -503,7 +503,7 @@ mcp-use servers env unset srv_abc123 API_KEY --yes
 
 - `env list` returns metadata only. Values are never returned.
 - `env set` creates a production variable when `--branch` is omitted, or a preview variable when it is supplied.
-- `--secret` marks a value sensitive, but this must be passed on **every** write to the variable, including updates and rotations. Omitting `--secret` on an update stores the value as non-sensitive, even if the existing variable was previously marked sensitive. The CLI prints a warning (and sets `downgradedFromSecret: true` in `--json` output) when a write would downgrade an existing sensitive variable this way.
+- `--secret` stores a value as write-only, but this must be passed on **every** write to the variable, including updates and rotations. Omitting `--secret` on an update stores the value as non-sensitive, even if the existing variable was previously write-only. The CLI prints a short warning (`Warning: <key> is no longer write-only.`) when a write would downgrade an existing write-only variable this way; the resulting state is always reflected in the existing `secret` field of `--json` output.
 - `--branch <name>` scopes preview environment variables to a branch. Omit `--branch` for production resolution.
 - `env unset` prompts for confirmation unless `--yes` is set.
 
--- a/docs/v2/typescript/api-reference/cli-reference.mdx
+++ b/docs/v2/typescript/api-reference/cli-reference.mdx
@@ -503,7 +503,7 @@ mcp-use servers env unset srv_abc123 API_KEY --yes
 
 - `env list` returns metadata only. Values are never returned.
 - `env set` creates a production variable when `--branch` is omitted, or a preview variable when it is supplied.
-- `--secret` marks a value sensitive.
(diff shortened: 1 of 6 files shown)


aishwary-dongre:
Superseded by #2411, which fixed this in `servers env unset` (and `servers delete`) before I filed the issue — I was working from a stale fork and didn't check `upstream/main`. Their approach throws `env_variable_not_found` rather than exiting 0, which is the better call. Closing.

The one gap left was that the CLI reference didn't document the new failure behavior; that's #2467.

PR #2471, opened 2026-09-07
fix(client): guard listAllResources against pagination loops and empty cursor truncation

### LINKED ISSUE #2470: client: listAllResources lacks cycle detection, truncates empty-string cursors, and passes dirty initial params
### Summary
In `@mcp-use/client` (`packages/client/src/transport/base.ts`), `BaseConnector.listAllResources` auto-paginates through `resources/list`. However, its pagination loop contains three reliability and defensive flaws:

1. **Missing cycle detection (Denial of Service / OOM)**:
   Unlike `listAllSkills` (`session.ts:665`) which defensively maintains `seenCursors = new Set<string>()` and throws `"skills/list returned a repeated pagination cursor"`, `listAllResources` has no cycle detection. If an upstream server returns a repeated or cyclic cursor, `do ... while (cursor)` enters an infinite loop, consuming 100% CPU on the event loop and continually accumulating objects until an out-of-memory crash kills the Node.js process (in proxies or backend agent services) or freezes the browser (in `useMcp`).

2. **Premature termination on empty-string cursors (`""`)**:
   The loop terminates on `while (cursor);`. In JavaScript, `""` is falsy (`Boolean("") === false`). If an upstream server yields `nextCursor: ""` (e.g. an offset marker or zero-byte pagination token), pagination halts after page 1, silently discarding all subsequent resource pages. MCP specification pagination requires checking `while (cursor !== undefined);`.

3. **Dirty initial parameter payload (`{ cursor: undefined }`)**:
   On page 1, `let cursor: string | undefined = undefined;` is passed directly as `client.listResources({ cursor }, options)`, constructing `{ cursor: undefined }`. In strict JSON-RPC servers or transport serializers, this can serialize as `{"cursor": null}` or fail parameter schemas. It should be passed cleanly as `cursor !== undefined ? { cursor } : undefined`.

### Expected Behavior
- `listAllResources` should track visited cursors in `seenCursors = new Set<string>()` and throw `"resources/list returned a repeated pagination cursor"` upon detecting a cycle, maintaining parity with `listAllSkills`.
- The loop condition should check `while (cursor !== undefined);` so non-undefined falsy cursors like `""` are not prematurely truncated.
- The initial call should pass `undefined` instead of `{ cursor: undefined }`.

rohith500:
Fixes #2470

### Summary of Changes
1. **Defensive Cycle Detection**:
   - Added `seenCursors = new Set<string>()` inside `BaseConnector.listAllResources` loop.
   - Throws `new Error("resources/list returned a repeated pagination cursor")` if an upstream MCP server yields an already-seen cursor, matching the existing defensive pattern in `listAllSkills` (`session.ts:673`).
   - Prevents unconstrained memory growth and event-loop-blocking DoS crashes caused by circular cursor loops.

2. **Empty-String Cursor Handling**:
   - Changed loop condition from `while (cursor);` to `while (cursor !== undefined);`.
   - Ensures servers that use empty-string cursor values (such as `""` offset markers) do not have pagination prematurely halted on page 1.

3. **Clean First-Page Parameter Hygiene**:
   - Passed `cursor !== undefined ? { cursor } : undefined` to `client.listResources(...)` instead of `{ cursor }`.
   - Prevents sending `{ cursor: undefined }` to strict JSON-RPC servers on initial listing requests.

4. **Unit Tests**:
   - Added `packages/client/tests/unit/client/list-all-resources.test.ts` verifying:
     - Multi-page resource aggregation and clean initial parameters.
     - Immediate repeated cursor cycle detection.
     - Multi-hop cursor cycle detection (`A -> B -> A`).
     - Empty-string cursor continuation without premature termination.

---
## Summary by cubic
Makes `listAllResources` resilient to misbehaving pagination cursors. Previously, a repeated cursor caused an infinite loop and an empty-string cursor stopped pagination after the first page. Now repeated cursors throw an error and empty-string cursors continue correctly.

**Changes**
- Detects repeated pagination cursors and throws an error instead of looping forever.
- Treats empty-string cursors as valid continuation tokens instead of terminating pagination.
- Omits the `cursor` parameter on the first request to avoid sending `{ cursor: undefined }`.
- Adds unit tests covering multi-page aggregation, cycle detection, and empty-string cursor continuation.

<sup>Written for commit 45fd45bcc2a483d59ea8101b287fa1c34467048b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/packages/client/src/transport/base.ts
+++ b/libraries/typescript/packages/client/src/transport/base.ts
@@ -752,14 +752,26 @@ export abstract class BaseConnector {
       logger.debug("Listing all resources (with auto-pagination)");
       return await this.executeRequest(async () => {
         const allResources: any[] = [];
+        const seenCursors = new Set<string>();
         let cursor: string | undefined = undefined;
 
         do {
           const result: { resources?: any[]; nextCursor?: string } =
-            await client.listResources({ cursor }, options);
+            await client.listResources(
+              cursor !== undefined ? { cursor } : undefined,
+              options
+            );
           allResources.push(...(result.resources || []));
           cursor = result.nextCursor;
-        } while (cursor);
+          if (cursor !== undefined) {
+            if (seenCursors.has(cursor)) {
+              throw new Error(
+                "resources/list returned a repeated pagination cursor"
+              );
+            }
+            seenCursors.add(cursor);
+          }
+        } while (cursor !== undefined);
 
         return { resources: allResources };
       });
--- a/libraries/typescript/packages/client/tests/unit/client/list-all-resources.test.ts
+++ b/libraries/typescript/packages/client/tests/unit/client/list-all-resources.test.ts
@@ -1,0 +1,151 @@
+import { describe, expect, it, vi } from "vitest";
+
+import { BaseConnector } from "../../../src/transport/base.js";
+
+class TestConnector extends BaseConnector {
+  async connect(): Promise<void> {}
+
+  get publicIdentifier(): Record<string, string> {
+    return { type: "test" };
+  }
+}
+
+describe("listAllResources pagination resilience", () => {
+  it("paginates across multiple pages until nextCursor is undefined", async () => {
+    const connector = new TestConnector() as BaseConnector & {
+      client: unknown;
+      capabilitiesCache: unknown;
+    };
+    connector.capabilitiesCache = { resources: {} };
+
+    const listResourcesMock = vi
+      .fn()
+      .mockResolvedValueOnce({
+        resources: [{ uri: "res://1", name: "One" }],
+        nextCursor: "cursor-page-2",
+      })
+      .mockResolvedValueOnce({
+        resources: [{ uri: "res://2", name: "Two" }],
+        nextCursor: "cursor-page-3",
+      })
+      .mockResolvedValueOnce({
+        resources: [{ uri: "res://3", name: "Three" }],
+        nextCursor: undefined,
+      });
+
+    connector.client = { listResources: listResourcesMock };
+
+    const result = await connector.listAllResources();
+
+    expect(result).toEqual({
+      resources: [
+        { uri: "res://1", name: "One" },
+        { uri: "res://2", name: "Two" },
+        { uri: "res://3", name: "Three" },
+      ],
+    });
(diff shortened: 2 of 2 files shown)


tonxxd:
Thanks for digging into the cursor edge cases. I revalidated this against the current published stack before deciding whether we should add another paginator:

- scaffold: `create-mcp-use-app@latest`
- resolved packages: `mcp-use@2.4.3`, `@mcp-use/client@2.3.0`, and `@modelcontextprotocol/client@2.0.0`
- fixture: a two-page stdio MCP server that recorded every request

On that stack, `session.listResources()` returned both pages and sent the second cursor automatically. `session.listResources("resources-page-2")` returned only the requested page. A raw `session.request("resources/list", {})` returned the first page plus its `nextCursor`. `session.listAllResources()` also returned both pages, but its first underlying no-cursor SDK call had already drained the pagination, so the custom outer loop did not perform the pagination described here.

That is intentional in the official TypeScript SDK v2: no-cursor list calls aggregate pages, explicit-cursor calls return one page, and `Client` defaults to a 64-page safety limit. See [Calling MCP servers: let the SDK walk the pages](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/clients/calling.md#let-the-sdk-walk-the-pages) and the [v2 migration note](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/docs/migration/upgrade-to-v2.md#client-list-methods).

So this extra pagination loop would duplicate the SDK and create a second, different policy. I’m closing the PR in favor of keeping `listResources()` as the canonical all-pages API and `listAllResources()` only as the existing compatibility alias. There is a narrower upstream question about how the official SDK reports a repeated cursor—it currently stops rather than throwing—but that should be addressed in the official paginator rather than wrapped here.

PR #2482, opened 2026-09-08
fix(python): use monotonic clock for execution timing

### LINKED ISSUE #2481: fix(python): CodeExecutor reports zero execution_time on Windows
**Describe the bug**

`CodeExecutor.execute()` frequently reports an `execution_time` of exactly `0.0` for short executions on Windows.

I discovered this while setting up the repository and running the documented Python unit-test command:

```powershell
pytest tests/unit
```

The test suite returned:

```text
304 passed, 1 failed
```

The failure was:

```text
tests/unit/test_code_executor.py::TestCodeExecutorBasics::test_execute_simple_code

assert result["execution_time"] > 0
E assert 0.0 > 0
```

`CodeExecutor.execute()` currently measures elapsed time using `time.time()`. On this Windows environment, that clock has a resolution of approximately 15.625 milliseconds, so short executions often start and finish within the same clock tick.

**To Reproduce**

1. Clone the repository and enter the Python package:

```powershell
git clone https://github.com/mcp-use/mcp-use.git
cd mcp-use\libraries\python
```

2. Create a virtual environment and install the development dependencies:

```powershell
uv venv
uv pip install --python .\.venv\Scripts\python.exe -e ".[dev,anthropic,openai,search,e2b]"
```

3. Run the affected test:

```powershell
.\.venv\Scripts\python.exe -m pytest -q tests\unit\test_code_executor.py::TestCodeExecutorBasics::test_execute_simple_code
```

4. Repeat the test to expose the platform-dependent failure:

```powershell
$passed = 0
$failed = 0

1..12 | ForEach-Object {
    .\.venv\Scripts\python.exe -m pytest -q `
        tests\unit\test_code_executor.py::TestCodeExecutorBasics::test_execute_simple_code

    if ($LASTEXITCODE -eq 0) {
        $passed++
    } else {
        $failed++
    }
}

Write-Output "passed=$passed failed=$failed"
```

Observed on my machine:

```text
passed=2 failed=10
```

**Expected behavior**

`execution_time` should be measured with a monotonic, high-resolution clock and should retain useful timing information for short executions.

The result should remain a floating-point number expressed in seconds, without changing the existing return structure, timeout behavior, or error handling.

**Screenshots**

Not applicable. The failure is visible in the pytest output included above.

- OS: Windows x86_64
- Browser: Not applicable
- Version: Python 3.11.9, mcp-use 1.7.0, repository commit `08bbc7a1dcd84bbafbae7f19633ae25fa5e0774b`

- Device: Not applicable
- OS: Not applicable
- Browser: Not applicable
- Version: Not applicable

**Additional context**

The relevant implementation currently uses the wall clock:

```python
start_time = time.time()
...
execution_time = time.time() - start_time
```

Clock information on the affected machine:

```text
time:
  implementation = GetSystemTimeAsFileTime()
  monotonic = False
  adjustable = True
  resolution = 0.015625

perf_counter:
  implementation = QueryPerformanceCounter()
  monotonic = True
  adjustable = False
  resolution = 1e-07
```

I plan to replace the two `time.time()` calls with `time.perf_counter()`, which is designed for measuring short durations.

I will also make the existing test deterministic by mocking `perf_counter()` with known start and end values instead of relying on the host machine’s clock resolution.

The proposed change is limited to:

- `libraries/python/mcp_use/client/code_executor.py`
- `libraries/python/tests/unit/test_code_executor.py`

No dependencies or public APIs will change. The prepared fix passes all 305 Python unit tests and the Ruff lint and format checks. A PR will be submitted shortly.

OT-WuLong:
# Pull Request Description

## Language / Project Scope

- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Fix `CodeExecutor.execute()` reporting an `execution_time` of exactly `0.0` for short executions on Windows.

The executor previously measured elapsed time using `time.time()`. On the affected Windows environment, that clock has a resolution of approximately 15.625 milliseconds, so fast executions often started and finished within the same clock tick.

This PR replaces the wall clock with `time.perf_counter()`, a monotonic, high-resolution clock designed for measuring short durations.

## Review Readiness

- [x] I verified the affected package path with the commands listed below

## Implementation Details

1. Replace both `time.time()` calls in `CodeExecutor.execute()` with `time.perf_counter()`.
2. Update the existing unit test to mock the performance counter with deterministic start and end values.
3. Verify the exact elapsed-time calculation with `pytest.approx()`.
4. Preserve the existing return structure, timeout handling, and error behavior.
5. No dependencies or public APIs were changed.

---

## TypeScript Checklist

Not applicable. This PR only changes the Python package.

---

## Python Checklist

- [x] Code formatted with `ruff format format`
- [x] Added or updated tests
- [x] Documentation update is not required because the public API and usage remain unchanged

---

## Example Usage (Before)

On Windows, a short execution could lose all timing information:

```python
result = await executor.execute("return 1 + 1")

assert result["result"] == 2
assert result["execution_time"] == 0.0
```

This also caused the existing unit test to fail intermittently:

```text
assert result["execution_time"] > 0
E assert 0.0 > 0
```

Running the affected test 12 times on the reporting machine produced:

```text
passed=2 failed=10
```

## Example Usage (After)

Short executions are measured with a monotonic, high-resolution clock:

```python
result = await executor.execute("return 1 + 1")

assert result["result"] == 2
assert result["execution_time"] > 0
```

A 100-execution verification after the fix produced:

```text
zero_count=0
```

## Documentation Updates

No documentation changes are required. The `execution_time` field remains a floating-point duration in seconds, and no public API changed.

## Testing

Run from `libraries/python`:

```powershell
uvx --from "ruff==0.15.22" ruff check .
uvx --from "ruff==0.15.22" ruff format --check .
.\.venv\Scripts\python.exe -m pytest -q tests\unit\test_code_executor.py
.\.venv\Scripts\python.exe -m pytest -q tests\unit
```

Results:

```text
Ruff lint: passed
Ruff format: passed
CodeExecutor tests: 21 passed
Full Python unit suite: 305 passed
```

Manual verification:

```text
100 short executions
zero execution times before fix: frequently reproducible
zero execution times after fix: 0
```

## Backwards Compatibility

This change is fully backwards compatible.

The result schema, duration unit, timeout behavior, exception handling, and public API remain unchanged. The only behavioral difference is that short executions now retain accurate timing information.

## Related Issues

Closes #2481

---
## Summary by cubic
Fixes `CodeExecutor.execute()` reporting `execution_time` as `0.0` for short executions on Windows by replacing `time.time()` with `time.perf_counter()`. The old clock had ~15.625ms resolution, so fast executions often finished within the same tick; the new monotonic clock measures short durations accurately.

- Updates the unit test to mock `perf_counter` with deterministic start/end values and assert exact elapsed time.
- Preserves the existing return structure, timeout handling, and error behavior; no public API or dependencies changed.

<sup>Written for commit fea9b2348b80775dc21f06ac5504f1907894a2e0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2482?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/python/mcp_use/client/code_executor.py
+++ b/libraries/python/mcp_use/client/code_executor.py
@@ -9,8 +9,8 @@ direct tool calls.
 import asyncio
 import io
 import re
-import time
 from contextlib import redirect_stderr, redirect_stdout
+from time import perf_counter
 from typing import TYPE_CHECKING, Any
 
 from mcp_use.logging import logger
@@ -59,7 +59,7 @@ class CodeExecutor:
             logger.debug("Connecting to configured servers for code execution...")
             await self.client.create_all_sessions()
 
-        start_time = time.time()
+        start_time = perf_counter()
         logs: list[str] = []
         result = None
         error = None
@@ -93,7 +93,7 @@ class CodeExecutor:
             error = str(e)
             logger.error(f"Code execution error: {e}")
 
-        execution_time = time.time() - start_time
+        execution_time = perf_counter() - start_time
 
         # Capture any stdout/stderr that wasn't captured by our print wrapper
         if stdout_capture.getvalue():
--- a/libraries/python/tests/unit/test_code_executor.py
+++ b/libraries/python/tests/unit/test_code_executor.py
@@ -5,7 +5,7 @@ Tests the code execution functionality for MCP code mode.
 """
 
 import asyncio
-from unittest.mock import AsyncMock, MagicMock, Mock
+from unittest.mock import AsyncMock, MagicMock, Mock, patch
 
 import pytest
 
@@ -36,12 +36,13 @@ class TestCodeExecutorBasics:
         """Test executing simple Python code."""
         code = "result = 1 + 1\nreturn result"
 
-        result = await code_executor.execute(code, timeout=5.0)
+        with patch("mcp_use.client.code_executor.perf_counter", side_effect=[100.0, 100.125]):
+            result = await code_executor.execute(code, timeout=5.0)
 
         assert result["error"] is None
         assert result["result"] == 2
         assert isinstance(result["execution_time"], float)
-        assert result["execution_time"] > 0
+        assert result["execution_time"] == pytest.approx(0.125)
 
     @pytest.mark.asyncio
     async def test_execute_with_print(self, code_executor):


MohammedAlkindi (approved):
Checked that the two call sites are the only ones: `import time` is removed outright rather than narrowed, so nothing else in `code_executor.py` referenced it, and the module would not import if it did.

The mock supplies exactly two values and `execute` makes exactly two `perf_counter` calls on one linear path, so the 0.125 assertion is pinned rather than approximately right by luck.

The behaviour this fixes is real. `time.time()` can step backward on an NTP correction mid-execution, which can make `execution_time` negative, and the old `> 0` assertion was the only thing between that and a silently wrong number in the result dict. `perf_counter` cannot step.

One question rather than a blocker: is `code_executor.py` the only duration in the Python library measured off the wall clock, or are there sibling paths worth the same change while this is in hand?

---

OT-WuLong:
> Checked that the two call sites are the only ones: `import time` is removed outright rather than narrowed, so nothing else in `code_executor.py` referenced it, and the module would not import if it did.
> 
> The mock supplies exactly two values and `execute` makes exactly two `perf_counter` calls on one linear path, so the 0.125 assertion is pinned rather than approximately right by luck.
> 
> The behaviour this fixes is real. `time.time()` can step backward on an NTP correction mid-execution, which can make `execution_time` negative, and the old `> 0` assertion was the only thing between that and a silently wrong number in the result dict. `perf_counter` cannot step.
> 
> One question rather than a blocker: is `code_executor.py` the only duration in the Python library measured off the wall clock, or are there sibling paths worth the same change while this is in hand?
> 
> _Generated by [Claude Code](https://claude.ai/code)_

Thanks for the thorough review and approval

I checked the rest of the Python package. `CodeExecutor` is not the only place where wall-clock time is used for elapsed durations. There are sibling paths in the agent execution telemetry, client/server middleware metrics, sandbox startup polling, tool indexing timeouts, and server startup timing.

However, the package also uses `time.time()` for real timestamps, hourly buckets, error timestamps, and generated-at values, where replacing it with `perf_counter()` would be incorrect. Some middleware paths also derive durations from shared wall-clock timestamp fields, so addressing them safely would require separate analysis and regression coverage rather than a mechanical replacement

#2481 specifically covers `CodeExecutor.execution_time`, I kept this PR narrowly scoped to the reproduced bug and its deterministic regression test. I’d prefer to audit and address the sibling duration/timeout paths in a separate issue and PR.

khandrew1:

khandrew1 (approved):
Thanks for the contribution!

PR #2486, opened 2026-09-08
chore(agent): deprecate the auth_token alias on both MCPServerConfig types

ayaangazali:
## What this does

Marks `auth_token` deprecated on both `MCPServerConfig` types instead of removing it. v2 renamed the option to `authToken`, and nothing has read the snake-case spelling since, so setting it produces no `Authorization` header. The field stays so existing TypeScript consumers keep compiling, and editors now strike it through at the call site with a pointer to `authToken`. Removal waits for the next major.

Two declarations carry the field, so both get the tag:

- `packages/agent/src/agents/types.ts`
- `packages/inspector/src/client/components/chat/types.ts`

## Bump

`patch` for both packages. Nothing is added or removed from either type, so no consumer has to change anything to keep compiling.

## Supported-option docs

`docs/v2/typescript/` does not advertise `auth_token` as a supported option. The only v2 mention is the rename row in `client/migration.mdx`, which documents `auth_token` to `authToken` and should stay. The entry in `docs/typescript/api-reference/agent/mcp-agent.mdx` is in the legacy v1 tree, so I left it alone. Happy to take that on separately if you want the legacy page touched.

## Verification

Type-only change, no runtime code. Local build and test runs were skipped this round on purpose: the machine is at 5.4GB swap and was SIGKILLed by its own memory guardian earlier today, so I am leaving verification to CI rather than risking another thrash.


ayaangazali:
cubic is right and I bumped both to `minor`.

The runtime behaviour was already nothing, so my instinct was that nothing breaks. But the type surface does shrink: anyone who still sets `auth_token` gets a compile error on upgrade, and a patch is not supposed to do that. The agent CHANGELOG labels the analogous v2 alias removals "Breaking", which is the precedent cubic pointed at.

I went with `minor` rather than `major` because this is a v2 canary line and the field never functioned, but say the word if you would rather it were `major` and I will change it.

MohammedAlkindi (commented):
The removal itself looks clean. The diff touches only the two interface declarations, so if anything in either package still read the snake-case spelling, `tsc` would fail rather than silently compile.

One question on the version bump. The changeset marks both packages `minor` while its own text says removing the field is a type-level break for anyone still setting it. Was `minor` deliberate on the grounds that the runtime behaviour was already nothing, or should this ride a major?

A narrower one: if these configs are ever loaded from a JSON or YAML file rather than written in TypeScript, an `auth_token` key there is not type-checked at all, so those users keep the silent no-auth behaviour and get no signal from this change. Was mapping `auth_token` onto `authToken` at parse time considered instead of dropping it, so the option that already looks supported starts working?

---

ayaangazali:
Thanks, both questions are fair.

**On the bump.** `minor` was deliberate but it is not my call to make. My first instinct was `patch`, on the grounds you name: the runtime behaviour was already nothing, so nothing that worked stops working. cubic pushed back that removing an exported field is a type-level break regardless, and the agent CHANGELOG labels the analogous v2 alias removals "Breaking", so I moved it to `minor`. I said on the thread that I will change it to `major` if @khandrew1 prefers, and that still stands.

**On JSON and YAML configs.** That is the sharper half and you are right that they get no signal. A `auth_token` key in a config file is not type-checked, so those users keep the silent no-auth behaviour and this change is invisible to them.

Mapping it onto `authToken` at parse time was considered, and it is exactly what I originally proposed in #2464. khandrew1 closed that:

> Closing this because removing `auth_token` in favor of `authToken` was intentional in v2. The agent docs still listing it is an oversight; we should clean those up rather than restore support for the alias.

So making the option work is off the table by decision, not by oversight. What that decision does not cover is the silence, and I think you have identified a real residual gap: the one group still affected is the one least able to notice. If the maintainers want it, a warning when a server config carries an unknown `auth_token` key would give those users a signal without reviving the alias, and it would be a separate change from this one. Happy to open that if @khandrew1 wants it.

khandrew1 (changes requested):
Thanks for following up, and sorry my earlier comment wasn’t clear enough. We intentionally removed runtime support in v2, but leaving the field in the exported types was our oversight. Removing it now would still break existing TypeScript consumers, which I’d prefer to avoid in a minor release. Let’s retain it with `@deprecated Use authToken instead`, keep it out of the supported-option docs, and defer removal to the next major.

ayaangazali:
Understood, and thanks for spelling it out. Switched to deprecation in 65cba1e.

`auth_token` stays on both `MCPServerConfig` declarations, now carrying `@deprecated Use authToken instead`, so nobody's build breaks and editors flag it at the call site. The changeset is `patch` for both packages rather than `minor`, since the type surface no longer changes in either direction. Rebased onto current canary while I was in there.

On keeping it out of the supported-option docs: `docs/v2/typescript/` does not list it as an option today. The only v2 mention is the `auth_token` to `authToken` row in `client/migration.mdx`, which I read as the right place for it to stay. There is still an `auth_token` ResponseField in `docs/typescript/api-reference/agent/mcp-agent.mdx`, but that is the legacy v1 tree, so I left it. Say the word if you want that page updated and I will do it separately.

One note on verification: this machine is deep in swap and got killed by its own memory guardian earlier today, so I did not run the build locally this round and am leaving it to CI. It is a two-line type annotation change with no runtime code.

khandrew1 (approved):

PR #2507, opened 2026-09-10
fix(client): replace SSE response tee with pass-through TransformStream to prevent socket leaks

### LINKED ISSUE #2506: client: HttpConnector.observeSseProgress response.body.tee() leaks HTTP sockets on aborted requests
### Summary

In `@mcp-use/client` (`libraries/typescript/packages/client/src/transport/http.ts`), `HttpConnector` intercepts HTTP responses from MCP servers and tees any `text/event-stream` response using `response.body.tee()` to extract progress notifications:

```typescript
const [body, observed] = response.body.tee();
void (async () => {
  const reader = observed.getReader();
  // ...
  while (true) {
    const { done, value } = await reader.read();
    if (done) break;
    // ...
  }
})();
return new Response(body, ...);
```

### The Problem

Under the WHATWG Streams standard, teeing a `ReadableStream` creates two independent branches. The underlying resource source (the TCP socket and HTTP connection) is **only canceled and closed when both branches are closed or canceled**.

1. **Active Socket Leaks on Aborted / Timed-Out Requests**:
   - The `observed` branch is consumed in a detached, unawaited background async IIFE (`void (async () => { ... })()`).
   - It does not listen to the caller's `AbortSignal` (`init?.signal` or `input.signal`), nor does it link cancellation of `body` back to `observed`.
   - When a client aborts a request mid-flight (e.g. timeout, user cancellation, component unmount, or gateway drop), the consumer cancels `body`, but `observed` remains blocked indefinitely on `await reader.read()`.
   - As a result, the underlying TCP connection and socket are kept open in the background until server keep-alive timeouts fire. In long-running servers, reverse proxies, and agent services, this causes file descriptor exhaustion.

2. **Memory Overhead & Backpressure Drift**:
   - `tee()` clones every raw byte packet into two separate memory queues in JavaScript. If the SDK reads slower than `observed`, or pauses, memory inflates.
   - It runs an ad-hoc regex parser (`buffer.split(/\r?\n\r?\n/)`) concurrently with the official SDK's `EventSourceParserStream`.

### Permanent Architectural Fix

Instead of splitting the stream with `tee()` and managing two divergent lifecycles with manual listener cleanup, replace `observeSseProgress` with a **single in-line pass-through `TransformStream`**:

```typescript
private observeSseProgress(response: Response): Response {
  if (
    !response.body ||
    !response.headers.get("content-type")?.includes("text/event-stream")
  ) {
    return response;
  }
  const decoder = new TextDecoder();
  let buffer = "";
  const observer = new TransformStream<Uint8Array, Uint8Array>({
    transform: (chunk, controller) => {
      buffer += decoder.decode(chunk, { stream: true });
      const events = buffer.split(/\r?\n\r?\n/);
      buffer = events.pop() ?? "";
      for (const event of events) {
        for (const line of event.split(/\r?\n/)) {
          if (!line.startsWith("data:")) continue;
          try {
            const message = JSON.parse(line.slice(5).trim()) as {
              method?: string;
              params?: unknown;
            };
            if (message.method === "notifications/progress") {
              this.forwardRoundProgress(message.params);
            }
          } catch {
            // Ignore malformed/non-JSON chunks
          }
        }
      }
      controller.enqueue(chunk);
    },
  });
  return new Response(response.body.pipeThrough(observer), {
    status: response.status,
    statusText: response.statusText,
    headers: response.headers,
  });
}
```

### Benefits of the In-Line Architecture:
- **Single Stream Owner**: The SDK remains the sole consumer. There is no second branch or detached reader.
- **Native Cancellation Propagation**: Under WHATWG Streams, when the consumer cancels the returned stream or aborts, `pipeThrough` automatically propagates cancellation directly upstream to the underlying HTTP response source and destroys the TCP connection.
- **50% Memory Reduction**: Eliminates memory duplication and buffer desynchronization.
- **Structured Concurrency**: Eliminates unawaited floating background async loops.

rohith500:
Fixes #2506

### Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

### Problem

In `@mcp-use/client` (`libraries/typescript/packages/client/src/transport/http.ts`), `HttpConnector.observeSseProgress` used `response.body.tee()` to extract progress notifications (`notifications/progress`).

Under the WHATWG Streams specification, a teed `ReadableStream` only closes its underlying resource source (the TCP socket and HTTP connection) when **both** branches are closed or canceled. Because the second branch was consumed in a detached, unawaited background loop (`void (async () => { while(true) await reader.read() })()`), any aborted request or consumer cancellation left the second reader hung indefinitely, leaking active TCP sockets and file descriptors.

### Solution

Replaced `response.body.tee()` and the uncoordinated background loop with an **in-line pass-through `TransformStream`**:

1. **Single Reader / Single Lifecycle**: The official SDK (`StreamableHTTPClientTransport`) is the sole stream consumer.
2. **In-Flight Inspection**: As chunks are pulled by the SDK, the `transform` hook inspects the text for `notifications/progress` events and immediately forwards them via `controller.enqueue(chunk)`.
3. **Native Cancellation Propagation**: When the SDK or caller cancels the returned stream (`await response.body.cancel()`) or when the request aborts via `AbortSignal`, WHATWG Streams automatically propagates cancellation upstream to the underlying HTTP source, immediately freeing the TCP socket and connection.
4. **Memory Optimization**: Eliminates dual-buffer memory allocation and backpressure drift caused by `tee()`.

### Testing

Added comprehensive unit test suite in `packages/client/tests/unit/transport/http-sse-progress-lifecycle.test.ts`:
- Verifies progress events are forwarded while delivering all raw SSE chunks intact to the consumer.
- Verifies that calling `await reader.cancel()` propagates cancellation upstream to the underlying stream source immediately.
- Verifies that non-SSE responses bypass the transform stream.
- Verifies graceful handling of empty response bodies, malformed JSON lines, and stream errors.

Ran:
```bash
pnpm --filter @mcp-use/client exec vitest run tests/unit/transport/
pnpm build
pnpm docs:api
pnpm lint
pnpm format:check
```
All suites passed with zero errors or warnings.

---
## Summary by cubic
Fixes an HTTP socket leak in `@mcp-use/client`'s SSE progress observer by replacing `response.body.tee()` with an in-line pass-through `TransformStream`. Previously, aborted requests or consumer cancellation left a detached background reader hanging, leaking TCP sockets and file descriptors. Now the stream has a single reader, so cancellation propagates upstream immediately and frees the connection.

- Progress notifications are still forwarded to `forwardRoundProgress` while all raw SSE chunks pass through unchanged.
- Adds unit tests covering cancellation propagation, non-SSE passthrough, empty bodies, malformed JSON, stream errors, and exact full-stream text preservation.
- Includes a changeset for the `@mcp-use/client` patch release.

<sup>Written for commit c842d52d5310cd01ceb2c686a85c5c534263df02. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2507?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/http-sse-transform-progress-lifecycle.md
+++ b/libraries/typescript/.changeset/http-sse-transform-progress-lifecycle.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Fix HTTP connection and socket leak on aborted requests in `HttpConnector.observeSseProgress` by replacing stream teeing with an in-line pass-through `TransformStream`.
--- a/libraries/typescript/packages/client/src/transport/http.ts
+++ b/libraries/typescript/packages/client/src/transport/http.ts
@@ -584,8 +584,9 @@ export class HttpConnector extends BaseConnector {
   }
 
   /**
-   * Tee an SSE response so v2 MRTR progress can be correlated even when the
-   * upstream SDK does not carry the original callback to retry request IDs.
+   * Observe an SSE response via an in-line pass-through TransformStream so v2 MRTR
+   * progress can be correlated without duplicating stream buffers or leaking sockets
+   * on aborted requests.
    */
   private observeSseProgress(response: Response): Response {
     if (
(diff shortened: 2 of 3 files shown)


khandrew1:
Thanks for investigating this. We’re superseding this PR with #2548, which removes the SSE observer entirely. It was a workaround for SDK v2 beta behavior that upstream now handles.

We also found and fixed a notification-wrapper bug that the observer was masking, so removing it preserves progress delivery without maintaining another stream parser.

PR #2521, opened 2026-09-11
fix(cli): reject an inline value on boolean flags instead of discarding it

### LINKED ISSUE #2520: cli: boolean flags silently discard an inline value, so mcp-use dev --tunnel=false opens a public tunnel
### Summary

`parseArgs` in `libraries/typescript/packages/cli/src/bin/args.ts` splits `--flag=value` for every `--`-prefixed token (lines 98-107), but the boolean cases in the switch (lines 147-164) never read the inline value. They assign unconditionally:

```ts
      case "--tunnel":
        args.tunnel = true;
        break;
      case "--no-open":
        args.open = false;
        break;
```

So `--tunnel=false` is parsed as `--tunnel`. The `false` is discarded and the flag fires anyway.

### The Problem

1. `mcp-use dev --tunnel=false` and `mcp-use start --tunnel=false` open a public tunnel. A tunnel exposes a local dev server to the internet, so the result is the inverse of an explicit instruction, in the direction that widens exposure.
2. There is no diagnostic. Unknown flags throw at line 167, so a user reasonably concludes that an accepted flag was understood.
3. `args.tunnel` reaches a real tunnel: `main.ts:285` passes it to `runStart`, `start.ts:172` gates on `options.tunnel === true`, and `dev.ts:1017` calls `tunnelManager.start(port)`.
4. The same defect affects `--no-open`, `--no-inspector`, `--with-inspector`, `--source-maps`, `--inline`, `--help` and `--version`.
5. The docstring advertises `--flag value` and `--flag=value` as generally supported, without excluding boolean switches.

Reproduced on `main` at `2ef6367`. Unit level:

```
parseArgs(["dev", "--tunnel=false"]).tunnel             // true
parseArgs(["dev", "--no-open=false"]).open              // false
parseArgs(["build", "--source-maps=false"]).sourceMaps  // true
```

End to end through `main()`, with only `@mcp-use/tunnel` stubbed at its module boundary so no real tunnel is created:

```
main(["start", "--path", cwd, "--port", "4567", "--tunnel=false"])
// resolves 0
// createTunnelManager called once
// tunnelManager.start called with 4567
// stdout: "mcp-use public MCP URL: https://public-test.local.mcp-use.run/mcp"
```

Existing coverage in `tests/bin-start.test.ts:227-274` exercises these flags in bare form only; `=value` is tested on string and number flags only (`--port=8080`, `--mcp-dir=src/mcp`).

### Proposed fix

Reject a value on a boolean switch rather than interpret it. The CLI has no truthiness parsing anywhere else, so honouring the value invites follow-up questions about `--tunnel=0` and `--tunnel=no`. `takeValue` records that it consumed the inline value; any flag reaching the end of the loop with an unconsumed inline value is an error.

```ts
    let inlineConsumed = false;

    const takeValue = (): string => {
      if (inline !== undefined) {
        inlineConsumed = true;
        return inline;
      }
      // ...unchanged
    };

    switch (flag) { /* ...unchanged... */ }

    if (inline !== undefined && !inlineConsumed) {
      throw new Error(`${flag} does not take a value`);
    }
```

`--tunnel=false` then exits with `--tunnel does not take a value`. Bare booleans, `--flag value`, and `--flag=value` on value-taking flags are unaffected.

Happy to switch to honouring the value instead if you would prefer that direction.

aishwary-dongre:
Fixes #2520

# Pull Request Description

## Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

`parseArgs` split `--flag=value` for every `--`-prefixed token but the boolean cases never read the inline value, so `mcp-use dev --tunnel=false` parsed as `--tunnel` and opened a public tunnel — the inverse of what was asked, with no diagnostic. Boolean switches now reject an inline value.

## Review Readiness

## Implementation Details

1. `takeValue` sets `inlineConsumed` when it reads the inline value.
2. After the switch, an unconsumed inline value throws `` `${flag} does not take a value` ``. Only the boolean cases can reach that, because they never call `takeValue`.
3. Docstring updated to state that boolean switches take no value, and to list the new throw.

No behaviour change for bare booleans, for `--flag value`, or for `--flag=value` on value-taking flags. No dependencies added or modified.

---

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [x] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

---

## Example Usage (Before)

```bash
$ mcp-use dev --tunnel=false
# `false` is discarded, the tunnel opens anyway:
mcp-use public MCP URL: https://<subdomain>.local.mcp-use.run/mcp
```

## Example Usage (After)

```bash
$ mcp-use dev --tunnel=false
--tunnel does not take a value

# omit the flag to run without a tunnel
$ mcp-use dev
```

## Documentation Updates

None. The behaviour change is described in the `parseArgs` TSDoc, which is where the `--flag=value` support was documented.

## Testing

New test file `libraries/typescript/packages/cli/tests/bin-args-boolean-values.test.ts`, 6 cases:

- `--tunnel=false` and `--tunnel=true` are rejected
- a value is rejected on `--no-open`, `--no-inspector`, `--with-inspector`, `--source-maps`, `--inline`, `--help`, `--version`
- bare boolean switches still parse as before
- inline values still work on `--port`, `--host`, `--path`, `--entry`, `--mcp-dir`, `--views-dir`
- an `=` inside a positional or after `--` is left alone
- an unknown flag is still reported ahead of its inline value

Commands run:

```bash
cd libraries/typescript
pnpm --filter @mcp-use/cli exec vitest run tests/bin-args-boolean-values.test.ts
pnpm --filter @mcp-use/cli exec vitest run
pnpm build && pnpm lint && pnpm format:check
```

Results: the new file is 6 passed, and 2 of those cases fail against unmodified `args.ts`, so the test does pin the regression. `pnpm build`, `pnpm lint` and `pnpm format:check` are clean.

The full `cli` suite is 301 passed, 1 failed. The failure is `tests/cli/dev.test.ts > runDev (views) > discards a candidate superseded while its server entry is evaluating`, which is unrelated to this change: it passes in isolation both with and without the patch, it never calls `parseArgs`, and `src/cli/dev.ts` imports only `resolveHost` and `resolvePort` from `bin/args.js`. It looks like the load-dependent `dev` flake previously discussed in #2273.

Also noticed while running the suite, and deliberately left out of this PR: `packages/cli/tests/cli/.tmp/` is not gitignored and holds unformatted generated fixtures, so `pnpm format:check` fails locally after any `cli` test run until the directory is removed. Happy to open a separate issue for that.

## Backwards Compatibility

Technically breaking for anyone passing `=value` to a boolean flag today, but that invocation currently does the opposite of what it says, so the affected scripts are already misbehaving. Shipped as a `patch` changeset for `@mcp-use/cli`; happy to relabel if you would rather treat it as a minor.

## Related Issues

Closes #2520

---
## Summary by cubic
`@mcp-use/cli` boolean flags now reject an inline `=value` instead of silently discarding it, so `--tunnel=false` fails rather than opening a public tunnel.

- Previously `parseArgs` split `--flag=value` for every `--` token, but boolean cases never read the inline value, so `--tunnel=false` parsed as `--tunnel` and opened a public tunnel.
- The parser now throws `--tunnel does not take a value` for all boolean switches, including `--no-open`, `--no-inspector`, `--with-inspector`, `--source-maps`, `--inline`, `--help`, and `--version`.
- Bare booleans, `--flag value`, and `--flag=value` on value-taking flags are unchanged.
- Adds a regression test for the rejected inline value in the existing `parseArgs` suite.
- Shipped as a `minor` changeset since it breaks previously accepted invocations.
- Closes #2520.

<sup>Written for commit 0fe7a35f601b6f47207eae17fa736a343a737925. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2521?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/quiet-lamps-refuse.md
+++ b/libraries/typescript/.changeset/quiet-lamps-refuse.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Reject an inline value on boolean CLI switches instead of silently discarding it. `mcp-use dev --tunnel=false` previously parsed as `--tunnel` and opened a public tunnel; it now fails with `--tunnel does not take a value`. The same applies to `--no-open`, `--no-inspector`, `--with-inspector`, `--source-maps`, `--inline`, `--help`, and `--version`. Flags that take a value continue to accept both `--flag value` and `--flag=value`.
--- a/libraries/typescript/.changeset/quiet-lamps-refuse.md
+++ b/libraries/typescript/.changeset/quiet-lamps-refuse.md
@@ -1,5 +1,5 @@
 ---
-"@mcp-use/cli": patch
+"@mcp-use/cli": minor
 ---
 
 Reject an inline value on boolean CLI switches instead of silently discarding it. `mcp-use dev --tunnel=false` previously parsed as `--tunnel` and opened a public tunnel; it now fails with `--tunnel does not take a value`. The same applies to `--no-open`, `--no-inspector`, `--with-inspector`, `--source-maps`, `--inline`, `--help`, and `--version`. Flags that take a value continue to accept both `--flag value` and `--flag=value`.
--- a/libraries/typescript/packages/cli/tests/bin-args-boolean-values.test.ts
+++ b/libraries/typescript/packages/cli/tests/bin-args-boolean-values.test.ts
@@ -1,0 +1,77 @@
+/**
+ * Tests for boolean switches given an inline `=value` in the `mcp-use` bin.
+ *
+ * `parseArgs` splits `--flag=value` for every `--` token, but the boolean
+ * cases never read the inline value. `--tunnel=false` therefore used to parse
+ * as `--tunnel` and open a public tunnel — the inverse of what was asked, with
+ * no diagnostic, because unknown flags do throw and so acceptance reads as
+ * support. Boolean switches now reject a value instead.
+ */
+import { describe, expect, it } from "vitest";
+
+import { parseArgs } from "../src/bin/args.js";
+
+describe("boolean switches reject an inline value", () => {
+  it("rejects --tunnel=false rather than opening a tunnel", () => {
+    expect(() => parseArgs(["dev", "--tunnel=false"])).toThrow(
+      "--tunnel does not take a value"
+    );
+    expect(() => parseArgs(["start", "--tunnel=true"])).toThrow(
+      "--tunnel does not take a value"
+    );
+  });
+
+  it("rejects a value on the remaining boolean switches", () => {
+    const cases: readonly string[] = [
+      "--no-open",
+      "--no-inspector",
+      "--with-inspector",
+      "--source-maps",
+      "--inline",
+      "--help",
+      "--version",
+    ];
+    for (const flag of cases) {
+      expect(() => parseArgs(["dev", `${flag}=false`])).toThrow(
+        `${flag} does not take a value`
+      );
+    }
+  });
+
+  it("still accepts boolean switches in bare form", () => {
+    expect(parseArgs(["dev", "--tunnel"]).tunnel).toBe(true);
+    expect(parseArgs(["dev", "--no-open"]).open).toBe(false);
+    expect(parseArgs(["dev", "--no-inspector"]).inspector).toBe(false);
+    expect(parseArgs(["dev", "--with-inspector"]).inspector).toBe(true);
+    expect(parseArgs(["build", "--source-maps"]).sourceMaps).toBe(true);
+    expect(parseArgs(["build", "--inline"]).inline).toBe(true);
+    expect(parseArgs(["dev", "--help"]).help).toBe(true);
+    expect(parseArgs(["dev", "--version"]).version).toBe(true);
+  });
+
+  it("still accepts inline values on flags that take one", () => {
+    expect(parseArgs(["start", "--port=8080"]).port).toBe(8080);
+    expect(parseArgs(["dev", "--host=0.0.0.0"]).host).toBe("0.0.0.0");
+    expect(parseArgs(["dev", "--path=./project"]).path).toBe("./project");
+    expect(parseArgs(["dev", "--entry=src/server.ts"]).entry).toBe(
+      "src/server.ts"
+    );
+    expect(parseArgs(["dev", "--mcp-dir=src/mcp"]).mcpDir).toBe("src/mcp");
+    expect(parseArgs(["dev", "--views-dir=src/views"]).viewsDir).toBe(
+      "src/views"
+    );
+  });
+
+  it("leaves an `=` inside a positional or a passthrough argument alone", () => {
+    expect(parseArgs(["dev=weird"]).command).toBe("dev=weird");
+    expect(
+      parseArgs(["typecheck", "--", "--extendedDiagnostics=false"]).passthrough
+    ).toEqual(["--extendedDiagnostics=false"]);
+  });
+
+  it("keeps reporting an unknown flag ahead of its inline value", () => {
+    expect(() => parseArgs(["dev", "--frobnicate=1"])).toThrow(
+      "Unknown option: --frobnicate"
+    );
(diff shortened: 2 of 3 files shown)


libraries/typescript/.changeset/quiet-lamps-refuse.md
```diff
@@ -0,0 +1,5 @@
+---
+"@mcp-use/cli": patch
```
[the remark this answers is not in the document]
aishwary-dongre: Agreed, and bumped to `minor` in 5463960.

One note on the reasoning rather than the conclusion: a `minor` bump does not stop caret ranges from picking this up silently either, since `^4.1.11` already resolves to `4.2.0`. Only a `major` would. So `minor` is the better label than `patch`, but the caret concern stands regardless of which of the two we pick.

I left it at `minor` on the view that the invocations this now rejects were already doing the opposite of what they asked for, so there is no correct behaviour being taken away. Happy to go to `major` if maintainers would rather be strict about it.

libraries/typescript/packages/cli/tests/bin-args-boolean-values.test.ts
```diff
@@ -0,0 +1,77 @@
+/**
```
khandrew1: remove all the tests we don't need them. just add one simple test to our existing parser suite to catch this bug.

something like this:
```ts
expect(() => parseArgs(["dev", "--tunnel=false"]))
  .toThrow("--tunnel does not take a value");
```

khandrew1 (changes requested):

libraries/typescript/packages/cli/tests/bin-args-boolean-values.test.ts
aishwary-dongre: Done in 0fe7a35. Removed `bin-args-boolean-values.test.ts` and left a single case in the existing `parseArgs` suite in `bin-start.test.ts`, right after `parses --tunnel`:

```ts
  it("rejects a value on a boolean flag instead of ignoring it", () => {
    expect(() => parseArgs(["dev", "--tunnel=false"])).toThrow(
      "--tunnel does not take a value"
    );
  });
```

You were right that the rest was redundant — the bare form of every boolean flag is already covered in that suite, so those cases were re-asserting existing coverage.

It still pins the regression: it fails with `expected [Function] to throw an error` against the unmodified `args.ts`, and passes with the fix. Full `cli` suite is 297 passing across 23 files, with `pnpm lint` and `pnpm format:check` clean.

aishwary-dongre:
Gentle nudge on this one, @khandrew1 — I think it is ready and just needs the review flag cleared.

Your request was addressed in [`0fe7a35`](https://github.com/mcp-use/mcp-use/commit/0fe7a35) on 13 Sept: the separate test file is gone and there is a single case in the existing `parseArgs` suite, using your assertion. You were right that the rest was redundant, since the bare form of every boolean flag is already covered there.

All checks are green, `MERGEABLE`, and the case still pins the bug, failing with `expected [Function] to throw an error` against unmodified `args.ts`. Full `cli` suite is 297 passing.

Also still open from cubic, if you have a preference: it asked for more than a `patch` bump and I moved the changeset to `minor`. `minor` does not actually stop caret ranges picking this up, since `^4.1.11` already resolves to `4.2.0`, so `major` is the only label that would. I left it at `minor` because the invocations this rejects were already doing the opposite of what they asked, but happy to change it.

No rush if you are working through a queue.

PR #2532, opened 2026-09-13
fix(client): propagate SSE observer cancellation

### LINKED ISSUE #2506: client: HttpConnector.observeSseProgress response.body.tee() leaks HTTP sockets on aborted requests

Elioooon:
## Summary

Fixes #2506.

`HttpConnector.observeSseProgress` previously used `ReadableStream.tee()`: the SDK consumed one branch while a detached observer consumed the other. Cancelling the SDK branch could leave the source response open until the observer completed, retaining an HTTP socket.

The observer is now an inline `TransformStream`. It forwards progress notifications while the SDK remains the single downstream consumer, so cancellation propagates back to the original response body.

## Design / boundary

- Reuses the existing SSE parsing and progress forwarding logic; no protocol behavior changes.
- Applies only to ordinary request/response SSE. `subscriptions/listen` remains untouched because the SDK owns its stream hooks and acknowledgement state.
- The regression test creates a source stream whose `cancel()` records the reason, then fails after one second if cancellation is not propagated.

## Validation

- `pnpm --filter @mcp-use/client test -- tests/unit/transport/http-lifecycle.test.ts` — 61 files / 374 tests passed
- `pnpm exec eslint packages/client/src/transport/http.ts packages/client/tests/unit/transport/http-lifecycle.test.ts`
- `pnpm --filter @mcp-use/client build`
- `pnpm exec prettier --check packages/client/src/transport/http.ts packages/client/tests/unit/transport/http-lifecycle.test.ts .changeset/silent-sse-cancellation.md`
- `git diff --check origin/main...HEAD`

The local Node version is 22.16.0 while the workspace declares `>=22.22.2`; the focused suite and build nevertheless completed successfully.

---
## Summary by cubic
Fixes #2506 by changing `observeSseProgress` so cancelling an observed SSE response now closes the source HTTP stream. Previously the method used `ReadableStream.tee()`, which left the source response open until the detached observer finished; it now uses an inline `TransformStream`, keeping the SDK as the single downstream consumer so cancellation propagates upstream.

**Bug Fixes**
- Forwards progress inside `queueMicrotask` so observation stays off the transport's backpressure path, and a throwing progress handler no longer breaks response delivery.
- The change is limited to ordinary request/response SSE; `subscriptions/listen` is untouched because the SDK owns its stream hooks and acknowledgement state.
- Adds regression tests that assert source cancellation propagation, byte-for-byte SSE progress forwarding, and readable bytes when a progress handler throws.

<sup>Written for commit a0ebe5f6ed2a4f1064ec4b8745f8c212c7829a3b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2532?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/libraries/typescript/.changeset/silent-sse-cancellation.md
+++ b/libraries/typescript/.changeset/silent-sse-cancellation.md
@@ -1,0 +1,5 @@
+---
+"@mcp-use/client": patch
+---
+
+Propagate cancellation of observed SSE responses to the upstream HTTP stream.
--- a/libraries/typescript/packages/client/src/transport/http.ts
+++ b/libraries/typescript/packages/client/src/transport/http.ts
@@ -577,8 +577,8 @@ export class HttpConnector extends BaseConnector {
   }
 
   /**
-   * Tee an SSE response so v2 MRTR progress can be correlated even when the
-   * upstream SDK does not carry the original callback to retry request IDs.
+   * Observe SSE progress while preserving a single cancellation path to the
+   * upstream response body. The upstream SDK remains the only consumer.
    */
   private observeSseProgress(response: Response): Response {
     if (
@@ -587,44 +587,33 @@ export class HttpConnector extends BaseConnector {
     ) {
       return response;
     }
-    const [body, observed] = response.body.tee();
-    void (async () => {
-      const reader = observed.getReader();
-      const decoder = new TextDecoder();
-      let buffer = "";
-      try {
-        while (true) {
-          const { done, value } = await reader.read();
-          if (done) break;
-          buffer += decoder.decode(value, { stream: true });
-          const events = buffer.split(/\r?\n\r?\n/);
-          buffer = events.pop() ?? "";
-          for (const event of events) {
-            for (const line of event.split(/\r?\n/)) {
-              if (!line.startsWith("data:")) continue;
-              try {
-                const message = JSON.parse(line.slice(5).trim()) as {
-                  method?: string;
-                  params?: unknown;
-                };
-                if (message.method === "notifications/progress") {
-                  this.forwardRoundProgress(message.params);
-                }
-              } catch {
-                // Ignore malformed/non-JSON SSE data; the SDK remains authoritative.
+    const decoder = new TextDecoder();
+    let buffer = "";
+    const observer = new TransformStream<Uint8Array, Uint8Array>({
+      transform: (chunk, controller) => {
+        buffer += decoder.decode(chunk, { stream: true });
+        const events = buffer.split(/\r?\n\r?\n/);
+        buffer = events.pop() ?? "";
+        for (const event of events) {
+          for (const line of event.split(/\r?\n/)) {
+            if (!line.startsWith("data:")) continue;
+            try {
+              const message = JSON.parse(line.slice(5).trim()) as {
+                method?: string;
+                params?: unknown;
+              };
+              if (message.method === "notifications/progress") {
+                this.forwardRoundProgress(message.params);
               }
+            } catch {
+              // Ignore malformed/non-JSON SSE data; the SDK remains authoritative.
             }
           }
         }
-      } catch (error) {
-        if (!(error instanceof DOMException && error.name === "AbortError")) {
-          logger.debug("Progress observer stream ended:", error);
-        }
-      } finally {
-        reader.releaseLock();
-      }
-    })();
-    return new Response(body, {
+        controller.enqueue(chunk);
+      },
+    });
+    return new Response(response.body.pipeThrough(observer), {
       status: response.status,
       statusText: response.statusText,
       headers: response.headers,
--- a/libraries/typescript/packages/client/tests/unit/transport/http-lifecycle.test.ts
+++ b/libraries/typescript/packages/client/tests/unit/transport/http-lifecycle.test.ts
@@ -108,6 +108,37 @@ async function fixture() {
 }
 
 describe("HTTP lifecycle with a real legacy MCP session", () => {
+  it("propagates cancellation of an observed SSE response to its source", async () => {
+    const connector = new HttpConnector("http://127.0.0.1:1/mcp", {
+      protocolNegotiation: "legacy",
+    });
+    let cancelReason: unknown;
+    let resolveCancelled: () => void;
+    const cancelled = new Promise<void>((resolve) => {
+      resolveCancelled = resolve;
+    });
+    const source = new ReadableStream<Uint8Array>({
+      cancel(reason) {
+        cancelReason = reason;
+        resolveCancelled();
+      },
+    });
+    const response = new Response(source, {
+      headers: { "content-type": "text/event-stream" },
+    });
+
+    const observed = (
+      connector as unknown as {
+        observeSseProgress(response: Response): Response;
+      }
+    ).observeSseProgress(response);
+    const reason = new Error("consumer cancelled");
+    await observed.body?.cancel(reason);
+    await cancelled;
+
+    expect(cancelReason).toBe(reason);
+  });
+
   it("shares one session across concurrent connects and closes its stream on disconnect", async () => {
     const server = await fixture();
     const { connector } = server;
--- a/libraries/typescript/packages/client/tests/unit/transport/http-lifecycle.test.ts
+++ b/libraries/typescript/packages/client/tests/unit/transport/http-lifecycle.test.ts
@@ -134,7 +134,22 @@ describe("HTTP lifecycle with a real legacy MCP session", () => {
     ).observeSseProgress(response);
     const reason = new Error("consumer cancelled");
     await observed.body?.cancel(reason);
-    await cancelled;
+    let cancellationTimeout: ReturnType<typeof setTimeout> | undefined;
+    try {
+      await Promise.race([
+        cancelled,
+        new Promise<never>((_, reject) => {
+          cancellationTimeout = setTimeout(
+            () => reject(new Error("source cancellation was not propagated")),
+            1_000
+          );
+        }),
+      ]);
+    } finally {
+      if (cancellationTimeout) {
+        clearTimeout(cancellationTimeout);
+      }
+    }
 
     expect(cancelReason).toBe(reason);
   });


MohammedAlkindi (commented):
The cancellation fix is right. `tee()` only cancels the source once both branches are cancelled, and the observer branch was read by a detached loop that never cancelled, so the SDK cancelling its branch could never reach upstream. Piping through a TransformStream leaves one consumer and one path back to the source.

Coverage is what I would push on. On main, `observeSseProgress` appears only in `http.ts`, and `forwardRoundProgress` only at its definition in `base.ts` plus the single call site here, so no existing test exercises progress forwarding through this path. The new test cancels a stream that never carries a byte, so nothing asserts that a `notifications/progress` frame still reaches the handler after the rewrite. A case that writes one progress event plus an ordinary data chunk, asserting both the handler call and the bytes arriving at the reader, would pin the half that actually changed.

Separately, the parse and `forwardRoundProgress` now run inline before `controller.enqueue`, so a consumer's progress handler sits on the SSE delivery path rather than on a separate branch. Deliberate?

---

Elioooon:
Added 75dac00a to cover the missing semantic path. The new test feeds a `notifications/progress` SSE frame followed by an ordinary data frame, then asserts both that the active progress handler receives the params and that the reader receives byte-for-byte identical SSE content. Focused client tests pass (61 files, 375 tests); Prettier and ESLint pass for the touched file.

The inline forwarding is deliberate: it preserves source order and keeps the transform as the single reader/cancellation path. `forwardRoundProgress` invokes the same synchronous callback contract already used for SDK notifications, so a user callback can consume time before the next chunk in either path. I did not introduce a detached task queue because that would change notification/data ordering and recreate independent lifecycle work; if asynchronous progress dispatch is desired, it should be a separate contract-level change with explicit ordering semantics.

Elioooon:
Thanks — addressed in a0ebe5f6 (with the progress-path regression coverage added in 75dac00a).

The lifecycle test now writes a `notifications/progress` SSE event followed by an ordinary data event, asserts the handler receives `{ progress: 1 }`, and verifies the reader receives the byte-for-byte concatenation.

The synchronous handler call was not intended to make progress callbacks part of the SSE delivery/backpressure path. The transform now enqueues the original chunk first and schedules progress forwarding in a microtask; callback exceptions are contained so they cannot fail response delivery. A second regression test installs a throwing progress handler and verifies the original SSE bytes remain readable.

Validated on the pushed head:
- `pnpm --filter @mcp-use/client test -- --runInBand tests/unit/transport/http-lifecycle.test.ts` — 61 files / 376 tests passed
- `pnpm exec prettier --check packages/client/src/transport/http.ts packages/client/tests/unit/transport/http-lifecycle.test.ts`
- `git diff --check`

The local Node is 22.16.0 while the workspace declares >=22.22.2; the focused suite passed despite that warning, and the repository Actions run will provide the declared-engine confirmation.

Elioooon:
Addressed both points in the follow-up commits.

- `75dac00a` adds an SSE fixture with a `notifications/progress` frame followed by an ordinary data frame. It asserts the full bytes remain readable and, after the queued observer runs, that the progress handler receives `{ progress: 1 }`.
- `a0ebe5f6` moves progress forwarding into `queueMicrotask`, so parsing/enqueueing stays on the delivery path while handlers remain observational. It also isolates a throwing handler; the response bytes are still readable.

Verified on the current head with:

```sh
pnpm --dir libraries/typescript/packages/client exec vitest run tests/unit/transport/http-lifecycle.test.ts
# 5 passed
```

Elioooon:
Thanks — addressed in the latest two commits. `75dac00a` adds a streamed progress event followed by an ordinary SSE data event, and asserts both the handler payload and byte-for-byte delivery. `a0ebe5f6` moves forwarding to a microtask before enqueueing so a handler cannot add synchronous work or throw on the response delivery path; the transformed stream remains the SDK’s sole consumer, preserving cancellation propagation. I also reran `pnpm --filter @mcp-use/client exec vitest run tests/unit/transport/http-lifecycle.test.ts` (5 passed).

Elioooon:
Thanks — addressed in a0ebe5f6. The regression now writes a `notifications/progress` event followed by an ordinary data event through the observed response, asserts byte-for-byte delivery, then verifies the progress handler received `{ progress: 1 }`. The handler is deliberately scheduled with `queueMicrotask` before `controller.enqueue`: this keeps user callback latency and exceptions out of the TransformStream/backpressure path while retaining the single upstream cancellation path. A separate regression makes the handler throw and confirms the SSE bytes remain readable. The current PR head is green in GitHub checks.

Elioooon:
Thanks — addressed in follow-up commits `75dac00a` and `a0ebe5f6`.

The regression now emits a `notifications/progress` frame followed by an ordinary SSE data frame, asserts the exact response bytes reach the reader, and asserts the progress handler receives `{ progress: 1 }`.

I also changed observation to schedule `forwardRoundProgress` in a microtask before immediately enqueueing the original chunk. That is deliberate: progress callbacks are observational and may be slow or throw; they must not add backpressure to, or prevent, SDK response delivery. The callback still runs after the parsed frame is observed, while the SDK remains the sole stream consumer and retains cancellation propagation.

Validated with `pnpm --filter @mcp-use/client test -- tests/unit/transport/http-lifecycle.test.ts` (61 files, 376 tests passed). The local Node is `22.16.0` versus the repo’s requested `>=22.22.2`, so CI remains the authoritative target-version check.

Elioooon:
Addressed in the latest two commits.

- The lifecycle test now writes both a `notifications/progress` event and an ordinary SSE data event, then asserts the handler receives `{ progress: 1 }` and the reader receives the exact concatenated bytes.
- Progress observation is intentionally queued with `queueMicrotask` before `controller.enqueue`: handlers are observational and must not add latency/backpressure or make SSE delivery fail. A throwing handler is covered too.

Validated with:
```sh
pnpm --dir libraries/typescript --filter @mcp-use/client test:run -- tests/unit/transport/http-lifecycle.test.ts
pnpm --dir libraries/typescript --filter @mcp-use/client build
pnpm --dir libraries/typescript exec eslint packages/client/src/transport/http.ts packages/client/tests/unit/transport/http-lifecycle.test.ts
```
All passed locally.

Elioooon:
Yes — keeping the progress callback off the delivery path is deliberate. `forwardRoundProgress` is observational, so the transform enqueues the original chunk synchronously and schedules the callback in a microtask; a slow or throwing handler cannot delay or fail the SDK's SSE reader.

The focused regression is in `http-lifecycle.test.ts`: it writes a `notifications/progress` frame followed by a normal SSE data frame, asserts that `observed.text()` is byte-for-byte identical to both frames, then asserts the progress handler received `{ progress: 1 }`. I re-ran that focused test locally; the full GitHub TypeScript build, lint, client, Windows CLI, and conformance checks are also green on this head.

Elioooon:
Thanks — addressed in the current head.

`75dac00a` adds a regression that sends both a `notifications/progress` SSE frame and an ordinary data frame through the observed response. It asserts the progress handler receives `{ progress: 1 }` and that the reader receives the exact concatenated original bytes.

For the delivery-path question: I intentionally moved `forwardRoundProgress` into a `queueMicrotask` in `a0ebe5f6`. Progress observation remains best-effort: a slow or throwing handler must not add backpressure to, or interrupt, the SDK’s authoritative SSE stream. The existing throwing-handler case covers that boundary.

Validated locally:
- `pnpm --dir libraries/typescript/packages/client exec vitest run tests/unit/transport/http-lifecycle.test.ts` (5 passed)
- `pnpm --dir libraries/typescript/packages/client exec eslint src/transport/http.ts tests/unit/transport/http-lifecycle.test.ts`
- `git diff --check upstream/main...HEAD`

Elioooon:
Thanks — addressed in `a0ebe5f`.

- Added a regression case with one `notifications/progress` event and one ordinary SSE data event; it asserts both the forwarded progress params and byte-for-byte response delivery.
- The observer now schedules `forwardRoundProgress` with `queueMicrotask` after parsing, so callback work (and callback exceptions) cannot sit in the stream's backpressure path or prevent the SDK reader from receiving bytes.
- The cancellation regression remains: cancelling the SDK-facing response body cancels the upstream source.

Verified with `./node_modules/.bin/vitest run tests/unit/transport/http-lifecycle.test.ts` (5 passed). The repository CI is also green on the branch.

Elioooon:
Thanks — addressed in the latest two commits. The lifecycle test now feeds a `notifications/progress` SSE frame followed by an ordinary data frame, asserts the original bytes are readable unchanged, and asserts the progress handler receives the parsed params.

I also moved progress delivery to a microtask and contain handler exceptions, so an observer cannot sit on the SDK response delivery/backpressure path. I reran `pnpm --filter @mcp-use/client test -- --run tests/unit/transport/http-lifecycle.test.ts`; Vitest completed 61 files / 376 tests successfully (the local Node 22.16.0 is below the repo’s declared 22.22.2 minimum, while GitHub Actions is green on 22.23.1).

Elioooon:
Thanks — addressed in `a0ebe5f`.

The new lifecycle test writes one `notifications/progress` SSE frame followed by an ordinary data frame, then asserts both that the active progress handler receives `{ progress: 1 }` and that the reader receives the byte-identical concatenated stream. Progress forwarding is intentionally queued in a microtask: it is observational and must not add handler latency or exceptions to the SDK's SSE delivery/cancellation path.

Rechecked on the current head:
- `packages/client/node_modules/.bin/vitest run tests/unit/transport/http-lifecycle.test.ts` — 5 passed
- `eslint packages/client/src/transport/http.ts packages/client/tests/unit/transport/http-lifecycle.test.ts` — passed

Elioooon:
Addressed in the current head (`a0ebe5f6`). The regression test now sends a `notifications/progress` frame followed by a regular JSON-RPC data frame, asserts the complete byte stream is still readable, then waits for and asserts the progress callback. The callback is deliberately queued: it stays observational, so a slow or throwing handler cannot place backpressure on SSE delivery. I also added the throwing-handler case to make that boundary explicit.

Validated with `pnpm --dir libraries/typescript --filter @mcp-use/client test -- --runInBand tests/unit/transport/http-lifecycle.test.ts` (61 files, 376 tests passed). The local Node version is 22.16.0 while the workspace requests >=22.22.2; this produced engine warnings only, and the test suite completed successfully.

Elioooon:
Thanks for the careful read. The current head adds that exact regression boundary in `http-lifecycle.test.ts`: it writes a `notifications/progress` SSE frame followed by a normal data frame, asserts the response bytes are unchanged, then flushes the microtask queue and asserts the progress handler receives `{ progress: 1 }`.

The handler is intentionally kept off the delivery path: `forwardRoundProgress` is scheduled with `queueMicrotask` before `controller.enqueue`, and callback exceptions are contained. That preserves byte delivery and source cancellation while making progress observation best-effort. All CI checks are currently green.

Elioooon:
Closing as superseded by the merged #2548. Its SDK notification-handler correction removes the obsolete SSE observer entirely and covers concurrent/retry progress plus cancellation at the owning layer, which is a better long-term resolution than this stream-level cancellation fix. Thanks for the review; the added progress/byte regression coverage confirmed the observer boundary, but should not compete with the upstream approach.

khandrew1:
@Elioooon Could you please keep PR updates to substantive changes or new questions? Several recent comments repeat the same information, which makes the review harder to follow.

Elioooon:
Understood. I will keep any future updates to new substantive changes, evidence, or questions, and avoid repeating prior verification. Thank you for the guidance.

PR #2539, opened 2026-09-14
fix(cli): reject an empty value on flags that take one

### LINKED ISSUE #2538: cli: an empty flag value is accepted, so mcp-use dev --host= binds every interface and --port= picks a random port
### Summary

`takeValue` in `libraries/typescript/packages/cli/src/bin/args.ts` (lines 108-117) returns an inline value without checking whether it is empty:

```ts
    /** Consume the flag's value: inline (`=`) or the next argv token. */
    const takeValue = (): string => {
      if (inline !== undefined) return inline;
      const next = argv[++i];
      if (next === undefined || next.startsWith("-")) {
        throw new Error(`Missing value for ${flag}`);
      }
      return next;
    };
```

So `--host=` yields `""` and `--port=` yields `""`. Neither is treated as a missing value, and both then mean something the user did not ask for.

### The Problem

1. **`--host=` binds every interface instead of loopback.** `resolveListenHost` returns `explicitHost` whenever it is not `undefined`, so `""` survives:

   ```ts
   if (explicitHost !== undefined) return explicitHost;
   const envHost = env.HOST?.trim();
   if (envHost !== undefined && envHost !== "") return envHost;
   return configuredHost ?? DEFAULT_LISTEN_HOST;
   ```

   `start.ts:153` and `dev.ts:361` pass the result to `listen`, and Node binds `""` as `::`:

   ```
   host=""        -> {"address":"::","family":"IPv6","port":50666}
   host=127.0.0.1 -> {"address":"127.0.0.1","family":"IPv4","port":50667}
   ```

   `mcp-use dev --host=` therefore serves on every interface, while `--help` documents `--host <host>  Host to bind (dev/start; default: $HOST or 127.0.0.1)` (`main.ts:90`, and again at 112 and 155). A dev server reachable from the network instead of loopback is a wider exposure than the documented default, with no diagnostic.

2. **The env path is already guarded, the flag path is not.** Note the asymmetry above: `HOST=""` is explicitly skipped via `envHost !== ""`, and `resolveListenPort`'s `parsePort` returns `undefined` for a blank string. Only the flag bypasses these guards, so the same empty value means "fall back to the default" from the environment and "bind everything" from the command line.

3. **`--port=` silently selects an ephemeral port.** `parsePort` in `args.ts:232` does `Number(value)`, and `Number("")` is `0`, which `isValidPort` accepts because `0` is legitimately "pick a free port":

   ```
   parseArgs(["start", "--port="]).port      // 0
   parseArgs(["start", "--port", ""]).port   // 0
   ```

   `mcp-use start --port=` starts on a random port instead of reporting a bad argument. Both the inline and space-separated forms are affected, since the `next.startsWith("-")` check does not reject `""`.

4. Unknown flags throw at line 167, so an accepted flag reasonably reads as an understood one.

Reproduced on `main` at `dcaa0b8`. No existing test passes an empty value to any flag; `tests/bin-start.test.ts:195` covers `--port=8080` only.

For completeness, two adjacent quirks I am **not** proposing to change, since `Number` coercion is the documented behaviour and neither is harmful: `--port=0x10` yields `16` and `--port=1e3` yields `1000`.

### Proposed fix

Treat a blank value as absent, in the one place both forms pass through:

```ts
    const takeValue = (): string => {
      const value = inline !== undefined ? inline : argv[++i];
      if (
        value === undefined ||
        value.trim() === "" ||
        (inline === undefined && value.startsWith("-"))
      ) {
        throw new Error(`Missing value for ${flag}`);
      }
      return value;
    };
```

`--host=` and `--port=` then fail with `Missing value for --host` / `Missing value for --port`, which is the error this function already raises for a missing value. Non-empty values are unaffected, and the flag path now agrees with the `PORT`/`HOST` env handling.

None of the six value-taking flags here (`--port`, `--host`, `--entry`, `--path`, `--mcp-dir`, `--views-dir`) has an "empty clears it" meaning, so nothing depends on the current behaviour. Note this is unlike `servers update --root-dir ""`, where empty deliberately clears a field, but that is a different parser (`node:util` `parseArgs` in `commands/servers.ts`) and is untouched.

Related: #2520 also concerns `takeValue`, from the other direction (boolean flags accepting a value they discard). Separate defect, and PR #2521 touches the same function, so whichever lands second will need a trivial rebase. Happy to fold this into that PR instead if you would prefer one change.

aishwary-dongre:
Fixes #2538

# Pull Request Description

## Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

`takeValue` returned an inline value without checking whether it was blank, so an empty value was accepted as a value. `mcp-use dev --host=` resolved to an empty bind address, which Node binds as `::`, every interface, instead of the documented `127.0.0.1` default. `mcp-use start --port=` became `Number("") === 0`, an ephemeral port. Both now report a missing value.

## Review Readiness

## Implementation Details

1. `takeValue` now resolves the inline or next-token value first, then rejects it when it is `undefined` or blank after `trim()`.
2. The existing `startsWith("-")` guard is kept, scoped to the space-separated form as before, so `--port=-1` still reaches `parsePort` and reports `Invalid port: -1` rather than a missing value.
3. Error message reuses the function's existing `Missing value for ${flag}` wording.

One function, +12/-4. No new dependencies.

Non-empty values are unaffected, and none of the six value-taking flags here (`--port`, `--host`, `--entry`, `--path`, `--mcp-dir`, `--views-dir`) has an "empty clears it" meaning. This is unlike `servers update --root-dir ""`, where empty deliberately clears a field, but that is a different parser (`node:util` `parseArgs` in `commands/servers.ts`) and is untouched.

---

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [x] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

No docs change needed: `--help` already documents `default: $HOST or 127.0.0.1`, which is what this restores.

## Example Usage (Before)

```bash
$ mcp-use dev --host=
# serves on every interface, not loopback:
#   listen("") -> {"address":"::","family":"IPv6"}

$ mcp-use start --port=
# starts on a random ephemeral port, because Number("") === 0
```

## Example Usage (After)

```bash
$ mcp-use dev --host=
Missing value for --host

$ mcp-use start --port=
Missing value for --port
```

## Documentation Updates

None.

## Testing

One case added to the existing `parseArgs` suite in `tests/bin-start.test.ts`, covering `--host=`, `--port=`, and the space-separated `--port ""`.

```bash
cd libraries/typescript
pnpm --filter @mcp-use/cli exec vitest run tests/bin-start.test.ts   # 41 passed
pnpm build && pnpm lint && pnpm format:check                          # clean
```

It pins the regression: against unmodified `args.ts` it fails with `expected [Function] to throw an error`.

I also confirmed the `::` binding against real Node rather than inferring it:

```
host=""        -> {"address":"::","family":"IPv6","port":50666}
host=127.0.0.1 -> {"address":"127.0.0.1","family":"IPv4","port":50667}
```

**One thing to flag honestly.** `tests/cli/build.test.ts` times out locally for me, but it does so on unmodified `main` as well, on a different test each run, so it is unrelated to this change. Examples across runs: `builds a view that contains </script> in source as external assets` on clean `main`, and `resolves host tsconfig aliases in standalone view builds` with the patch applied. Each is a 60s timeout rather than an assertion failure, and it started after a dependency install invalidated my local Vite cache. `tests/bin-start.test.ts`, the file this PR touches, is fully green, and CI is the better signal here.

## Backwards Compatibility

Technically breaking for anyone passing an empty value today, but those invocations are already doing something other than what they say: `--host=` widens the bind address and `--port=` randomises the port. Shipped as a `minor` changeset for `@mcp-use/cli`, consistent with the labelling discussion on #2521.

## Related Issues

Closes #2538

Related: #2520 and PR #2521 concern the same `takeValue` function from the other direction, boolean flags accepting a value they then discard. Separate defect, but whichever lands second will need a trivial rebase. Happy to fold this into #2521 instead if you would rather review one change.

---
## Summary by cubic
Fixes #2538 by making CLI flags that take one value reject empty values instead of silently accepting them. `mcp-use dev --host=` previously bound every interface and `mcp-use start --port=` used an ephemeral port; both now fail with `Missing value for <flag>`. This is technically breaking for anyone relying on an empty value, but those uses already produced unintended behavior.

**Bug Fixes**
- The check applies to both inline and space-separated forms for `--port`, `--host`, `--entry`, `--path`, `--mcp-dir`, and `--views-dir`.
- `--port=-1` still reaches `parsePort` and reports `Invalid port: -1`, not a missing value.
- `servers update --root-dir ""` still allows empty values because that flag deliberately clears a field through a different parser.

<sup>Written for commit 6268aece444f36786630d54ae990dfd5fc01949c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2539?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


aishwary-dongre:
Retargeted this from `main` to `canary`, since that appears to be where behaviour fixes land — you retargeted both #2570 and #2597 to `canary` yourself before merging them, so this saves you doing it a third time. It applies cleanly with no rebase needed.

Happy to move it back if `main` was right after all.

PR #2558, opened 2026-09-16
fix(python): isolate cancelled WebSocket requests

### LINKED ISSUE #2557: Python WebSocket request cancellation can fail unrelated requests
**Describe the bug**

Cancelling a Python `WebSocketConnector` request leaves its cancelled future in `pending_requests`. If that response arrives later, `_receive_messages()` calls `set_result()` or `set_exception()` on the cancelled future. The resulting `InvalidStateError` exits the receiver and is then delivered to other pending requests, so cancelling one tool call can fail unrelated calls on the connection.

Cancellation during `ws.send()` and send failures also leave pending entries behind, because sending occurs outside the cleanup block and `CancelledError` is not an `Exception`.

**To reproduce**

This offline example shows the retained cancelled request:

```python
import asyncio
from unittest.mock import AsyncMock, MagicMock

async def main():
    connector = WebSocketConnector("ws://example.test/mcp")
    sent = asyncio.Event()

    async def send(message):
        sent.set()

    connector.ws = MagicMock()
    connector.ws.send = AsyncMock(side_effect=send)
    request = asyncio.create_task(connector.call_tool("slow_tool", {}))
    await sent.wait()
    request.cancel()
    try:
        await request
    except asyncio.CancelledError:
        pass
    print(len(connector.pending_requests))  # 1; expected 0

asyncio.run(main())
```

For the connection-wide failure, start two requests, cancel the first, and deliver its response before the second response. The second request raises `InvalidStateError` rather than returning its own result. Both late success and late error responses trigger it.

**Expected behavior**

Cancelled or failed requests should release their pending entries. A late response to an already-completed future should not terminate the shared receiver or affect other requests.

**Environment**

- Python 3.12.10, Windows
- mcp-use main `dcaa0b8df5721f11d4d2aefae1f85df2a1903ba5` (package version 1.7.1)
- No live server or model API needed for the reproduction

**Proposed fix**

Put sending and awaiting the response inside a `try/finally` that removes the request. Check `future.done()` before delivering a response, covering the race before the cancelled request resumes its cleanup. I have five regression cases covering cancellation while sending/awaiting, send failure, and late success/error responses; all five fail before the fix and the full 310-test unit suite passes afterward.

Regression tests and prepared fix: https://github.com/RaycarlLei/mcp-use/commit/8ab38c8548f80526d2d39530518682e078a7e3d4

RaycarlLei:
## Changes  Fixes #2557. Cancelling a WebSocket request can leave a cancelled future behind; its late response then stops the receiver with `InvalidStateError` and fails unrelated requests. Clean up pending entries in `finally` (including cancellation/send failure) and skip completed futures when receiving replies.  ## Review readiness  - [x] Python-only change, scoped to the request lifecycle. - [x] Checked existing issues and PRs for duplicate fixes. - [x] Added five deterministic regression cases; all fail before the fix. - [x] No dependency, public API, or unrelated formatting changes.  ## Testing  From `libraries/python`, Python 3.12.10, with `MCP_USE_ANONYMIZED_TELEMETRY=false`:  ```sh pytest tests/unit -q  # 310 passed ruff check . ruff format --check . ```  Ruff checks pass with CI's pinned version, 0.15.22. Repository-root pre-commit hooks also pass. Regressions cover cancellation while sending and awaiting a response, failed sends, and late success/error responses while another request is pending. The WebSocket transport is mocked; no model API or remote server is used.  An additional `ty check mcp_use/client/connectors/websocket.py` reports the existing assignment error at `self.ws = await self._connection_manager.start()`. The unmodified base reports the same error; this patch does not change that code.    --- ## Summary by cubic Fixes #2557 by cleaning up cancelled WebSocket requests so a late response no longer stops the shared receiver or fails unrelated requests.  **Changes** - Pending requests are removed in a `finally` block, covering cancellation and send failures. - The receiver skips completed futures when routing replies, and regression tests cover cancellation, send failures, and late success/error responses.  <sup>Written for commit 8ab38c8548f80526d2d39530518682e078a7e3d4. Summary will update on new commits.</sup>  <a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2558?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>     
Linux verification of the exact PR commit also passed in all four Python 3.11/3.12 × latest/minimum direct-dependency combinations: [fork CI run](https://github.com/RaycarlLei/mcp-use/actions/runs/35088587785). Each latest-dependency job reports 310 passed; each minimum-dependency job reports 295 passed and 1 skipped. CI-pinned Ruff checks pass in every job. Upstream CI still awaits maintainer approval.


khandrew1:

PR #2559, opened 2026-09-16
fix(python): correct Google resource parameter schemas

RaycarlLei:
## Language / Project Scope

- [x] Python

## Changes

Google resource tools currently put `{"input_schema": {"type": "object", "properties": {}}}` into `parameters_json_schema`. The extra `input_schema` wrapper is not a JSON Schema constraint: the resulting schema accepts null, arrays, and scalar values instead of describing an argument object, as Google's function declaration API requires.

Remove that wrapper so resource tools declare an object schema directly, consistent with this adapter's tool and prompt declarations. Resource names, filtering, and execution are unchanged.

## Review Readiness

- [x] Scoped to one behavior fix
- [x] Problem, fix, and validation described below
- [x] No unrelated formatting, generated files, or dependency changes
- [x] Affected Python adapter and tests verified
- [x] Searched for existing PRs addressing this issue

## Python Checklist

- [x] `ruff format --check .`
- [x] `ruff check .`
- [x] Regression and resource execution/filtering tests added
- [x] Root pre-commit hooks passed
- Documentation: no public API change

## Testing

Before the change, the schema regression fails because the serialized Google SDK configuration accepts `null` as function arguments. Afterward, all three new tests pass. The tests also check that invoking the generated tool reads the original MCP resource URI and that disallowed resources are not registered.

Local related adapter/schema tests: 12 passed. Full Windows unit suite: 307 passed, one existing clock-resolution failure in `test_execute_simple_code` (tracked in #2481; its implementation and test are unchanged).

The optional SDK was explicitly installed for verification (`google-genai==2.23.0`). Tests skip when it is absent, matching the optional adapter import. Verification covers SDK serialization and JSON Schema semantics without live Gemini requests or API credentials.

Linux Python 3.11 and 3.12: **308 tests passed on each version**, including all three new Google adapter tests, plus whole-package Ruff lint/format checks. The [fork verification run](https://github.com/RaycarlLei/mcp-use/actions/runs/35091243986) checks out this exact contribution commit and asserts that the optional tests execute rather than skip.

## Backwards Compatibility

No changes to resource names, executor arguments, dependencies, or public APIs.

---
## Summary by cubic
Fixes the Google resource tool schema so it declares an argument object directly instead of wrapping it in `input_schema`, which let the schema accept null, arrays, and scalars.

**Bug Fixes**
- Removes the `input_schema` wrapper from `parameters_json_schema`.
- Adds regression tests for schema validation, original URI read-back, and disallowed resource filtering.
- Resource names, filtering, execution, and public APIs are unchanged.

<sup>Written for commit eb0d744e5b7a731d4d749feb0d11b84997aff48b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2559?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khandrew1:
Thanks for taking the time to contribute. Our focus right now is on the TypeScript SDK, so we're limiting Python PRs to fixes for concrete problems that contributors have run into firsthand. This PR doesn't appear to address that kind of issue, so we're closing it for now, along with the linked issue. If this was prompted by something you hit in practice, please reply on the issue with what you were doing when it came up and how to reproduce it, and we'll reopen the issue and take another look

PR #2561, opened 2026-09-16
fix(oauth): refresh an expired stored token instead of dropping it

### LINKED ISSUE #2560: client(python): OAuth.initialize() drops an expired access token without trying the stored refresh_token, forcing the interactive browser flow on every reconnect
**Describe the bug**

With a stored token file that holds an expired access token together with a `refresh_token`, `OAuth.initialize()` (`libraries/python/mcp_use/client/auth/oauth.py:261`) takes the `logger.debug("Existing token is expired")` branch, skips refresh entirely, and returns `None`. `HttpConnector.connect()` (`libraries/python/mcp_use/client/connectors/http.py:143-147`) then calls `await self._oauth.authenticate()`, which starts the local callback server, opens a browser and waits up to 300 s for a user — so every return trip to an OAuth-protected server re-runs the interactive flow instead of silently refreshing.

`OAuth.refresh_token()` exists (`libraries/python/mcp_use/client/auth/oauth.py:766`, decorated `@telemetry("oauth_refresh_token")`) and works when invoked by hand, but it has **zero call sites anywhere in `mcp_use/`**.

**To Reproduce**

```python
"""Save an expired token + refresh_token, then connect. The refresh path is never taken."""
import asyncio, json, tempfile
from pathlib import Path

from mcp_use.client.auth import oauth as oauth_mod
from mcp_use.client.auth.oauth import OAuth, OAuthClientProvider, ServerOAuthMetadata
from mcp_use.client.connectors.http import HttpConnector

SERVER_URL = "http://127.0.0.1:8199/mcp"
tmp = Path(tempfile.mkdtemp())
storage = oauth_mod.FileTokenStorage(base_dir=tmp)
oauth_mod.FileTokenStorage = lambda *a, **k: storage  # keep tokens out of ~/.mcp_use

hits = []

async def token_endpoint(reader, writer):
    hits.append("POST /token")          # logs every hit on the refresh token endpoint
    await reader.read(65536)
    body = json.dumps({"access_token": "refreshed-token", "token_type": "Bearer",
                       "expires_in": 3600, "refresh_token": "r2"}).encode()
    writer.write(b"HTTP/1.1 200 OK\r\nContent-Type: application/json\r\nContent-Length: "
                 + str(len(body)).encode() + b"\r\nConnection: close\r\n\r\n" + body)
    await writer.drain()
    writer.close()

async def main():
    await storage.save_tokens(SERVER_URL, {
        "access_token": "expired-token", "token_type": "Bearer",
        "expires_at": 1000000000.0,     # long past
        "refresh_token": "r1",
    })
    provider = OAuthClientProvider(id="p", display_name="p", metadata=ServerOAuthMetadata(
        issuer="http://127.0.0.1:8199",
        authorization_endpoint="http://127.0.0.1:8199/authorize",
        token_endpoint="http://127.0.0.1:8199/token",
        code_challenge_methods_supported=["S256"],
    ))
    connector = HttpConnector(SERVER_URL, auth={"oauth_provider": provider.model_dump(), "client_id": "cid"})

    async def fake_authenticate():
        raise RuntimeError("interactive browser flow would start here (300s callback wait)")
    connector._oauth.authenticate = fake_authenticate   # only the human flow is stubbed

    try:
        await connector.connect()
    except Exception as e:
        print("connect() stopped at:", type(e).__name__, e)

    print("token endpoint hits during connect():", hits)
    td = await storage.load_tokens(SERVER_URL)
    print("stored access_token after connect():", td.access_token)
    print("stored refresh_token still present:", td.refresh_token)

    # the refresh path itself works when called by hand (no browser involved)
    server = await asyncio.start_server(token_endpoint, "127.0.0.1", 8199)
    auth = await connector._oauth.refresh_token()
    print("manual OAuth.refresh_token() ->", auth.token.get_secret_value() if auth else None)
    print("token endpoint hits after manual refresh:", hits)
    server.close(); await server.wait_closed()

asyncio.run(main())
```

**Actual output**

Verifier 1 (connect-level, token endpoint the only unstubbed I/O):

```
UNPATCHED: connect() stopped at: HumanNeeded | browser callback wait (300s)
UNPATCHED: human browser-wait entered: True
UNPATCHED: token endpoint calls during connect(): 0
UNPATCHED: Authorization header reaching transport: []
PATCHED:   human browser-wait entered: False / token calls: 1
PATCHED:   Authorization header reaching transport: ['Bearer refreshed-token', 'Bearer refreshed-token']
```

Verifier 2 (independent repro; only `initialize()`/`authenticate()` left unmocked apart from a local token endpoint + hit log):

```
A) initialize() return: None
B) token endpoint hits during initialize: []
C) stored access_token after initialize: expired-token
D) stored refresh_token still present: r1
E) manual refresh_token() -> refreshed-token
F) token endpoint hits after manual refresh: ['POST /token HTTP/1.1']
```

**Expected behavior**

An expired stored access token with a valid `refresh_token` should be refreshed silently; the interactive browser flow should only start when no refresh token is available or the refresh itself fails.

Basis, from this repo:

- `docs/python/client/authentication/index.mdx:37` — "Full OAuth flow with automatic discovery, PKCE, and token refresh."
- `docs/python/client/authentication/index.mdx:78-79` — "3. Open your browser for authorization / 4. Store and refresh tokens automatically"
- `docs/python/client/authentication/bearer.mdx:116` — Bearer-vs-OAuth table: `| **Token refresh** | Manual | Automatic |` (OAuth column = Automatic)
- `libraries/python/mcp_use/client/auth/oauth.py:192` — class docstring: "Manages token storage and refresh automaticlly"
- `libraries/python/mcp_use/client/auth/oauth.py:767` — `refresh_token()` docstring: "Refresh the access token if possible."
- `TokenData.refresh_token` is persisted by `FileTokenStorage`, and the TypeScript side ships a refresher (merged PR #1943 "fix: refresh OAuth tokens with resource") — so this is an unwired path in the Python client, not a deliberate no-refresh design.

- mcp-use version: 1.7.1 (`libraries/python/pyproject.toml`), commit `dcaa0b8df5721f11d4d2aefae1f85df2a1903ba5` on `main`
- Package: `libraries/python` (Python client)
- Python version: requires-python `>=3.11`; reproduced on CPython 3.11
- OS: macOS (darwin 25.6.0)
- Browser: N/A — that is the point; on a headless/CI host the browser never opens and `authenticate()` dies after the 300 s callback timeout instead of refreshing

**Additional context**

Root cause: `libraries/python/mcp_use/client/auth/oauth.py:261`. The `if token_data:` branch only tests `self._is_token_valid(token_data)`; the `else` arm logs "Existing token is expired" and falls straight through to metadata discovery and `return None`, without ever consulting `token_data.refresh_token`. The counterpart on the connector side, `libraries/python/mcp_use/client/connectors/http.py:143-147`, treats a `None` from `initialize()` as "go interactive" and calls `authenticate()`.

Reachability: `MCPClient(config={"mcpServers": {"srv": {"url": ..., "auth": {"client_id": ...}}}})` → `create_session()` → `HttpConnector.connect()`. Any second process run after the stored access token expires — the normal return trip for every OAuth-protected server — hits `oauth.py:261`.

Approach I would take in a PR (no diff here): in `OAuth.initialize()`, once metadata is available (note `refresh_token()` returns `None` when `self._metadata` is unset, so discovery has to happen before or be forced), if `token_data` exists, is not valid, and carries a `refresh_token`, call `await self.refresh_token()` and return that `BearerAuth` when it is not `None`, only falling through to the `None` return when the refresh fails. Equivalently, let `HttpConnector.connect()` try `oauth.refresh_token()` before `oauth.authenticate()`. Would also add a unit test with a stubbed token endpoint asserting no interactive flow is started for an expired token that has a `refresh_token`. Happy to open the PR if that approach looks right — or happy to be assigned.

Related: #1943 (merged, TS-side OAuth token refresh with resource — the behavior this Python path is missing), #2258 (`fix(client): refresh OAuth discovery on every new challenge`), #2534 (separate bug in `getOAuthTokenExpiry`, expired-token handling in the TS client). No existing issue or PR covers the Python stored-refresh-token path.

Repro script used by the verifiers: `/tmp/mcpuse-hunt/oauth_refresh_repro.py`. Every claim above was checked against the source at commit `dcaa0b8df5721f11d4d2aefae1f85df2a1903ba5`.

BlueX888:
Closes #2560

## Language / Project Scope

Check all that apply:
- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

`OAuth.initialize()` now attempts a refresh when the stored token is expired and carries a
`refresh_token`, returning the refreshed `BearerAuth` instead of `None`. Previously it logged
`"Existing token is expired"`, skipped refresh entirely, and returned `None`, so
`HttpConnector.connect()` fell through to `OAuth.authenticate()` — the interactive browser flow.

## Review Readiness

## Implementation Details

1. `libraries/python/mcp_use/client/auth/oauth.py` — 8 added lines in `OAuth.initialize()`, placed
   after metadata discovery and before the final `return None`. The refresh is guarded by
   `token_data and token_data.refresh_token`; if `refresh_token()` returns `None` (rejected or
   unusable refresh token) the method keeps its previous behaviour and returns `None`, so the
   caller still falls back to `authenticate()`.
2. No new dependencies, no signature changes, no renames.

**Root cause.** `libraries/python/mcp_use/client/auth/oauth.py:260-261` — the expired branch only
logs and falls through; `token_data.refresh_token` is never consulted. The refresh implementation
already existed at `libraries/python/mcp_use/client/auth/oauth.py:765` (`refresh_token()`,
decorated `@telemetry("oauth_refresh_token")`) but had no call site anywhere in `mcp_use/`.

**Ordering.** The refresh must run after the discovery block at
`libraries/python/mcp_use/client/auth/oauth.py:266-270`: `refresh_token()` returns early when
`self._metadata` is unset, and it needs `metadata.token_endpoint` to reach the token service.

## Python Checklist

- [ ] Updated documentation if needed — not needed; no public API or documented behaviour changed

## Example Usage (Before)

```python
# ~/.mcp_use/tokens/<host>_mcp.json holds an expired access_token plus a refresh_token.
connector = HttpConnector("https://mcp.example.com/mcp", auth={...})
await connector.connect()
# OAuth.initialize() -> None, HttpConnector.connect() -> OAuth.authenticate()
# -> local callback server, browser opened, up to 300 s wait for a human,
#    on every reconnect, even though the refresh token was still usable.
```

## Example Usage (After)

```python
connector = HttpConnector("https://mcp.example.com/mcp", auth={...})
await connector.connect()
# OAuth.initialize() -> BearerAuth("refreshed-token") from the stored refresh_token;
# the new token (and a rotated refresh_token, if the server sends one) is persisted.
# The browser flow only starts when there is no refresh_token or the refresh is rejected.
```

## Documentation Updates

None.

## Testing

Unit tests added in `libraries/python/tests/unit/test_oauth_initialize.py` (5 tests, `pytest-asyncio`,
matching `pytest.ini`'s `asyncio_mode = auto` and the existing `tests/unit` layout). `AsyncOAuth2Client`
is patched, so no network is involved.

Covered: refresh on an expired token (asserts the call reaches `metadata.token_endpoint` with the
stored refresh token, that the returned `BearerAuth` carries the new access token, and that the new
and rotated tokens are persisted); refresh when metadata comes from discovery rather than
`oauth_provider`; `None` when there is no refresh token; `None` when the refresh is rejected with
`OAuth2Error` (asserting the refresh was actually attempted); and an unexpired token still short-
circuiting without any client being constructed.

Commands and results (Python 3.12, `libraries/python`):

```
# new test, unpatched tree (source restored from main) — 3 of the 5 fail:
$ MCP_USE_ANONYMIZED_TELEMETRY=false python -m pytest tests/unit/test_oauth_initialize.py -q
FAILED tests/unit/test_oauth_initialize.py::TestOAuthInitializeExpiredToken::test_refreshes_expired_token_with_stored_refresh_token
FAILED tests/unit/test_oauth_initialize.py::TestOAuthInitializeExpiredToken::test_refreshes_expired_token_when_metadata_is_discovered
FAILED tests/unit/test_oauth_initialize.py::TestOAuthInitializeExpiredToken::test_failed_refresh_returns_none
3 failed, 2 passed, 5 warnings in 0.42s

# new test, patched tree:
$ MCP_USE_ANONYMIZED_TELEMETRY=false python -m pytest tests/unit/test_oauth_initialize.py -q
5 passed, 5 warnings in 0.37s

# existing unit suite (CONTRIBUTING.md §4):
$ MCP_USE_ANONYMIZED_TELEMETRY=false python -m pytest tests/unit -q
1 failed, 309 passed, 37 warnings in 24.75s
```

The one failure, `tests/unit/test_config.py::TestConnectorCreation::test_create_sandboxed_stdio_connector`,
is pre-existing and unrelated: it reproduces identically on a stashed (unpatched) tree, with
`ImportError: E2B SDK (e2b-code-interpreter) not found ... (mcp_use/client/connectors/sandbox.py:110)`
— an optional extra that isn't installed in the verification environment.

```
$ ruff check .
All checks passed!

$ ruff format --check .
225 files already formatted

$ ty check
Found 203 diagnostics
```

`ty` reports the same 203 diagnostics before and after the change (all pre-existing, in
`mcp_use/telemetry/`); none are in `mcp_use/client/auth/oauth.py`.

End-to-end check of the issue's repro — expired token plus `refresh_token` in `FileTokenStorage`,
with a real HTTP token endpoint on `127.0.0.1:8199` and only `authenticate()` stubbed:

```
unpatched: initialize() -> None            | token endpoint hits: 0 | stored access_token: expired-token       | refresh_token: r1
patched:   initialize() -> refreshed-token | token endpoint hits: 1 | stored access_token: refreshed-token   | refresh_token: r2
```

## Backwards Compatibility

Backwards compatible. The only behavioural change is on the path that previously returned `None`
and forced the interactive flow: that path now first tries the already-implemented `refresh_token()`.
Every other branch of `initialize()` is untouched. A server that rejects the refresh still ends up
in `authenticate()` exactly as before.

## Related Issues

Closes #2560

---
## Summary by cubic
Fixes #2560: `OAuth.initialize()` now refreshes an expired stored OAuth token when a refresh token is available, instead of returning `None` and triggering the interactive browser flow. Successful refreshes return the new auth and persist updated tokens; missing or rejected refresh tokens retain the existing fallback.

- Adds unit coverage for discovered and configured metadata, token rotation, failed refreshes, and valid-token handling.

<sup>Written for commit a3b3f233e69bbb2088590bbc3a14e21a1b9e145a. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2561?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khandrew1:

PR #2568, opened 2026-09-17
fix(server): forward cancellation to proxied prompts

### LINKED ISSUE #2473: server: mcp-proxy drops AbortSignal cancellation for proxied prompts
### Summary
In `@mcp-use/server` (`packages/server/src/mcp-proxy.ts`), `server.proxy()` mounts upstream MCP connections. For tools and resources, requests forward the client request's cancellation signal:
- `callTool` forwards `{ signal: ctx.signal, onprogress: ... }`
- `readResource` forwards `{ signal: ctx.signal }`

However, in `mountPlan`, `registerPrompt` drops `ctx` entirely:
```typescript
host.registerPrompt(
  { ... },
  async (params) => plan.connection.getPrompt(upstreamName, params)
);
```
Furthermore, the `ProxyConnection.getPrompt` interface omits `options?: ProxyRequestOptions`.

As a result, when a downstream client disconnects or aborts a prompt request (or when a gateway/proxy timeout occurs), the upstream server never receives the cancellation signal and continues generating the prompt as a zombie task. In LLM applications, prompt generation often triggers upstream sampling or expensive computation, wasting tokens and server capacity.

### Expected Behavior
1. `ProxyConnection.getPrompt` should accept `options?: ProxyRequestOptions` mirroring `readResource` and `callTool`.
2. Proxied prompt handlers should pass `{ signal: ctx.signal }` to `connection.getPrompt()`.

Elioooon:
## Summary

Fixes #2473. Proxied prompt handlers now pass the downstream `AbortSignal` to `ProxyConnection.getPrompt`, matching existing tool and resource forwarding.

## Design

The signal is optional in the structural proxy contract, so existing connection implementations remain compatible. The forwarding is intentionally limited to request cancellation; prompts have no progress callback surface.

## Validation

- `pnpm --filter @mcp-use/client build`
- `pnpm --filter mcp-use test --run tests/proxy.test.ts -t "forwards downstream cancellation"`
- `pnpm --filter mcp-use typecheck`

The targeted regression asserts that the exact downstream signal reaches the proxied prompt call. A full `proxy.test.ts` run was attempted; six pre-existing integration cases could not load `#mcp-use-skills-loader` in this fresh checkout because the CLI build artifact is absent.

## Risk boundary

This changes only the low-level proxy path. It does not alter prompt results, namespacing, tool/resource behavior, or cancellation behavior for callers that do not provide a signal.

---
## Summary by cubic
Fixes #2473 by forwarding downstream cancellation signals to proxied prompt requests. Proxied prompt handlers now pass the AbortSignal to the upstream connection, matching tool and resource forwarding; previously cancellation was not forwarded. The signal is optional so existing connections stay compatible, and a regression test verifies the exact signal reaches the upstream prompt call.

<sup>Written for commit a27cdac17ed6ec10dc6def2b2cdfaad898fa5ecf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2568?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MohammedAlkindi (commented):
Checked the gap and it is real. `callTool` and `readResource` on `main` both already take `ProxyRequestOptions`, and `mountPlan` already passes `{ signal: ctx.signal }` to each, so `getPrompt` was the one proxied request that dropped downstream cancellation. This lines it up with the other two rather than adding a new pattern.

I also checked that the signal reaches the wire, since widening the structural interface alone would not prove that. `getPrompt` in `client/src/core/session.ts` and in `client/src/transport/base.ts` both already declare `options?: RequestOptions`, so the implementation was ahead of the `ProxyConnection` contract and the third argument is consumed rather than ignored.

One thing on the test. The only assertion is `expect(options?.signal).toBe(signal)` inside the fake `getPrompt`, so if the mounted callback ever stopped reaching `getPrompt`, the test would pass having run zero assertions instead of failing. A call counter, or `expect.assertions(1)`, would make it fail closed.

---

Elioooon:
Addressed in a27cdac1. The prompt-forwarding test now counts `getPrompt` invocations and requires exactly one after the mounted callback runs, so it fails if the callback no longer reaches the upstream connection.

Validation: `pnpm --dir libraries/typescript --filter @mcp-use/client build` and `pnpm --dir libraries/typescript --filter mcp-use build` pass; Prettier and `git diff --check` pass. The focused proxy suite exercised this case successfully, but six pre-existing cases are blocked locally by the unresolved `#mcp-use-skills-loader` module during server startup; CI remains the authoritative full-suite run.

Elioooon:
Thanks — addressed in a27cdac1. The regression now increments `getPromptCalls` and asserts it is exactly 1 after invoking the mounted prompt, so it fails if the callback no longer reaches the proxied connection.

Validated with:
```
pnpm --filter mcp-use exec vitest run tests/proxy.test.ts -t "forwards downstream cancellation to proxied prompts"
pnpm --filter mcp-use typecheck
git diff --check origin/main...HEAD
```

The focused regression and typecheck pass. The unfiltered `proxy.test.ts` run has six pre-existing failures when unrelated cases dynamically import `#mcp-use-skills-loader`; the new focused test does not exercise that path.

Elioooon:
Thanks — addressed in a27cdac. The prompt proxy test now counts getPrompt invocations and asserts exactly one after invoking the mounted callback, so it fails if the callback no longer reaches the connection. I re-ran the focused test and test TypeScript check:

```text
pnpm exec vitest run tests/proxy.test.ts --testNamePattern "forwards downstream cancellation to proxied prompts"
pnpm exec tsc -p tsconfig.test.json --noEmit
```

Elioooon:
Thanks — addressed in a27cdac by adding `getPromptCalls` and asserting it is exactly one after invoking the mounted callback, so the test now fails if the callback no longer reaches `getPrompt`. I re-ran the focused test file locally: the new cancellation test passes, while six pre-existing cases remain blocked by a missing generated `#mcp-use-skills-loader` module in this checkout. `pnpm run typecheck` passes.

Elioooon:
Addressed in a27cdac: the fake `getPrompt` now increments an explicit call counter, and the test asserts it was invoked exactly once after the mounted callback runs. That makes the regression fail if the callback stops reaching the proxied request rather than passing with only the in-body signal assertion.

Verified with:
```text
pnpm --dir packages/server exec vitest run tests/proxy.test.ts -t "forwards downstream cancellation to proxied prompts"
```
(1 passed.)

Elioooon:
Implemented in a27cdac: the proxy prompt test now counts `getPrompt` calls and asserts exactly one invocation, so a disconnected mount callback fails the test instead of passing with no assertions. Re-verified with `pnpm --dir libraries/typescript/packages/server exec vitest run tests/proxy.test.ts -t "forwards downstream cancellation to proxied prompts"` and `pnpm --dir libraries/typescript/packages/server run typecheck`.

Elioooon:
Implemented the fail-closed test guard in `a27cdac`: the proxied `getPrompt` fake now requires `expect.assertions(1)`, so the test fails if the mounted callback stops reaching the proxy. The focused change is on the current head, and the full PR check suite is green.

Issue #2499, opened 2026-09-09
ci: Test with Deno (PR Preview) hangs for 18m and fails with 408 Request Timeout on esm.sh


rohith500:
**Describe the bug**

In `.github/workflows/typescript-pr-preview.yml`, the `Test with Deno (PR Preview)` job frequently hangs for ~18 minutes and fails across PRs (e.g. #2488, #2498, and Dependabot PRs) due to an external `408 Request Timeout` on `esm.sh` when attempting to fetch the ephemeral preview package published to `pkg.pr.new`.

Even when all 34 core test suites (TypeScript Conformance Tests v2, TypeDoc validation, unit/integration test suites on Ubuntu and Windows, Knip, lint, build, changesets) pass 100% green, the overall PR check status is marked with a red failure because of this external third-party CDN timeout.

---

**Evidence across Recent PRs & Workflow Runs**

1. **PR #2488** ([Run 34364889325](https://github.com/mcp-use/mcp-use/actions/runs/34364889325/job/102511512514) & [Run 34368728244](https://github.com/mcp-use/mcp-use/actions/runs/34368728244/job/102524638436)):
   - 34 of 35 checks passed.
   - `Test with Deno (PR Preview)` hung for 17m 51s and failed with `408 Request Timeout`.
2. **PR #2498** ([Run 34370061707](https://github.com/mcp-use/mcp-use/actions/runs/34370061707/job/102529196271)):
   - Core maintainer PR faced the same 18-minute hang in `Wait for esm.sh to pick up the package`.
3. **Dependabot PRs** ([Run 34337582315](https://github.com/mcp-use/mcp-use/actions/runs/34337582315)):
   - Failed with the same preview package fetch timeout.

**Failure Log**:
```text
Waiting for esm.sh to pick up the preview package...
Checking URLs:
  PR URL: https://esm.sh/pr/mcp-use/mcp-use@2488
  Commit URL: https://esm.sh/pr/mcp-use/mcp-use@fe032de
Waiting for packages... (attempt 1/10, PR: false, Commit: false)
...
Waiting for packages... (attempt 10/10, PR: false, Commit: false)
⚠️ Packages didn't become fully available after 10 attempts (5 minutes)
  PR URL ready: false
  Commit URL ready: false
Proceeding with tests anyway...

Pre-caching Deno dependencies to avoid download timeouts during tests...
Cache attempt 1/3...
...
⚠️ Failed to cache after 3 attempts
[...]

PR #2570, opened 2026-09-17
fix(cli): resolve the account default organization in whoami

### LINKED ISSUE #2569: cli: whoami reports organization: null for the organization every other command resolves
### Summary

Three places in the CLI resolve the active organization. Two fall back to the account default; `mcp-use whoami` does not.

`src/commands/identity.ts:277-280`:

```ts
  const config = await readCloudConfig();
  const identity = await (await CloudApi.create()).identity();
  const organization =
    identity.organizations.find((item) => item.id === config.orgId) ?? null;
```

`src/commands/organizations.ts:85-88`, for `org current`:

```ts
      const organization =
        identity.organizations.find(
          (item) => item.id === (config.orgId ?? identity.defaultOrganizationId)
        ) ?? null;
```

`src/commands/cloud-api.ts:262-268`, in `cloudApiForOrganization`, which every `servers` and `deployments` command goes through:

```ts
      : identity.organizations.find(
          (item) => item.id === (config.orgId ?? identity.defaultOrganizationId)
        );
```

### The Problem

1. **The divergence is reachable without any unusual setup.** `CloudApi.create` accepts `MCP_USE_API_KEY` from the environment, so authentication does not require `mcp-use login` to have ever written `config.json`. With no local config, `config.orgId` is `undefined` and the account default is used everywhere except `whoami`.

2. **`whoami` then denies the organization the CLI is actively using:**

   ```
   export MCP_USE_API_KEY=mcp_...      # never ran `mcp-use login`
   mcp-use whoami --json               # {"userId":...,"email":...,"organization":null}
   mcp-use org current                 # Acme (acme)
   mcp-use servers list                # lists Acme's servers
   ```

   The human output at `src/commands/identity.ts:286-290` also drops the ` — <name>` suffix, so `whoami` prints only the email.

3. `whoami` is the command users run to answer "who am I and where am I pointed", which is precisely the question it answers incorrectly here. It is also the natural thing to check before a destructive command like `servers delete`, where believing no organization is selected is the wrong impression to leave.

Reproduced on `main` at `dcaa0b8`. With `readCloudConfig` returning `{}` and `identity.defaultOrganizationId` set to `org_1`, `runIdentity("whoami", ["--json"])` prints:

```json
{"userId":"user_1","email":"person@example.com","organization":null}
```

**Test coverage:** none. `tests/commands/identity.test.ts` covers logout and `login --device-code` only; `whoami` appears nowhere except `tests/commands/help-contract.test.ts`, which just checks `-h`.

### Proposed fix

Use the same expression the other two resolvers already use:

```ts
  const organization =
    identity.organizations.find(
      (item) => item.id === (config.orgId ?? identity.defaultOrganizationId)
    ) ?? null;
```

An explicit `mcp-use org use` selection still wins, since `config.orgId` is checked first. The only behaviour change is that `whoami` now names the organization it is actually pointed at instead of reporting none.

Worth noting one alternative reading: that `whoami` deliberately reports only local config, and `null` means "nothing selected locally". If that is the intent, the inconsistency is still worth closing, because `org current` reads local config too and does apply the fallback. Happy to take it in whichever direction you prefer.

aishwary-dongre:
Fixes #2569

# Pull Request Description

## Language / Project Scope

- [x] TypeScript
- [ ] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

`mcp-use whoami` matched `config.orgId` only, while `org current` and `cloudApiForOrganization` both fall back to `identity.defaultOrganizationId`. Authenticating with `MCP_USE_API_KEY` without running `mcp-use login` leaves `config.orgId` undefined, so `whoami` reported `organization: null` for the organization every `servers`/`deployments` command was operating on.

## Review Readiness

## Implementation Details

1. `src/commands/identity.ts` now resolves the organization with `config.orgId ?? identity.defaultOrganizationId`, the same expression as `organizations.ts:85-88` and `cloud-api.ts:262-268`.
2. An explicit `mcp-use org use` selection still wins, because `config.orgId` is checked first.

One expression changed, plus a comment naming the two resolvers it now matches. No new dependencies.

---

### Packages Modified

- [ ] `docs`
- [x] `tests`
- [x] `cli`
- [ ] `create-mcp-use-app`
- [ ] `mcp-use` (server)
- [ ] `mcp-use` (client)
- [ ] `inspector`

### Pre-commit Checklist

## Example Usage (Before)

```bash
export MCP_USE_API_KEY=mcp_...   # never ran `mcp-use login`

$ mcp-use whoami --json
{"userId":"user_1","email":"person@example.com","organization":null}

$ mcp-use org current
Acme (acme)

$ mcp-use servers list
# lists Acme's servers
```

## Example Usage (After)

```bash
$ mcp-use whoami --json
{"userId":"user_1","email":"person@example.com","organization":{"id":"org_1","name":"Acme","slug":"acme","role":"admin"}}

$ mcp-use whoami
person@example.com — Acme
```

## Documentation Updates

None.

## Testing

One case added to the existing `tests/commands/identity.test.ts`, in a new `whoami` describe block, since that file previously covered only logout and `login --device-code`. `whoami` had no test anywhere except the `-h` check in `help-contract.test.ts`.

The existing mock needed two small changes to reach `whoami`: `CloudApi.create` added alongside `withApiKey`, and `readCloudConfig` hoisted so a test can set it. Both are additive; the existing cases are untouched and still pass.

```bash
cd libraries/typescript
pnpm --filter @mcp-use/cli exec vitest run tests/commands/identity.test.ts   # 11 passed
pnpm lint && pnpm format:check                                                # clean
```

It pins the regression: against unmodified `identity.ts` it fails with `expected { userId: 'user_1', …(2) } to match object { organization: { id: 'org_1', …(1) } }`, received `"organization": null`.

## Backwards Compatibility

Backwards compatible. An explicit `org use` selection is still honoured first, and the only change is that `whoami` names the organization it is already pointed at rather than reporting none.

One alternative reading, in case you see it differently: that `whoami` intentionally reports only local config, so `null` means "nothing selected locally". If that is the intent I am happy to close this, though `org current` also reads local config and does apply the fallback, so the inconsistency would still be worth resolving one way or the other.

## Related Issues

Closes #2569

---
## Summary by cubic
Fixes `mcp-use whoami` reporting `organization: null` when authenticated with `MCP_USE_API_KEY` without running `mcp-use login`. It now resolves the organization the same way `org current` and every `servers`/`deployments` command do: falling back to the account default when the local config has no explicit selection.

- `whoami` now uses `config.orgId ?? identity.defaultOrganizationId`, matching `org current` and `cloudApiForOrganization`.
- An explicit `mcp-use org use` selection still takes precedence since `config.orgId` is checked first.
- Adds a regression test covering the API-key-only auth path.
- Updates the CLI reference docs to describe the account-default fallback and the `MCP_USE_API_KEY` case.

<sup>Written for commit bec317cf7c1711fd3221a0cf24da9ca82d9fe469. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2570?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


khandrew1:
Thanks for the contribution. The fallback is consistent with the other commands, but the added test mocks authentication and the identity response. Please add a verified reproduction using a valid API key and a clean local CLI config to #2569 so we can confirm the reported behavior end to end before merging.

aishwary-dongre:
Posted the end-to-end reproduction on #2569: https://github.com/mcp-use/mcp-use/issues/2569#issuecomment-5720468532

Short version: real CLI binary as its own process, a genuinely empty config via a temp `HOME`, and a real HTTP request carrying the API key. No `vi.mock` anywhere. On `main` the same run produces `whoami -> organization: null` alongside `org current -> Acme (acme)`, which is the disagreement rather than an unset selection.

Flagged there, and worth repeating here: I have no production API key, so the cloud API is a local stub answering `GET /api/v1/test-auth` in the shape `CloudApi.identity()` destructures. That means I have not verified the live API returns a non-null `default_profile_id` when there is no local selection. If it does not, this is a consistency fix rather than a user-visible bug. `cloudApiForOrganization` and `org current` both already rely on that field, which is suggestive, but I could not confirm it and did not want to imply otherwise.

aishwary-dongre:
Flagging that the red check here is infrastructure, not this change.

`Test with Deno (PR Preview)` failed with:

```
error: Module not found "https://esm.sh/pr/mcp-use/mcp-use@2570?no-dts&external=zod".
```

That is the esm.sh preview package never becoming resolvable, which is the open #2499. Three things make it unrelated to this diff:

- the same check passed on my other two open PRs, #2539 in 22s and #2555 in 19s
- this PR touches `commands/identity.ts`, its test, and a changeset — nothing in the `mcp-use` server package that `tests/deno/basic-server.test.ts` imports
- the failure is a CDN resolution error before any test executes

Every other check is green. A re-run should clear it; I cannot trigger one from a fork. Happy to push an empty commit if that is easier, though I would rather not add noise to the history.

MohammedAlkindi (commented):
Checked the parity claim rather than taking the changeset's word for it. `commands/organizations.ts` on main already resolves with `identity.organizations.find((item) => item.id === (config.orgId ?? identity.defaultOrganizationId)) ?? null`, which is byte-identical to the new `whoami` expression, and a second site in that same file marks the selected organization with the same expression plus a trailing `?? undefined`. So this closes a real divergence against an existing rule rather than inventing a third one.

The test cannot fail for the right reason, though. The hoisted fixture carries exactly one organization, `org_1`, with `defaultOrganizationId: "org_1"`, so `organization: { id: "org_1", name: "Acme" }` also passes under `organizations[0]`, under `organizations.find(() => true)`, and under a hardcoded `org_1`. None of those is the behavior the changeset describes. Two organizations, with `defaultOrganizationId` pointing at the one that is not first in the list, would pin the fallback you want. The `login` tests further down the same file already build that two-org shape, so most of the fixture work is there.

---

aishwary-dongre:
Live verification is up on #2569: https://github.com/mcp-use/mcp-use/issues/2569#issuecomment-5725373745

Two things you should know before deciding on this PR.

**Your skepticism was justified.** On a brand-new account the live API returns `default_profile_id: null`, and before/after are byte-identical — this fix changes nothing in that state, and `org current` fails alongside `whoami`, so there is no divergence at all. My earlier stub reproduction hid that by hard-coding the field to a real value. That was a genuine flaw in the report, not just missing evidence.

**It does reproduce, but only after `org use` has run once.** That calls `setDefaultOrganization`, so the account default becomes non-null server-side. On any machine with no local config after that — new laptop, container, CI with only `MCP_USE_API_KEY` — `org current` names the organization while `whoami` reports `null`. Live before/after for that case is in the issue comment.

So the accurate scope is narrower than I filed: not a fresh-signup problem, but a real one for any environment without local config once a default exists. Entirely reasonable to close this if that is too narrow to be worth the change; I would rather it close on accurate grounds than merge on the overstated version.

libraries/typescript/packages/cli/src/commands/identity.ts
```diff
@@ -276,8 +276,14 @@ async function whoami(argv: readonly string[], json: boolean): Promise<number> {
   });
   const config = await readCloudConfig();
   const identity = await (await CloudApi.create()).identity();
+  // Resolve the organization the way `org current` and
```
khandrew1: remove this comment

khandrew1 (changes requested):

aishwary-dongre:
Thanks @MohammedAlkindi — that is a more careful check than the changeset deserved, and you are right that `organizations.ts` already carries the identical expression, which is exactly why this is closing a divergence rather than inventing a rule.

@khandrew1 both review points are addressed in bec317c: the explanatory comment is gone, and the `whoami` note in the v2 CLI reference now records the account-default fallback, since it previously said only that the command reads `config.json`. Tests, lint and format are clean.

libraries/typescript/packages/cli/src/commands/identity.ts
aishwary-dongre: Removed in bec317c.

libraries/typescript/.changeset/brave-otters-report.md
```diff
@@ -0,0 +1,5 @@
+---
+"@mcp-use/cli": patch
+---
+
+Resolve the organization in `mcp-use whoami` the same way `org current` and every `servers`/`deployments` command already do, by falling back to the account default when the local config has no explicit selection. Authenticating with `MCP_USE_API_KEY` without running `mcp-use login` previously made `whoami` report `organization: null` while the rest of the CLI read and mutated that organization's resources.
```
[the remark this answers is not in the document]
aishwary-dongre: Added in bec317c, kept to one sentence.

The existing `whoami` note read only "The command reads `~/.mcp-use/config.json`", which this change makes incomplete, so updating it is warranted rather than extra scope:

> The command reads `~/.mcp-use/config.json`. When that file selects no organization, it reports the account default, matching `org current` and the `servers`/`deployments` commands. This is the usual case when authenticating with `MCP_USE_API_KEY` instead of `mcp-use login`.

Only the v2 reference is touched; `docs/typescript/` is marked legacy in `docs/AGENTS.md`.

khandrew1 (approved):
Thanks for the contribution!

PR #2588, opened 2026-09-19
fix(python): keep multi-line strings intact in the code executor

sclfcz:
# Pull Request Description

While reading `libraries/python/mcp_use/client/code_executor.py` to understand how code mode wraps agent code. No issue was filed for it; this is the reproduction against current `main`:

```python
import asyncio
from unittest.mock import MagicMock

client = MagicMock(spec=MCPClient)
client.sessions = {}
client.get_server_names.return_value = []

code = 'payload = """line1\nline2"""\nreturn payload\n'
print(repr(asyncio.run(CodeExecutor(client).execute(code))["result"]))
# before: 'line1\n    line2'
# after:  'line1\nline2'
```

`_execute_code` wraps the agent's code in an async function to support top-level `await` and `return`, which needs one extra level of indentation. It prefixed **every** line of the source, including the lines that belong to a multi-line string literal. The string is then compiled with four extra spaces on every line after the first, so the value the agent wrote is not the value the tool receives. Nothing fails: the code runs and returns silently altered data, which is easy to miss when the payload is SQL, JSON, Markdown or a nested script.

## Language / Project Scope

Check all that apply:
- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Indent only the line that starts a logical line, as reported by `tokenize`. Continuation lines (inside brackets, after a backslash, or inside a multi-line string literal) are copied verbatim, so the executed program is the program that was passed in.

## Implementation Details

1. Added `_indent_logical_lines(code, prefix)`: it walks the token stream and records the first token after each `NEWLINE` (skipping `NL`, `COMMENT`, `INDENT`, `DEDENT`, `ENDMARKER`), then prefixes exactly those lines.
2. `_execute_code` now builds the wrapper from `_indent_logical_lines(code)` instead of prefixing every line.
3. Line endings are normalized first: the tokenizer does not treat a lone `"\r"` as a line break while `compile()` does, so normalizing keeps the tokenizer's line numbers aligned with the lines being indexed (the compiled program is unaffected, `compile()` performs the same translation).
4. If the source cannot be tokenized (unterminated string, unbalanced brackets), it falls back to indenting every line and lets `compile()` report the syntax error, as before.
5. No new dependencies; `tokenize` and `textwrap` are standard library. `ruff format`/`ruff check` and `ty check` are clean (the 3 pre-existing `ty` diagnostics in this file are unchanged).

## TypeScript Checklist

## Python Checklist

### Pre-commit Checklist

## Example Usage (Before)

```python
code = 'payload = """SELECT *\nFROM users\n"""\nreturn await server.echo(payload=payload)\n'
```

The tool received `{"payload": "SELECT *\n    FROM users\n"}`: the second line picked up the wrapper's indentation.

## Example Usage (After)

```python
code = 'payload = """SELECT *\nFROM users\n"""\nreturn await server.echo(payload=payload)\n'
```

The tool receives `{"payload": "SELECT *\nFROM users\n"}`.

## Documentation Updates

* None: this is an internal fix to the code executor; no documented behaviour changes.

## Testing

Unit tests added in `tests/unit/test_code_executor.py`:

- `TestIndentLogicalLines`: statement lines are indented, bracket/backslash continuation lines are untouched, line endings are normalized.
- `TestCodeExecutorSourceFidelity`: multi-line string value, a multi-line string whose content is itself indented code, a multi-line f-string with a replacement field on a continuation line, the payload a mocked MCP tool receives, nested blocks with a standalone comment, and top-level `await`/`return` still working.

Edge cases considered:

- **Reverse verification**: restoring the previous line-by-line indentation makes the four fidelity tests fail (`4 failed, 26 passed`), so the tests pin the bug.
- **Python versions**: the wrapping logic was exercised on 3.11, 3.12 and 3.13 (the CI matrix for the Python library is 3.11/3.12). 3.12+ tokenizes f-strings into `FSTRING_*` tokens; the replacement-field tokens of a multi-line f-string do not trigger a `NEWLINE`, so no line inside a literal is indented on any of the three versions.
- **Leading indentation**: code that starts with an indent (agents occasionally emit it) still runs, as before, because `INDENT` is skipped and the first real token of the line is the one indented.
- **Syntax errors**: unterminated strings, unbalanced brackets and missing colons are still reported through `result["error"]` instead of raising.
- **Full suite**: `pytest tests/unit` → `314 passed` (the `test_create_sandboxed_stdio_connector` failure that appears without the optional `e2b` extra is unrelated to this change).

## Backwards Compatibility

Fully backwards compatible. Well-formed code keeps running exactly as before, and its behaviour only changes where the previous indentation silently altered a multi-line string literal, i.e. the bug being fixed. Code that relied on the injected indentation inside its own strings would be affected, but that is the corrupted value rather than a documented behaviour.

## Related Issues

None: no issue was filed for this, so there is nothing to close. Happy to open one instead if you prefer to track it that way.

---
## Summary by cubic
Fixes the Python code executor so multi-line strings inside agent code are no longer corrupted by the async wrapper. Previously every line received an extra four spaces, silently altering the value that reaches MCP tools; now only the first line of each logical statement is indented.

- Indents only statement-starting lines using `tokenize`, leaving continuation lines (inside brackets, after backslashes, or within multi-line strings) verbatim.
- Normalizes `\r\n`/`\r` to `\n` and splits on `\n` only, keeping tokenizer line numbers aligned even when strings contain form feed or other line-boundary characters.
- Falls back to indenting every line if the code cannot be tokenized, preserving existing syntax error reporting.
- Adds unit tests covering multi-line strings, f-strings, tool payloads, top-level `await`/`return`, and odd boundary characters.

<sup>Written for commit 86cf7dee4511127841c9b1e312301543cfc1c717. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2588?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sclfcz:
Both findings were valid, thank you. Fixed in 86cf7de:

- **P2 (line boundaries):** confirmed with your repro — `_indent_logical_lines('value = """a\fb"""\nreturn value\n')` returned `'    value = """a\x0c    b"""\nreturn value\n'`, injecting the prefix inside the literal and leaving `return value` unindented, where the previous blanket indent handled that input. The function now splits on `"\n"` only, which is the separator the tokenizer sees once line endings are normalized, so the element count always matches the tokenizer's line numbers. Covered by `TestIndentLogicalLines::test_other_line_boundary_characters_are_not_line_breaks` and an end-to-end `TestCodeExecutorSourceFidelity::test_string_with_form_feed_keeps_its_value`; both fail on the previous revision (`2 failed, 32 passed`).
- **P3 (fallback coverage):** added `test_unterminated_string_indents_every_line` and `test_unbalanced_brackets_indent_every_line`, which exercise the `TokenError` path directly.

Full suite: `318 passed`; `ruff check`/`ruff format --check` clean; `ty check` unchanged (same 3 pre-existing diagnostics).

khandrew1:

PR #2589, opened 2026-09-19
fix(python): keep a server's tools when one tool name is unusable

sclfcz:
# Pull Request Description

While reading `_build_namespace()` in `libraries/python/mcp_use/client/code_executor.py`. No issue was filed for it; this is the reproduction against current `main`:

```python
import asyncio
from unittest.mock import AsyncMock, Mock

from mcp.types import Tool  # an empty name passes the SDK model

broken = Tool(name="", description="No usable name", inputSchema={})
usable = Tool(name="read_file", description="Read a file", inputSchema={})

session = AsyncMock()
session.list_tools = AsyncMock(return_value=[broken, usable])

client = Mock(spec=MCPClient)
client.sessions = {"files": session}
client.get_server_names = Mock(return_value=[])

# 'files' namespace is missing, so read_file cannot be called:
# ERROR - Failed to load tools for server files: string index out of range
print(asyncio.run(CodeExecutor(client)._build_namespace())["__tool_namespaces"])
# before: []
# after:  ['files']
```

`_build_namespace()` sanitizes every tool name into a Python identifier:

```python
sanitized_name = re.sub(r"[^a-zA-Z0-9_]", "_", tool_name)
if not sanitized_name[0].isalpha() and sanitized_name[0] != "_":
```

For a tool whose name has no usable character (`""`, which `mcp.types.Tool` accepts) the sanitized name is empty, so indexing it raises `IndexError("string index out of range")`. That happens inside the `try` that wraps the per-server loop, so the handler drops the **whole namespace** and records a misleading message — the server did not fail to load, one of its tools has an unusable name.

The result is inconsistent state for the agent: `search_tools()` still lists every tool of that server (it reads `session.list_tools()` directly), while none of them can be called, because the namespace the wrappers are attached to was never registered.

## Language / Project Scope

Check all that apply:
- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Skip a tool whose name cannot be expressed as an identifier, log a warning that names the server and the tool, and keep building the namespace so the other tools of that server stay callable.

## Implementation Details

1. `_build_namespace()` checks the sanitized name before indexing it and `continue`s on an empty result, with a warning that includes the server and the raw tool name.
2. Behaviour for every other tool is untouched: the sanitization, the `_1abc` style prefix for names that start with a digit, and the alias that keeps the original name when it is already a valid identifier are all unchanged.

## TypeScript Checklist

## Python Checklist

### Pre-commit Checklist

## Example Usage (Before)

```python
# A server exposes one tool with an empty name and one usable tool
search_tools()          # -> lists both tools
await files.read_file(...)  # AttributeError: the namespace was dropped
```

## Example Usage (After)

```python
# The unusable tool is skipped with a warning, the rest of the server works
search_tools()          # -> lists both tools
await files.read_file(...)  # -> calls the tool
```

## Documentation Updates

* None: the change only affects how a malformed entry from a server is handled.

## Testing

Unit tests added in `tests/unit/test_code_executor.py` (`TestCodeExecutorToolNamespaces`):

- a server with an unusable name plus a usable tool keeps the usable tool callable (asserts the call reaches `session.call_tool`),
- a server whose only tool has an unusable name still registers its namespace.

Edge cases considered:

- **Reverse verification**: removing the guard makes both tests fail (`2 failed, 21 passed`), so they pin the bug.
- **Nothing becomes unreachable**: a tool with an empty name had no attribute on the namespace to be called through, so skipping it does not remove a capability that worked before.
- **Full suite**: `pytest tests/unit` → `307 passed`; `ruff check`/`ruff format --check` clean; `ty check` reports the same 3 pre-existing diagnostics as `main`.

## Backwards Compatibility

Fully backwards compatible. Servers with well-formed tool names behave exactly as before; the only change is that a server is no longer punished for one unusable entry.

## Related Issues

None: no issue was filed for this, so there is nothing to close. Happy to open one instead if you prefer to track it that way.

---
## Summary by cubic
Fixes the Python code executor's handling of tool names that can't be used as Python identifiers. Previously, one unusable name dropped the whole server namespace while `search_tools()` still advertised those tools; now the unusable tool is skipped with a warning, the rest of the server's tools stay callable, and discovery hides names that can't be exposed, so no callable tool is lost.

<sup>Written for commit 43311275977ddbe63ccb542cdcf6ee1839eb99a0. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2589?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sclfcz:
Both findings were valid, thank you. Fixed in 4331127:

- **`search_tools` advertised uncallable tools:** both paths now share a `_tool_identifier()` helper (the sanitization moved there), and `search_tools` skips a tool whose identifier would be empty, so discovery matches what the namespace can expose. Covered by `test_search_tools_does_not_advertise_unusable_tools` (results and `meta.total_tools`).
- **Test docstring overclaimed:** `test_unusable_names_do_not_hide_other_servers` now registers a second server, calls a tool through it, and asserts `__tool_namespaces` lists both, so the behaviour the docstring names is actually exercised.

Reverse verification with the previous implementation (the `sanitized_name[0]` indexing and no filter in `search_tools`): `3 failed, 21 passed`. Full suite: `308 passed`; `ruff check`/`ruff format --check` clean; `ty check` unchanged.

khandrew1:

PR #2590, opened 2026-09-19
fix(python): keep colliding tool names reachable in code mode

sclfcz:
# Pull Request Description

While reading `_build_namespace()` in `libraries/python/mcp_use/client/code_executor.py`, next to the unusable-name case reported in #2589. No issue was filed for it; this is the reproduction against current `main`:

```python
import asyncio
from unittest.mock import AsyncMock, Mock

from mcp.types import Tool

# Two distinct tools that sanitize to the same identifier
dotted = Tool(name="weather.get", description="Get the weather", inputSchema={})
dashed = Tool(name="weather-get", description="Get the weather as well", inputSchema={})

session = AsyncMock()
session.list_tools = AsyncMock(return_value=[dotted, dashed])
session.call_tool = AsyncMock(return_value=Mock(content=[Mock(text="ok")]))

client = Mock(spec=MCPClient)
client.sessions = {"weather": session}
client.get_session = Mock(return_value=session)
client.get_server_names = Mock(return_value=[])

code = "return await weather.weather_get()\n"
asyncio.run(CodeExecutor(client).execute(code))
print(session.call_tool.await_args.args[0])
# before: 'weather-get' -- the second wrapper had replaced the first, silently
# after:  'weather.get' -- each name keeps its own wrapper, the second tool is
#                        exposed as `weather_get_2` and a warning says so
```

`_build_namespace()` turns each tool name into a Python identifier and attaches the wrapper with `setattr(server_namespace, sanitized_name, wrapper)`. Sanitization is not injective — `re.sub(r"[^a-zA-Z0-9_]", "_", name)` maps `weather.get` and `weather-get` to the same `weather_get` — so the second tool silently replaced the first wrapper. The agent then calls a tool it did not ask for, and which of the two survives depends on the order the server happens to list them in. The same applies to the "keep the original name when it is a valid identifier" alias, which could overwrite an identifier already in use by another tool.

## Language / Project Scope

Check all that apply:
- [ ] TypeScript
- [x] Python
- [ ] Documentation only
- [ ] CI/CD or tooling

## Changes

Keep one wrapper per tool: expose a tool whose sanitized name is already taken under a suffixed identifier (`weather_get_2`, …, deterministic) and log which tool was renamed to what. The "original name" alias no longer overwrites an identifier that another tool already owns.

## Implementation Details

1. `_build_namespace()` tracks the identifiers it has exposed for a server in a local `exposed` dict mapping identifier to the tool name that owns it.
2. On a collision with a *different* tool, the identifier gets a numeric suffix (`_2`, `_3`, … until free) and a `logger.warning` names the tool, the identifier it wanted and the one it received. A server that repeats the same tool name twice is untouched: the owner matches, so the wrapper is simply refreshed.
3. The identifier alias (`tool_name` itself, when it is a valid Python identifier, e.g. `café`) is only attached when that name is still free, so it cannot replace another tool's wrapper.
4. Sanitization, digit-prefixing (`1tool` -> `_1tool`) and wrapper creation are otherwise unchanged.
5. Tools are still processed in the order the server returns them; only the identifier they end up with changes.

## TypeScript Checklist

## Python Checklist

### Pre-commit Checklist

## Example Usage (Before)

```python
search_tools()            # -> lists weather.get and weather-get
await weather.weather_get()   # -> calls weather-get (silently the wrong tool)
```

## Example Usage (After)

```python
search_tools()            # -> lists weather.get and weather-get
await weather.weather_get()   # -> calls weather.get
await weather.weather_get_2() # -> calls weather-get
# WARNING - Tool 'weather-get' of server weather also sanitizes to 'weather_get';
# exposing it as 'weather_get_2'
```

## Documentation Updates

* None: only the naming of a colliding wrapper changes, plus the warning that explains it.

## Testing

Unit tests added in `tests/unit/test_code_executor.py` (`TestCodeExecutorToolNameCollisions`):

- `test_colliding_names_do_not_replace_each_other`: both `weather_get` and `weather_get_2` are callable and each reaches its own tool (`call_tool` records `weather.get` then `weather-get`),
- `test_collision_is_reported`: both identifiers exist and the warning names the tool and the new identifier,
- `test_rewritten_name_keeps_its_original_identifier_alias`: the alias path (`café` -> `caf_` plus `café`) still works, which covers the changed alias condition.

Edge cases considered:

- **Reverse verification**: restoring the previous implementations makes the two collision tests fail (`2 failed, 22 passed`), so they pin the bug; the alias test passes on both revisions, as it guards behaviour that already existed.
- **Repeated names**: a server listing the same tool twice refreshes the same wrapper instead of creating a `_2` duplicate.
- **Unicode names**: `café` is a valid Python identifier but is rewritten by the sanitizer, so it keeps both names when they are free.
- **Full suite**: `pytest tests/unit` -> `308 passed`; `ruff check` / `ruff format --check` clean; `ty check` reports the same 3 pre-existing diagnostics as `main`.

## Backwards Compatibility

Mostly backwards compatible, with one intentional difference: when two tool names collide, the second tool is now reachable under a suffixed name instead of replacing the first wrapper. Code that relied on the silently overwritten wrapper was calling an arbitrary one of the two tools; both are now addressable and the rename is logged.

## Related Issues

None: no issue was filed for this, so there is nothing to close. Sibling fixes for the same builder that are independent of this one: #2588 (multi-line strings in executed code) and #2589 (tools whose name has no usable character).

---
## Summary by cubic
Fixes code-mode tool name collisions so distinct tool names that sanitize to the same Python identifier both stay reachable. Previously the last wrapper silently overwrote the first; now the later tool gets a suffixed identifier (`weather_get_2`) and a warning names the renamed tool. A tool whose original name is already the identifier keeps it, the original-name alias no longer overwrites another tool's identifier, and repeated identical names still produce one wrapper.

<sup>Written for commit 4bf69e4446b94ec99699789f9090d62b69b8ddc6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2590?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


sclfcz:
Both findings were valid, thank you. Fixed in 4bf69e4:

- **P2 (exact name should win):** agreed, and the ordering was the missing piece — the previous revision still bound `get_stock` to the `get-stock` tool. Tools whose name is already its own identifier are now registered first (`sorted` on `re.sub(...) != entry.name`, stable, so the server's order is kept within each group), so `get_stock` keeps `get_stock` and `get-stock` becomes `get_stock_2`. Covered by `test_exact_name_wins_the_identifier`, which calls both identifiers and asserts `call_tool` received `get_stock` then `get-stock`; removing only the ordering line makes that single test fail.
- **P3 (repeat case untested):** added `test_repeated_tool_name_is_not_renamed` — the same name twice yields one wrapper and no warning.

Reverse verification: restoring the previous implementation fails the three collision tests (`3 failed, 23 passed`). Full suite: `310 passed`; `ruff check`/`ruff format --check` clean; `ty check` unchanged.

khandrew1:

PR #2619, opened 2026-09-22
fix(cli): normalize configured skills directories

### LINKED ISSUE #2596: cli: skills.directory is validated trimmed but resolved untrimmed, so a stray space silently points at a missing directory
### Summary

`assertSafeDirectory` in `libraries/typescript/packages/cli/src/skills/node-loader.ts:25-32` validates the trimmed value but resolves the untrimmed one:

```ts
function assertSafeDirectory(projectRoot: string, directory: string): string {
  if (directory.trim() === "" || isAbsolute(directory)) {
    throw new TypeError(
      "skills.directory must be a non-empty project-relative path"
    );
  }
  const root = resolve(projectRoot);
  const resolved = resolve(root, directory);
```

`directory.trim()` gates the blank check, then `resolve(root, directory)` uses the original string. So whitespace is meaningful to the resolved path but not to validation.

### The Problem

1. `skills: { directory: " skills" }` resolves to `<root>/ skills`, a sibling directory whose name begins with a space, rather than `<root>/skills`.
2. The path passes the containment check, so nothing is rejected. `resolveConfiguredSkillsDirectory` (line 51) returns it, and the caller watches and reads a directory that does not exist, so no skills are discovered and nothing explains why.
3. The two checks disagree about the same value: a directory of `"   "` is rejected as blank, while `" skills"` is accepted and then used verbatim. Either whitespace is insignificant, in which case it should be trimmed before use, or it is significant, in which case the blank check should not trim.

Reproduced on `main` at `5bf924c`:

```
resolveConfiguredSkillsDirectory({ directory: " skills" }, "/tmp/project")
// "/tmp/project/ skills"   expected "/tmp/project/skills"
```

**Test coverage:** none. There is no test anywhere for `node-loader.ts` or `resolveConfiguredSkillsDirectory`; grepping `tests/` for either name returns nothing.

### Proposed fix

Trim once, then use the trimmed value for both the checks and the resolution:

```ts
  const trimmed = directory.trim();
  if (trimmed === "" || isAbsolute(trimmed)) {
    throw new TypeError(
      "skills.directory must be a non-empty project-relative path"
    );
  }
  const root = resolve(projectRoot);
  const resolved = resolve(root, trimmed);
```

Blank and whitespace-only values are still rejected, `../outside` is still rejected by the containment check, and ordinary values are unchanged. Trimming `isAbsolute` too means `" /etc"` is now correctly rejected rather than resolved as a relative path.

Low severity, since it needs a stray space in config. Filing it as a consistency defect rather than a user report.

luochen211:
## Summary

- normalize surrounding whitespace before validating and resolving `skills.directory`
- apply the same normalization to conventional directory overrides
- add regression coverage for whitespace, traversal, absolute paths, defaults, and explicit disablement

Fixes #2596

## Test plan

- `packages/cli/node_modules/.bin/vitest run packages/cli/tests/skills/node-loader.test.ts`

---
## Summary by cubic
Fixes #2596 by trimming surrounding whitespace from configured skills directory paths before validation and resolution, so discovery and watching use the same normalized project-relative path. Applies the same normalization to conventional directory overrides and adds regression tests for whitespace, traversal, absolute paths, defaults, and disabling.

<sup>Written for commit 5d0e71b5564fa1bdc260827449341244b48995a3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/mcp-use/mcp-use/pull/2619?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luochen211:
Superseded by merged #2597. Current main trims skills.directory before validation and resolution and includes regression coverage for padded paths, absolute paths, defaults, and traversal. Closing this duplicate PR.

